PR #10, opened 2016-09-19
rough carousel

danmarshall:
the styling leaves a lot to be desired, but some basics are sketched in


package.json
```diff
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "http-server": "^0.9.0",
+    "source-loader": "^0.2.0",
```
billba: Do we need this anymore? I'm running fine without this.

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+        };
+
+    }
+
+    private componentDidMount() {
+        var li = this.state.ul.firstChild as HTMLLIElement;
+        this.state.itemWidth = li.offsetWidth;
+    }
+
+    private scrollBy(increment: number) {
```
billba: I'm not necessarily disagreeing with this strategy, but I'd love you to explain to me why this is all going on here and not just adjusting state for successive renders.

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        this.state = {
+            scrollIndex: 0,
+            itemWidth: null
+        };
+
+    }
+
+    private componentDidMount() {
+        var li = this.state.ul.firstChild as HTMLLIElement;
```
billba: const (or just inline this as an expression in the next line)

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    private scrollBy(increment: number) {
+        //this.state.scrollIndex -= increment;
+        //this.state.scrollDiv.scrollLeft += increment * this.state.itemWidth;
+        //this.setState(this.state);
+
+        //TODO: cancel existing animation when clicking fast
+        //TODO: animate to boundaries
+
+        var unit = increment * this.state.itemWidth;
```
billba: const

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        var div = document.createElement('div');
```
billba: const

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+        div.className = 'wc-animate-scroll';
+        div.style.left = this.state.scrollDiv.scrollLeft + 'px';
+        document.body.appendChild(div);
+
+        //capture ComputedStyle every millisecond
+        var captureTimer = setInterval(() => {
```
billba: const

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+        var finishTimer = setTimeout(() => {
+            clearInterval(captureTimer);
+            document.body.removeChild(div);
+        }, 1000 * parseFloat(getComputedStyle(div).transitionDuration) || 5000);
+
+    }
+
+    render() {
+
+        let items = this.props.attachments.map((attachment, i) => {
```
billba: const

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+            return <li><Attachment attachment={attachment} actions={this.props.actions}/></li>;
```
billba: omit the brackets, return, and semicolon

src/Carousel.tsx
```diff
@@ -0,0 +1,80 @@
...
+        }) as React.ReactElement<any>[];
```
billba: do you need this type coercion at the end?

src/History.tsx
```diff
@@ -4,7 +4,7 @@ import { Timestamp } from './Timestamp.tsx';
 import { MessageGroup, HistoryActions } from './App.tsx';
 import { HistoryMessage } from './HistoryMessage.tsx';
 
-interface Props {
+export interface Props {
```
billba: Does this actually help anything? At some point I tried this and it didn't seem to make a difference.

Ditto with the other props you exported below.

tsconfig.json
```diff
@@ -1,22 +1,23 @@
...
-        "src/directLine.ts",
-        "src/mockLine.ts",
-        "src/HeroCard.tsx",
-        "src/Timestamp.tsx",
-        "src/HistoryMessage.tsx",
-        "src/Console.tsx",
-        "src/History.tsx",
-        "src/App.tsx"
-    ]
+  "compilerOptions": {
```
billba: what happened here? Did the indentation go away?

src/Carousel.tsx
danmarshall: Generally I'm trying to leverage the native scrolling and not reinvent it. I'm using the scrollLeft property which is driven by resize. This is still being worked out.

PR #16, opened 2016-10-03
Add DebugView panel

eanders-MS:
This PR adds the slide-out DebugView panel, to show JSON of selected message.


src/BotChat.tsx
```diff
@@ -1,222 +1,301 @@
...
+    isDebuggerVisible() {
+        return this.state.debugViewState === DebugViewState.visible;
+    }
+
+    isDebuggerEnabled() {
+        return this.state.debugViewState !== DebugViewState.disabled;
+    }
+
+    render() {
+        return <div className="wc-app">
```
billba: I prefer putting "return (" on the first line, and then indent the DIV underneath it so that all the tags line up under each other.

src/History.tsx
```diff
@@ -1,50 +1,59 @@
...
+    }
+
+    render() {
+        return (
+            <div className="wc-message-groups" ref={ref => this.scrollMe = ref} onScroll={ e => this.props.actions.setAutoscroll((e.target as any).scrollTop + (e.target as any).offsetHeight >= (e.target as any).scrollHeight) }>
+                <div className="wc-message-group">
+                { this.props.activities
+                    .filter(activity => activity.type === "message" && (activity.from.id != this.props.userId || !activity.id))
+                    .map((activity:Message) => {
+		    	        let clickable = this.props.debuggerVisible;
```
billba: there is a mix of tabs and spaces in the indentation here (and elsewhere) that is screwing things up.

eanders-MS: Ah, it must be my new diff tool. I will check its settings, fix this, and add another commit.

src/History.tsx
```diff
@@ -24,22 +26,30 @@ export class History extends React.Component<Props, {}> {
...
-                            <div className="wc-message-content">
-                                <svg className="wc-message-callout">
-                                    <path className="point-left" d="m0,0 h12 v10 z" />
-                                    <path className="point-right" d="m0,10 v-10 h12 z" />
-                                </svg>
-                                <HistoryMessage activity={ activity } actions={ this.props.actions }/>
+                    .map((activity:Message) => {
+                        let clickable = this.props.debuggerVisible;
+                        let selected = activity === this.props.selectedActivity;
+                        let contentClassName = 'wc-message-content' + (clickable ? ' clickable' : '') + (selected ? ' selected' : '');
```
billba: For different reasons, these locals don't really buy anything. Let's get rid of them.

eanders-MS: I think they buy some code clarity, but happy to change.

PR #19, opened 2016-10-03
lost a return stmt

eanders-MS:
Accidental deletion of a return statement. Looks like a big diff because if additional level of indentation.


billba:
Decided to fix this by removing the braces rather than adding a return.

PR #24, opened 2016-10-04
Preserve non-bundled js files from build

eanders-MS:
This might be a controversial request. I'm working on a new Bot Framework emulator - an electron app. I do not want a webpacked build for the webchat control because I'll be using a different package manager. These changes preserve the intermediate js files in a ./built folder.


billba:
Makes sense, but what you think of excluding the 'built' directory from git?

PR #25, opened 2016-10-05
Redux

billba:


danmarshall:
I don't know much about Redux, but I must say the code looks cleaner :)

PR #60, opened 2016-10-12
only show first image in cards

danmarshall:


billba:
I'm going to do a slightly bigger fix for this one, coming right up.

billba:
Aside from some cleanup, the main improvement in my fix (https://github.com/billba/botchat/commit/3f2676c8dd426f13b63f3346c3b665a45cefeeab) is that we check to see if the images field exists before we dereference it.

PR #67, opened 2016-10-16
Put Redux store in global space

eanders-MS:
Works in browser and Node.


eanders-MS:
Oops, I forgot that when I commit anything while a PR is open, the new commit goes into the existing PR automatically. That's annoying. Oh well, was going to issue a new PR for this commit anyway. Changes:

This is a pretty big change. Apologies for the review burden that places on you.
- Decoupled UI and BotChat components
- Added new component called App, which can render everything.
- Factored global Store into its own file: Store.ts
- Factored redux reducers for components out of BotChat and into the component's file. It felt weird to have DebugView's update code in BotChat. I wanted it to live closer to the component, but didn't want to move only DebugView's - so I moved all of them. The separation of logic/knowledge could be worked on further but is ok for now I think.
- Each component now unsubscribes from the Store when it unmounts. This seemed like a good idea for running in the Emulator.
- Added a new webpack build for the App component.
- Added app.html for testing App component.
- Unrelated: I changed the package.json entry point, since the webpack entry point doesn't work in the Emulator and I had to manually go change it every time I updated the package.

PR #88, opened 2016-10-28
Move Redux Store into Chat. Remove DebugView.

eanders-MS:
Tried removing the store subscription/forceUpdate from child components. It didn't work. The control stopped responding.


billba:
Oh yeah, we'd need one in chat. Let's leave it this way for now.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

src/Store.ts
```diff
@@ -1,4 +1,4 @@
-import { Store, Reducer, createStore, combineReducers } from 'redux';
+import { Store, Reducer, createStore as reduxCreateStore, combineReducers } from 'redux';
 import { Activity, IBotConnection, User } from './BotConnection';
 import { FormatOptions } from './Chat';
 import { strings, Strings } from './Strings';
```
billba: I propose exporting type ChatStore = Store<ChatState> - that way all the places where you set props: Store<ChatStore> can become props: ChatStore, which is one fewer import and a simpler declaration

src/FormattedText.tsx
```diff
@@ -8,15 +8,12 @@ interface Props {
...
-}
-
-export class FormattedText extends React.Component<Props, State> {
+export class FormattedText extends React.Component<Props, {}> {
     constructor(props: Props) {
         super(props);
     }
 
-    shouldComponentUpdate(nextProps: Props, nextState: State, nextContext: any): boolean {
+    shouldComponentUpdate(nextProps: Props, nextState: {}, nextContext: any): boolean {
```
billba: cleaner to just not list the last two parameters since you don't use them.

src/Store.ts
eanders-MS: OK, I've made this change.

src/FormattedText.tsx
eanders-MS: Ok, I've made this change.

PR #95, opened 2016-10-30
Extract FormattedText to it's own module.

eanders-MS:
- Reference FormattedText from new project at https://github.com/fuselabs/react-formattedtext. Will make this an npm module later.
- index.html: Add domain parameter to DirectLine's constructor call.


eanders-MS:
apostrophe fail in commit message.

PR #118, opened 2016-11-05
Show bot name below message if available. Fallback to id.

eanders-MS:
-Show `from.name` below incoming chat messages if it was provided. Fallback to `from.id` otherwise.
-Also deleted some dead css classes (were leftover from the debugview).


eanders-MS:
ping. :)

billba:
WHAT I AM WIDE AWAKE

PR #160, opened 2016-11-22
always autoscroll unless user scrolled away from bottom

danmarshall:
There is a threshold of 80 pixels from the bottom which we consider that we are "docked" to the bottom and will autoscroll.

### DIFF
--- a/src/History.tsx
+++ b/src/History.tsx
@@ -13,13 +13,29 @@ interface Props {
 export class History extends React.Component<Props, {}> {
     scrollMe: HTMLElement;
     scrollToBottom = true;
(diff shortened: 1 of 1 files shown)


billba:
This looks like a partial checkin. Since you changed the signature for History, at minimum you need Chat.tsx too, right?

danmarshall:
It's just initializing some members, not altering the signature.

danmarshall:
it's a complete checkin :)

billba:
That's what I get for trying to understand a PR on my phone. Will look more closely when I return to civilization.

PR #161, opened 2016-11-22
Fixed domain

jcmontes95:
The Direct Line URL is more leading than the localhost one.
When I had to implement my custom webchat I found the 'domain' and 'allowMessagesFrom' fields quite confusing. It took me a while until I realized that the DL url should be placed there. And since most of us don't have the possibility to run DL locally it would be better for everyone if it just had the right one to begin with.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -30,7 +30,7 @@
             const botConnection = new BotChat.DirectLine({
                 secret: params['s'],
                 token: params['t']
-                }, params["domain"]
+                }, params["domain"] //the domain should be the DL url https://directline.botframework.com/v3/directline
             );
 
             const user = {
@@ -47,7 +47,7 @@
                 botConnection,
                 user,
                 bot,
-//              allowMessagesFrom: ['http://localhost:8080']
+//              allowMessagesFrom: ['https://directline.botframework.com/v3/directline'] 
             }, document.getElementById("BotChatGoesHere"));
         </script>
     </body>


public/index.html
```diff
@@ -30,7 +30,7 @@
             const botConnection = new BotChat.DirectLine({
                 secret: params['s'],
                 token: params['t']
-                }, params["domain"]
+                }, params["domain"] //the domain should be the DL url https://directline.botframework.com/v3/directline
```
billba: This is not correct. The purpose of this optional parameter is to override the default domain you specified.

public/index.html
```diff
@@ -47,8 +47,8 @@
                 botConnection,
                 user,
                 bot,
-//              allowMessagesFrom: ['http://localhost:8080']
+//              allowMessagesFrom: ['https://directline.botframework.com/v3/directline']
```
billba: This is not correct. The purpose of this optional parameter is to allow the experimental "backchannel" feature. It is unrelated to (and would never be) the Direct Line URL.

billba:
It sounds like I need to document these parameters. They are purely optional.

public/index.html
jcmontes95: I found no documentation on default parameters. In the script, [params is empty](https://github.com/Microsoft/BotFramework-WebChat/blob/master/public/index.html#L20) 
and if I left it unset the whole thing wouldn't work.

public/index.html
jcmontes95: That one's on me, sorry

public/index.html
billba: if you look at the definition for DirectLine you will see that `domain` is optional. But you do need to pass a secret or a token as the first parameter.

billba: Also, `params` is _initialized_ to empty, but is then filled in based on the arguments in the querystring. The point of this file is to allow simple testing of WebChat by passing in arguments via the URL.

Issue #183, opened 2016-11-30
hardcoded placeholder string in Shell.tsx


Maarten88:
Nitpick encountered while trying to embed and customize the control:

    placeholder="Type your message..."

should be something like

    placeholder={this.state.format.strings.placeholder}

billba:
Oooh, good catch. Care to submit a PR for that?

PR #187, opened 2016-11-30
Make console placeholder a configurable string

Maarten88:
As requested in #183

### DIFF
--- a/src/Strings.ts
+++ b/src/Strings.ts
@@ -9,6 +9,7 @@ export interface Strings {
     messageFailed: string,
     messageSending: string,
     timeSent: string,
+    placeholder: string
 }
 
 interface LocalizedStrings {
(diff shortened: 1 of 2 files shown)


src/Strings.ts
```diff
@@ -27,6 +28,7 @@ const localizedStrings: LocalizedStrings = {
         messageFailed: "couldn't send",
         messageSending: "sending",
         timeSent: " at %1",
+        placeholder: "Type your message..."
```
danmarshall: perhaps we can name this more specifically - I can see a case where there may be other "placeholders"

Maarten88: I did think about better names, but could not decide. Shall I make it consolePlaceholder? textboxPlaceholder?

danmarshall: Yeah I hear that. How about sendTextboxPlaceholder ?

billba:
Perfect, thanks!

PR #189, opened 2016-12-02
conditional tags

### LINKED ISSUE #140: React error from Thumbnails
`Warning: validateDOMNesting(...): <h2> cannot appear as a descendant of <p>. See StatelessComponent > p > ... > h2.`

danmarshall:
Let me know if you don't like the syntax, it feels lightweight compared to the alternatives.

fixes #140

### DIFF
--- a/public/botchat.css
+++ b/public/botchat.css
@@ -294,6 +294,7 @@
 
     .wc-card.thumbnail img {
         float: right;
+        margin-bottom: 10px;
         margin-left: 10px;
         width: 100px;
     }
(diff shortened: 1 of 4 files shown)


danmarshall:
When I say alternatives, here is an example: https://github.com/ajwhite/render-if

billba:
After some thinking I prefer this:

`{ renderIfNonempty(attachment.content.title, title => <h1>{ title }</h1>) }`

This version of the function would evaluate the first param and, if it's nonempty, pass it to the closure in the second param, then return the results of that. Make sense?

It's a little more verbose but it's more flexible and expressive. I like it much better:

danmarshall:
Nice. Honestly, this seems like it should be part of React itself.

Issue #191, opened 2016-12-03
Start connection in Chat.tsx should happen in componentDidMount, not in constructor


Maarten88:
I am making a public website to promote my bot, with the Chat component on it. The site uses React and redux, so I could nicely add the WebChat components to my sourcecode and integrate in in the design without using the iframe.

My site is isomorphic, meaning it renders the first request on the server. Using the Chat component this way results in an error because the component tries to start the connection in the constructor, which in the server-side case is inside node, which doesn't work.

I think the connection should be opened in componentDidMount, together with setting the subscriptions. I moved the code in my version, now the Chat component has no problems running on the server.

billba:
Makes sense. Go ahead and submit a PR.

I think the only thing that needs to move is `props.botConnection.start()`, yes?

Maarten88:
I think initializing the related subscriptions should move too. They are cleared in componentWillUnmount, so they should be initialized in componentDidMount (says my instinct for symmetry, without having tested the scenario's).

PR #192, opened 2016-12-03
move starting connection into componentDidMount

Maarten88:
See issue #191

This change moves starting the connection and initializing related subscriptions from the constructor to componentDidMount to accomodate server-side rendering and probably fix issues when re-mounting a component into the dom.

### DIFF
--- a/src/Chat.tsx
+++ b/src/Chat.tsx
@@ -46,42 +46,10 @@ export class Chat extends React.Component<ChatProps, {}> {
 
         konsole.log("BotChat.Chat props", props);
 
-        this.store.dispatch({ type: 'Start_Connection', user: props.user, bot: props.bot, botConnection: props.botConnection, selectedActivity: props.selectedActivity } as ConnectionAction);
-
         if (props.formatOptions)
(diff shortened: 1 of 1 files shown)


billba:
Hmmm I think you moved more than is necessary. I think all you need to move is `props.botConnection.start()` - will you try moving just that and see if it solves your problem?

Maarten88:
I think initializing the related subscriptions should move too. They are cleared in componentWillUnmount, so they should be initialized in componentDidMount (says my instinct for symmetry, without having tested the scenario's).

billba:
I'm going to ponder that a bit.

Maarten88:
I've been reading a bit about the React component lifecycle and think my instinct of putting those subscriptions in componentDidMount is correct.

Both componentDidMount and componentWillUnmount do not run when rendering on the server; of course the constructor does run. Leaving those subscriptions in the constructor means they will be created and never get cleared when running on the server. I'm not sure how javascript and node handle this, but I'd expect that to cause a memory leak.

I do think it would be better to start the connection after creating the subscriptions (moving it to the bottom of componentDidMount), but I did not do that in this PR because I did not want to change existing logic or client-side behaviour.

billba:
Great, thanks for researching this.

Issue #199, opened 2016-12-07
Markdown should support links to sip protocol


danmarshall:
We should support the sip protocol markdown such as ```[call me](sip:danmar@microsoft.com)```

Related to https://github.com/Microsoft/BotBuilder/issues/1103

eanders-MS:
We're restricting the protocol to `http` or `https`, but perhaps just open it up to any protocol?

danmarshall:
Not *any* but we should add sip.

eanders-MS:
Bots may want to deep link into apps, especially on mobile. I think we should open it up to any protocol.

danmarshall:
actually it could be a config where the host may specify?

billba:
Agree with Eric. This is basic mobile functionality.

danmarshall:
Security-wise I think there should at least be a gating mechanism, where as the host of the HTML page, I could say "don't allow ftp links" or some such.

eanders-MS:
How about an optional link handler to Chat? Defaults to window.open in new tab.

PR #205, opened 2016-12-07
Don't filter links by protocol

eanders-MS:
Issue https://github.com/Microsoft/BotFramework-WebChat/issues/199


src/FormattedText.tsx
```diff
@@ -55,7 +55,7 @@ export class FormattedText extends React.Component<IFormattedTextProps, {}> {
             tables: true,
             breaks: false,
             pedantic: false,
-            sanitize: true,
+            sanitize: false,
```
danmarshall: Can we expose this setting to the developer?

eanders-MS: Or, something like it. I think it needs to be more flexible. What about an optional hyperlink handler? Let's close issue 199 and open another one for this topic.

PR #211, opened 2016-12-07
Add websocket client support.

sebavan:


src/directLine.ts
```diff
@@ -181,7 +193,7 @@ export class DirectLine implements IBotConnection {
         );
     }
 
-    private getActivity$() {
+    private getGetPollingActivity$() {
```
danmarshall: getGet ?

danmarshall: Can you rename to getPollingActivity

danmarshall: oic, then you might leave it

billba: let's call them: pollingGetActivity$ and webSocketActivity$

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
             .delay(5 * 1000)
         );
     }
+
+    /**
+     * Gets an observable of streamURL by requestings a new one from the conversation APIs. 
+     * @return An Observable of stream url.
+     */
```
billba: We don't do this kind of documentation anywhere else in this project, so let's not do it here

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+    private getWebSocketStreamURL$(): Observable<string> {
+        return this.connectionStatus$
```
billba: let's call this webSocketURL$ ('$' is a shorthand for Stream, so it's redundant to use both)

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+            // If connection is OK
+            .filter(connectionStatus => connectionStatus === ConnectionStatus.Online)
+            // Create an Ajax observable
+            .flatMap(_ => {
+                if (this.streamUrl) {
+                    var copy = this.streamUrl;
```
billba: var -> const

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
     }
+
+    /**
+     * Gets an observable of streamURL by requestings a new one from the conversation APIs. 
+     * @return An Observable of stream url.
+     */
+    private getWebSocketStreamURL$(): Observable<string> {
+        return this.connectionStatus$
+            // If connection is OK
+            .filter(connectionStatus => connectionStatus === ConnectionStatus.Online)
```
billba: Believe it or not I'm not against comments, but none of these seem to add anything to the code. Clearly we are checking for a good connection here...

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+                else {
+                    return Observable.ajax({
+                        method: "GET",
+                        url: `${this.domain}/conversations/${this.conversationId}`,
+                        timeout,
+                        headers: {
+                            "Accept": "application/json",
+                            "Authorization": `Bearer ${this.token}`
+                        }
+                    })// Takes one result
```
billba: ... and this just rephrases what the code already says

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+    /**
+     * Gets an observable of streamURL by requestings a new one from the conversation APIs. 
+     * @return An Observable of stream url.
+     */
+    private getWebSocketStreamURL$(): Observable<string> {
+        return this.connectionStatus$
+            // If connection is OK
+            .filter(connectionStatus => connectionStatus === ConnectionStatus.Online)
+            // Create an Ajax observable
+            .flatMap(_ => {
```
billba: ... and this doesn't add anything to the code (which literally says Observable.Ajax...

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+                var ws$ = Observable.webSocket({
+                    url: url,
+                    resultSelector: e => ({                 
+                        type: e.data ? "ACTIVITY" : "PING",
+                        message: e.data ? JSON.parse(e.data) : null
+                    })
+                });
+
+                // timeout seconds ping for keep alive even if already in the browser.
+                Observable.interval(timeout)
```
billba: I'd rewrite this comment as "Ping the server with empty messages to see if we're still connected to it."

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+                })
+                .delay(timeout)
+            );
+    }
+
+    /**
+     * Gets an Observable of activity from the Web Socket connectivity of direct line.
+     * @return An Observable of activity.
+     */
+    private getWebSocketActivity$(): Observable<Activity> {
```
billba: Again, I'd skip this header comment and most of the below comments that don't add any context.

src/directLine.ts
```diff
@@ -12,34 +12,46 @@ interface ActivityGroup {
     watermark: string
 }
 
+interface WebSocketResponse {
+    type: string;
+    message: ActivityGroup; 
+}
```
billba: let's rename message -> activityGroup (this is the naming convention used elsewhere in this project)

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+                            "Authorization": `Bearer ${this.token}`
+                        }
+                    })// Takes one result
+                    .take(1)
+                    // Extracts the stream URL
+                    .map(result => result.response.streamUrl);
+                }
+            })            
+            // Retries every Timeout seconds in case of connection issue other than security
+            .retryWhen(error$ => error$
```
billba: This is a good comment. Adds a little context.

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+                        type: e.data ? "ACTIVITY" : "PING",
+                        message: e.data ? JSON.parse(e.data) : null
+                    })
+                });
+
+                // timeout seconds ping for keep alive even if already in the browser.
+                Observable.interval(timeout)
+                    .timeInterval()
+                    .subscribe(_ => { ws$.next({}) });
+
```
billba: A question about this code: what does it do if you are e.g. under a tunnel? Does WebSocket.send(...) have a notion of timeout? Or will it sit there forever? The reason I ask is,   I don't want this to throw an error on timeout and then go fetch a new URL, because that won't solve the problem.

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+    /**
+     * Gets an Observable of activity from the Web Socket connectivity of direct line.
+     * @return An Observable of activity.
+     */
+    private getWebSocketActivity$(): Observable<Activity> {
+        // From the Web Socket stream URL
+        return this.getWebSocketStreamURL$()
+            // Create an observable Web Socket
+            .flatMap<string, WebSocketResponse>(url => {
+                // Use a result selector to prevent crash from incoming empty message (it could also helps in timeout detection)
```
billba: helpful comment but instead of "prevent crash" be a little more detailed, e.g. "Observable.webSocket runs JSON.parse() on all incoming messages, but DirectLine sends us empty WebSocket messages, which will crash JSON.parse(). This custom resultSelector avoids the problem."

src/directLine.ts
```diff
@@ -214,4 +226,92 @@ export class DirectLine implements IBotConnection {
...
+            // Only deals with activity message (discard the ping)
+            .filter(data => data.type == "ACTIVITY")
+            // Maps to activity group
+            .map(data => data.message)
+            // Creates observable of activityfor the client as well as updating the watermark           
+            .flatMap(activityGroup => {
+                if (activityGroup.watermark)
+                    this.watermark = activityGroup.watermark;
+                return Observable.from(activityGroup.activities);
+            });
```
billba: This is probably not necessary, but what do you think of extracting this little bit of code, which is shared with the getPolling code, into its own function? We'd end up with something like `flatMap(activityGroup => observableFromActivityGroup(activityGroup))` - a little less repetition, a little more expressive. I'm cool either way.

sebavan:
The PR has been updated following the comments.

built/History.js
```diff
@@ -4,22 +4,22 @@ var __extends = (this && this.__extends) || function (d, b) {
     function __() { this.constructor = d; }
     d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
 };
-var React = require('react');
```
danmarshall: Not related to your code, but to your setup: There seems to be a nuance of difference in your build. Would be nice that all contributors do not keep overwriting each other's build files due to these differences. Perhaps in the future PRs should only include source changes so that churn is minimized.

sebavan: Yep, we discussed it with @billba today and it seems there is a plan for not integrating build files in the PR. This would also highly reduce the merge friction between PRs.

I ll only PR source files in subsequent PRs.

danmarshall: Great, thanks for your contributions!!
I'm still curious as to why there are differences. Perhaps it is the TS compiler?

sebavan: I am in 2.0.9 for tsc.exe if that helps tracking it down. I am not in the default 2.0.1 of vscode and did not go in 2.1 yet due to current project in prod where I need to triple check for breaking changes around constrctor initilization and the extents of built-in types.

PR #231, opened 2016-12-12
refactor Message into separate component

danmarshall:
Small refactoring of the Message into its own component. This is the basis of some upcoming autoscroll fixes but I want this to bake for a bit.

### DIFF
--- a/src/History.tsx
+++ b/src/History.tsx
@@ -4,6 +4,7 @@ import { Activity } from './BotConnection';
 import { HistoryAction, ChatStore } from './Store';
 import { ActivityView } from './ActivityView';
 import { trySendMessage } from './Chat';
+import { Strings } from './Strings';
 
 interface Props {
     store: ChatStore,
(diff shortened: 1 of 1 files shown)


src/History.tsx
```diff
@@ -120,3 +85,70 @@ export class History extends React.Component<Props, {}> {
...
+
+interface MessageProps {
+    activity: Activity;
+    autoscroll: () => void;
+    fromMe: boolean;
+    onClick?: React.MouseEventHandler<HTMLDivElement>;
+    selected: boolean;
+    showTimestamp: boolean;
+    store: ChatStore;
+    strings: Strings;
```
billba: Let's pass either `store` or `strings`, but not both. My preference is to pass in `store`, because that's the convention we use for most of our other React components. Then if you want, set a local variable to `store.getState().format.strings`. (If you do it this way, you can get rid of the `import { Strings } ... ` up top.

src/History.tsx
```diff
@@ -56,58 +57,22 @@ export class History extends React.Component<Props, {}> {
...
-                            <ActivityView store={ this.props.store } activity={ activity } onImageLoad={ this.autoscroll }/>
-                        </div>
-                    </div>
-                    <div className={ 'wc-message-from wc-message-from-' + who }>{ timeLine }</div>
-                </div>
-            );
-        });
+        const wrappedActivities = activities.map((activity, index) => 
+            <Message 
+                key={'message' + index}
```
billba: will you please pad each object with spaces, e.g. `foo={ bar }`,  as we do elsewhere in WebChat?

src/History.tsx
```diff
@@ -120,3 +85,70 @@ export class History extends React.Component<Props, {}> {
         );
     }
 }
+
+interface MessageProps {
+    activity: Activity;
```
billba: If this is really just for messages than this should be `message: Message`

danmarshall: I suppose we could rename it to ActivitySomething

src/History.tsx
danmarshall: I was trying to avoid a call to getState() for every message being rendered. Seemed repeatedly expensive.

billba: Probably not actually an expensive call. but okay. This is good as-is.

danmarshall: Thinking about it, conceptually - Strings might be Props instead of State?

PR #233, opened 2016-12-13
autoscroll when new message appears

### LINKED ISSUE #190: auto-scroll can stop working
When I connect to the Bot Framework's TestBot (an internal tool used for feature testing), and I send it the 'runall' command, auto-scroll in the client will quickly fail. I have to manually scroll to keep up with its output.

danmarshall:
Yet another change to the paradigm. This is the simple approach: 
> when a new message comes in, scroll to the bottom.

Fixes #190

### DIFF
--- a/src/History.tsx
+++ b/src/History.tsx
@@ -6,41 +6,16 @@ import { ActivityView } from './ActivityView';
 import { trySendMessage } from './Chat';
 import { Strings } from './Strings';
 
-interface Props {
+interface HistoryProps {
     store: ChatStore,
     selectActivity?: (activity: Activity) => void
 }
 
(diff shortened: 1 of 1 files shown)


billba:
I'm not sold on this paradigm. For bots, especially, there will be times I want to scroll back and look at previous results. I don't want the system scrolling me to the end every time something new comes in. Consider a bot that sends me updates on server issues even minute. If I'm looking at a big problem that happened an hour ago, I really want to keep looking at it.

danmarshall:
Ok. As @jecristi has pointed out- we should also add a Toast mechanism telling the user that there are new messages. I will retract this PR.

PR #234, opened 2016-12-13
add german locale

pmizel:


billba:
Thank you, this is great.

src/Store.ts
```diff
@@ -24,7 +24,7 @@ export const format: Reducer<FormatState> = (
         options: {
             showHeader: true
         },
-        strings: strings('en-us')
+        strings: strings(window.navigator.language)
```
danmarshall: window.navigator.language || 'en-us'
to avoid a null ref for unknown languages

danmarshall: oic, you hande it below. nvm, thanks!

Issue #213, opened 2016-12-08
Initiate conversation from Javascript


kanwal019:
Hi, 

My bot application requires the username to start the chat conversation. Currently, I am sending it through chat window. Is there a way, I can change the botchat.js to accept the username and send it as a message before starting the formal conversation ?  

Could you provide more documentation of the type script and javascript code ?

danmarshall:
Yes, you can pass userid and username through the url:
https://github.com/Microsoft/BotFramework-WebChat/blob/master/public/index.html#L37

kanwal019:
@danmarshall  Can I add the conversation name in the similar manner ?

billba:
@kanwal019 WebCht doesn't have a concept of "'conversation name"

danmarshall:
@kanwal019 you do have the option of changing the titleBar of the control, through the string localization: https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/Strings.ts#L21

@billba I suppose it would be good to document how someone may pass in their localized string table (without doing their own build)

billba:
@danmarshall currently you can't pass in a localized string table without doing your own build. Interesting idea though.

billba:
@kanwal019 The bot won't have access to the username until WebChat sends a message (at which point you can access it in the `from` field in the message). If you'd like to proactively send a message to the bot before the formal conversation begins, you can send a postBack message (a message that is not shown by the control).

In your web page, after you create the botConnection, do this: 

`
botConnection.postActivity({
    type: 'message',
    text: 'username',
    from: user
}).subscribe();
`

Your bot should look for a message with the text 'username'. In the future there will be a somewhat better way to go about this, but this works.

Please let me know if all the above information solves your problem so that I can close this issue.

kanwal019:
Thank You @billba, I was talking about the conversation id that is displayed in JSON request in emulator. I hope it will help the channel recognize the conversation as a user can have multiple.

The postback message provided got an ajax 403 error, as displayed below.
![capture](https://cloud.githubusercontent.com/assets/15017434/21080989/bb4ccb4c-bfe2-11e6-9620-542198d22674.PNG)

However, passing the username as shown in (https://github.com/Microsoft/BotFramework-WebChat/blob/master/public/index.html#L37), worked for me. I will really help a lot, if I am able to generate and pass a conversation name/id in the same manner.

Thank You very much again :+1:

billba:
[...]

Issue #229, opened 2016-12-12
Deprecate DirectLine.start()


billba:
I’d like to make a subtle change to the DirectLine lifecycle.
 
A while back I added a `start()` method to DirectLine in order to distinguish between creating a DirectLine object (using the constructor) and opening an actual connection to the bot. I also added a corresponding `end()` method to unsubscribe to the various things in DirectLine subscribes too.
 
This turns out to cause a problem with backchannel. The way backchannel works is that the web page creates a DirectLine object which it passes to Chat but also keeps a hold of to send and/or receive messages with the bot. The problem is that, right now, the web page needs to wait until the Chat calls start() by checking `connectionStatus$`. This works, but it's clunky.
 
On analysis, we don't really need `start()` at all. Instead, DirectLIne can just open the connection (start the conversation) when any of the following happen:
·         Someone subscribes to activity$
·         Someone calls postActivity
·         Someone calls postMessageWithAttachments
 
In other words, the connection is made when it’s needed. Again, the current state can be determined by checking `connectionStatus$`.

I would keep a non-functional version of start() around for a little while, so as give @eanders-MS time to update the emulator by removing this call.

eanders-MS:
Sounds fine to me @billba, but no need to keep a stubbed `start()` for my benefit. It should be trivial to adapt the emulator.

PR #238, opened 2016-12-21
Better Direct Line lifecycle, error handling

billba:


billba:
Implements #229 
Fixes #213

PR #241, opened 2016-12-22
Added tag propperty to query string

anset:
Hi,

In order use one bot on multiple websites, I need the bot to know from what site he was called. This way the bot does not need to ask the user this information.

The easiest way I found was to add a "tag" parameter to the URL querystring and have the WebChat UI send the string in this tag to the bot as soon as the connection is made. It is then up to the bot implementation to act upon this first message properly.

I am sending this pull request because I think more users could benefit from this change and the changes are minimal.

Example usage: http://localhost:8000/?s=[bot secret]&tag=makemeasandwhich

Please do let me know if you need more information or want something refactored?

Best regards,

Jhon

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -42,7 +42,8 @@
             
             var bot = {
                 id: params['botid'] || 'botid',
-                name: params["botname"] || 'botname'
+                name: params["botname"] || 'botname',
+                tag: params['tag']
             };
 
             BotChat.App({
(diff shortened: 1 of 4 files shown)


billba:
Thank you for the excellent suggestion. I am working on a more general-purpose approach for sending messages from the enclosing webpage to the bot, but I very much like the ease-of-use of passing a single message at startup. I am going to close this particular PR, but I will likely incorporate your idea.

For now may I suggest that you use `sendPostBack` instead of `sendMessage` as it will send the tag without showing it to the user.

PR #250, opened 2016-12-29
Add Polish location strings and logic

peterblazejewicz:
This is simple PR that adds Polish strings to supported translation together with required changes in logic to support `pl` or `pl-Pl` language codes

Thanks!


billba:
Thank you!

PR #267, opened 2017-01-08
Adding support for rendering flex cards

mousaz:


mousaz:
- Definition of FlexCard.
- Adding support for rendering Flex cards in WebChat.

mousaz:
@billba @danmarshall 
Please have a look on the PR and let me know of any needed changes.
Thanks,
Mousa

src/Attachment.tsx
```diff
@@ -41,6 +41,16 @@ export const AttachmentView = (props: {
         return searchPattern.test(contentType); 
     }
 
+    const imageWithActionWithOnLoad = (image: {url: string, tap?: Button}, thumbnailUrl?: string, autoPlay?:boolean, loop?: boolean) => {
```
billba: Instead of adding two new functions, let's just use the existing ones, conditionally adding the `onClick` parameter if the optional `tap?` field exists in `image`.

Also, minor stylistic nit: when defining objects and types please put a space after `{` and before `}`

src/Attachment.tsx
```diff
@@ -153,6 +163,20 @@ export const AttachmentView = (props: {
                 </div>
             );
 
+        // only supported for 'skype' channel.
```
billba: This comment doesn't really make sense for WebChat (which is specific to the DirectLine channel). Can you clarify that this type is intended for the Emulator for testing Skype-specific bots?

src/BotConnection.ts
```diff
@@ -75,7 +75,23 @@ export interface Receipt {
...
+}  
+
+// only supported for 'skype' channel.
+export interface FlexCard {
+    contentType: "application/vnd.microsoft.card.flex",
+    content: {
+        title?: string,
+        subtitle?: string,
+        text?: string,
+        images?: {
```
billba: truly a small nit, but to keep consistency with the other attachments: please put this on one line

src/BotConnection.ts
```diff
@@ -75,7 +75,23 @@ export interface Receipt {
...
+export interface FlexCard {
+    contentType: "application/vnd.microsoft.card.flex",
+    content: {
+        title?: string,
+        subtitle?: string,
+        text?: string,
+        images?: { 
+            url: string,
+            tap?: Button 
+        }[],
```
billba: Truly a small nit, but to keep consistency with the other attachment definitions, please put this whole image definition on one line.

Issue #230, opened 2016-12-12
Thumbnail Cards Grow Very Wide


IanBoggs:
I've been asked to raise this as a separate Issue, based upon issue #114 

When I show a carousel to the user in the web chat client, the width grows to exceed the window, thus pushing the text the user has entered further to the right outside of the screen.

Basically, this screen shot shows before the carousel is show to the user

![webchat1](https://cloud.githubusercontent.com/assets/18186052/21107309/2ef4f154-c089-11e6-8e14-20bdbcf7a8e5.png)

As you can see, the text entered by the user is shown in blue on the right hand side of the window. Then once the carousel is show this happens

![webchat2](https://cloud.githubusercontent.com/assets/18186052/21107317/37bcfa84-c089-11e6-8c17-9b83fd18e4cf.png)

The text entered by the users is now off the window to the right. I've had a look at the CSS in the borwser developer tools and it appears all the div tags starting 'wc-message' have had their width increased when the carousel is drawn. I'm not much of a CSS/javascript guy though so I'm having difficulty finding the issue

Just for clarity, I am using the code from the '/public' folder for the source code

In terms of the code being used to produce the carousel, I'm using the following C#:

```
               foreach (LawBotExample.Models.Item contentFound in result.ResultList)
                {
                    ThumbnailCard tCard = new ThumbnailCard()
                    { Title = contentFound.company_number,
                        Subtitle = contentFound.title + " - Status " + contentFound.company_status,
                        Images = null
                      };
                    tCard.Text = string.Empty;
                    string tmp = "";
                    if (contentFound.address.address_line_1 != string.Empty)
                        tmp += contentFound.address.address_line_1;
                    if (contentFound.address.address_line_2 != string.Empty)
                        tmp += "," + contentFound.address.address_line_2;
                    if (contentFound.address.postal_code != string.Empty)
                        tmp += "," + contentFound.address.postal_code;
                    if (contentFound.address.locality!= string.Empty)
                        tmp += "," + contentFound.address.locality;
                    while (tmp[0] == ',')
                    {
                        tmp = tmp.TrimStart(',');
                    }

                    CardAction cAction = new CardAction();
                    cAction.Title = "Get More Info";
                    cAction.Type = "imBack";
                    cAction.Value = contentFound.company_number;
                    tCard.Buttons = new List<CardAction>();
                    tCard.Buttons.Add(cAction);
                    if (tmp.Length > 0)
                        tCard.Text += tmp;
[...]

Issue #133, opened 2016-11-11
carousel always shows one card and a bit of the next


billba:
That's okay for very narrow windows, but on a wide window we could show multiple cards, instead we show one VERY WIDE ONE and bit of the next. Waste of space.

billba:
Working better now.

Issue #114, opened 2016-11-03
Herocards, and Thumbnails are VERY WIDE


billba:
Herocards seem to be as wide as the widest single piece of text. Even on a wide screen I would expect a reasonable maximum width, but they just keep on going. I also expect that they would get narrower as I resize the browser down, but they don't.

Each card in a carousel is significantly wider than the widest single piece of text. They also do not get narrower as I resize the browser down, so I typically can't even see the ">" control without scrolling.

Thumbails aren't as bad, but they seem wider than necessary, and also do not get narrower as I resize the browser down.

Tested on Safari and Chrome on Mac.

billba:
Okay now they are resizing down, which is much better. But they are still very large. This makes the hero and thumbnails merely ugly, but in the case of carousel it means you don't get to see many of the cards, and that means you have to do a lot of unnecessary scrolling.

danmarshall:
Other than Carousel, most everything behaves well responsively even at very small sizes, like 100px wide. The Carousel is the bad player here, tracking this issue in #111

billba:
Agreed on splitting off Carousel.

I agree the cards look fine at small sizes. I'm suggesting that we may want to have a maximum width, because they just look bad when wide.

danmarshall:
I have a 320px max height for images, currently not setting a max width. I suppose it should be based on @media query breakpoints.

billba:
I don't think the images are the culprit here. Look at "hero" in testbed. Little image, wide card.

danmarshall:
@kevkor should we clamp down cards to a max-width like 640px or so?

IanBoggs:
Sorry to ask, but is this still being looked at?
I'm still seeing the carousel being very wide and moving the user entered messages out of view to the right

billba:
what do you mean about the user-entered messages? can you please post a screenshot and tell us which OS/browser?

this is on the list to fix but currently low priority

IanBoggs:
Basically, this screen shot shows before the carousel is show to the user
![webchat1](https://cloud.githubusercontent.com/assets/18186052/20970161/1f2c480c-bc84-11e6-8357-2b068e37d674.png)

As you can see, the text entered by the user is shown in blue on the right hand side of the window. Then once the carousel is show this happens

![webchat2](https://cloud.githubusercontent.com/assets/18186052/20970162/1f4eebf0-bc84-11e6-8e50-15a143d24527.png)

[...]

PR #282, opened 2017-01-14
Cards and Carousel fixes

danmarshall:
Merged WrappedActivities with ActivityView.
Fixed size of cards.
Added overflow measurement to carousel so that it respects the chat width.
Removed "table" css.
Add padding to messages to "autoscroll" mimicking vertical-align:bottom.


src/Chat.tsx
```diff
@@ -168,6 +168,21 @@ export const sendFiles = (dispatch: Dispatch<HistoryAction>, files: FileList, fr
...
+export const measure = { 
+    innerHeight: (el: HTMLElement) : number => {
+        var s = window.getComputedStyle(el);
+        var paddingTop = parseInt(s.getPropertyValue('padding-top'));
+        var paddingBottom = parseInt(s.getPropertyValue('padding-bottom'));
+        return el.offsetHeight - paddingTop - paddingBottom;
+    },
+    outerWidth: (el: HTMLElement) : number => {
+        var s = window.getComputedStyle(el);
+        var marginLeft = parseInt(s.getPropertyValue('margin-left'));
```
danmarshall: I suppose I could write a helper function to parse a propertyValue

danmarshall:
fixes #230, #133, #114

danmarshall:
@billba can you take a look at this PR? I'd like to move to the next batch :)

botchat.css
```diff
@@ -122,6 +116,11 @@
         position: relative;
     }
 
+    .wc-message-wrapper.carousel .wc-message {
```
danmarshall: this is correct - it is selecting a wc-message inside of a wrapper with the carousel layout

billba: but your code is generating a space between `.wc-message-wrapper` and `.carousel`

src/History.tsx
```diff
@@ -119,6 +119,26 @@ export const History = connect(
...
 )(HistoryContainer)
 
+const getComputedStyleValues = (el: HTMLElement, stylePropertyNames: string[]) => {
+    const s = window.getComputedStyle(el);
+    const result: { [key: string]: number } = {};
+    stylePropertyNames.forEach(name => result[name] = parseInt(s.getPropertyValue(name)));
+    return result;
+}
+
+const measure = {
```
billba: Please just make these two separate functions. We don't use this convention anywhere else in this codebase.

danmarshall: ok, I was preparing for the possibility of the other combinatorials (outerHeight, innerWidth...)

danmarshall: taking a cue from Jquery

botchat.css
danmarshall: DOM and CSS selectors are separate syntaxes. In DOM, className are indeed separated by a space character, but in CSS selectors they have no space character. Another example would be an ID - in the DOM it's an attribute e.g. id="foo" but in CSS it's #foo

danmarshall: (reprint- comment did't appear here)
DOM and CSS selectors are separate syntaxes. In DOM, className are indeed separated by a space character, but in CSS selectors they have no space character. Another example would be an ID - in the DOM it's an attribute e.g. id="foo" but in CSS it's #foo

PR #303, opened 2017-01-25
Adding youtube card

bnookala:
Looks like this: 
youtube-card

And uses the embeddable player with an iframe to drop it in: https://developers.google.com/youtube/player_parameters#Overview

### DIFF
--- a/botchat.css
+++ b/botchat.css
@@ -149,6 +149,10 @@
         max-width: 100%;
     }
 
+    .wc-message-content .video iframe {
+        border: 0;
+    }
+
     .wc-message-content audio, .wc-message-content video {
         max-width: 100%;
     }
--- a/src/Attachment.tsx
+++ b/src/Attachment.tsx
@@ -1,9 +1,14 @@
 import * as React from 'react';
+import * as URI from 'urijs';
+
 import { Attachment, Button } from 'botframework-directlinejs';
 import { renderIfNonempty, konsole } from './Chat';
 import { FormatState } from './Store';
 
 const regExpCard = /\^application\/vnd\.microsoft\.card\./i;
+
+const YOUTUBE_DOMAIN = "youtube.com";
+const YOUTUBE_SHORT_DOMAIN = "youtu.be";
 
 const buttons = (
     buttons: Button[],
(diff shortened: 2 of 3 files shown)


src/Attachment.tsx
```diff
@@ -13,6 +15,50 @@ const buttons = (
         { buttons.map((button, index) => <li key={ index }><button onClick={ () => onClickButton(button.type, button.value) }>{ button.title }</button></li>) }
     </ul>;
 
+const isYoutube = (src: string) => {
+    let youtubeUrl = new URI(src);
```
billba: Our convention in WebChat is to use `const` when possible. Please humor me here and below.

src/Attachment.tsx
```diff
@@ -13,6 +15,50 @@ const buttons = (
...
+                    .domain("youtube.com")
+                    .port("")
+                    .segment(["embed", embedId])
+                    .search(embedUrlQuery)
+                    .toString();
+
+    return <iframe
+        type="text/html"
+        src={ embedUrl } >
+    </iframe>
```
billba: I like my tags to line up. Will you either put this all on one line, or do:

```
return (
    <iframe
        ...
    </iframe>
);

File: src/Attachment.tsx [hunk 3]
```diff
@@ -13,6 +15,50 @@ const buttons = (
...
+
+    if ((youtubeUrl.domain() !== 'youtube.com')) {
+        return false;
+    }
+
+    if (embedId === undefined) {
+        return false;
+    }
+
+    return true;
```
billba: This is too wordy for my taste. How about we collapse this all into:

```
return youtubeUrl.domain() === "youtube.com" && youtubeUrl.search(true).v !== undefined;
```

src/Attachment.tsx
```diff
@@ -25,6 +71,10 @@ const Media = (props: {
     const { type, ... mediaProps } = props; // this allows us to keep 'type' out of the final HTML
     switch (type) {
         case 'video':
+            if (isYoutube(mediaProps.src)) {
+                return <Youtube { ... mediaProps } />
+            }
+
```
billba: It bugs me that we end up calling URI twice on the same string. And if we add future special cases, we'll do it more.

How about we turn this into a function called Video which will look at the URI for each special-case domain in turn, returning the appropriate `<iframe>` as appropriate, or returning a `<video>` tag if none apply. All that logic can be in one function.

botchat.css
```diff
@@ -149,6 +149,10 @@
         max-width: 100%;
     }
 
+    .wc-message-content .video iframe {
+        border: 0px;
```
danmarshall: 0 is ok, without the px units

danmarshall:
we don't necessarily need to include another library - have a look at this: http://stackoverflow.com/questions/3213531/creating-a-new-location-object-in-javascript

src/Attachment.tsx
```diff
@@ -13,6 +15,50 @@ const buttons = (
         { buttons.map((button, index) => <li key={ index }><button onClick={ () => onClickButton(button.type, button.value) }>{ button.title }</button></li>) }
     </ul>;
 
+const isYoutube = (src: string) => {
+    let youtubeUrl = new URI(src);
+    let queryParams = youtubeUrl.search(true);
+    let embedId = queryParams.v;
+
+    if ((youtubeUrl.domain() !== 'youtube.com')) {
```
danmarshall: Youtube also uses their own short url format http://youtu.be

src/Attachment.tsx
danmarshall: also, why compare so strictly against ```undefined``` - we can also just say ```!embedId```

danmarshall:
@jecristi can you review the visual design of this? It may be that we also want to display the URL in addition to the player.

danmarshall:
@bnookala nice work, thanks!

src/Attachment.tsx
danmarshall: I tend to agree, since it may become a verbose block:
```
 if (isYouTube...)
   return <YouTube...>
 if (isVimeo...)
   return <Vimeo...>
 if (isFacebook...)
   return <Facebook...>
 if (isInstagram...)
   return <Instagram...>
```

src/Attachment.tsx
bnookala: 👍

src/Attachment.tsx
bnookala: 👍

src/Attachment.tsx
bnookala: 👍

botchat.css
bnookala: 👍

src/Attachment.tsx
billba: @danmarshall You really get me.

src/Attachment.tsx
bnookala: 👍

src/Attachment.tsx
bnookala: 👍

bnookala:
@danmarshall, @billba once more for good measure, incorporating your suggestions.

@danmarshall I'm a bit partial to using the URI.js library it provides many convenience methods to things a`Location` object doesn't: ie: building/decomposing the querystring from or into/an object, building up a directory structures, etc. I imagine as we think about adding more embeddable attachments, it will prove to be quite useful.

billba:
This turned out great, thanks!

PR #308, opened 2017-01-27
cleanup

billba:


src/Attachment.tsx
```diff
@@ -22,23 +22,22 @@ const Youtube = (props: {
...
-                    .subdomain("")
-                    .port("")
-                    .segment(["embed", props.embedId])
-                    .search(embedUrlQuery)
-                    .toString();
-
-    return <iframe type="text/html" src={ embedUrl }></iframe>
-}
+}) =>
+    <iframe
```
bnookala: Should this be `return <iframe …`?

billba: you can either do `foo => { return bar; }` or` foo => bar;` - they are equivalent;

PR #316, opened 2017-01-31
Removing URI.js

bnookala:
Removing the URI.js dependency, which was a bit heavier weight than necessary.

### DIFF
--- a/src/Attachment.tsx
+++ b/src/Attachment.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import * as URI from 'urijs';
 
 import { Attachment, Button } from 'botframework-directlinejs';
 import { renderIfNonempty, konsole } from './Chat';
@@ -8,8 +7,50 @@ import { FormatState } from './Store';
 const regExpCard = /\^application\/vnd\.microsoft\.card\./i;
 
 const YOUTUBE_DOMAIN = "youtube.com";
+const YOUTUBE_WWW_DOMAIN = "www.youtube.com";
 const YOUTUBE_SHORT_DOMAIN = "youtu.be";
+const YOUTUBE_WWW_SHORT_DOMAIN = "www.youtu.be";
 const VIMEO_DOMAIN = "vimeo.com";
+const VIMEO_WWW_DOMAIN = "www.vimeo.com";
+
+interface VideoEmbedQuery {
+    loop?: number;
+    autoplay?: number;
+    modestbranding?: number;
+    title?: number;
+    byline?: number;
+    portait?: number;
+    badge?: number;
+    [propName: string]: number;
+}
+
+export const queryParams = (
+    src: string
+) => {
+    const queryObject = {};
+
+    src.substr(1)
+        .split('&')
+        .forEach(field => {
+            const keyValue = field.split('=');
+            queryObject[decodeURIComponent(keyValue[0])] = decodeURIComponent(keyValue[1]);
+        });
+
+    return queryObject;
+}
+
+const buildUrl = (
+    src: string,
+    query: VideoEmbedQuery,
+) => {
+    return [
+        src,
+        '?',
+        Object.keys(query)
+            .map(key => encodeURIComponent(key) + '=' + encodeURIComponent(query[key].toString()))
+            .join('&')
+    ].join('');
+}
 
 const buttons = (
     buttons: Button[],
@@ -26,17 +67,15 @@ const Youtube = (props: {
 }) =>
     <iframe
         type="text/html"
-        src={ new URI()
-            .domain(YOUTUBE_DOMAIN)
-            .subdomain("")
-            .port("")
-            .segment(["embed", props.embedId])
-            .search({
-                "modestbranding": 1,
-                "loop": props.loop ? 1 : 0,
-                "autoplay": props.autoPlay ? 1 : 0
-            })
-            .toString()
+        src={
+            buildUrl(
+                `https://${YOUTUBE_DOMAIN}/embed/${props.embedId}`,
+                {
+                    modestbranding: 1,
+                    loop: props.loop ? 1 : 0,
+                    autoplay: props.autoPlay ? 1 : 0
+                }
+            )
         }
     />;
 
@@ -47,20 +86,18 @@ const Vimeo = (props: {
 }) =>
     <iframe
         type="text/html"
-        src={ new URI()
-            .domain(VIMEO_DOMAIN)
-            .subdomain("player")
-            .port("")
-            .segment(["video", props.embedId])
-            .search({
-                "title": 0,
-                "byline": 0,
-                "portrait": 0,
-                "badge": 0,
-                "autoplay": props.autoPlay ? 1 : 0,
-                "loop": props.loop ? 1 : 0
-            })
-            .toString()
+        src={
+            buildUrl(
+                `https://player.${VIMEO_DOMAIN}/video/${props.embedId}`,
+                {
+                    title: 0,
+                    byline: 0,
+                    portrait: 0,
+                    badge: 0,
+                    autoplay: props.autoPlay ? 1 : 0,
+                    loop: props.loop ? 1 : 0
+                }
+            )
         }
     />;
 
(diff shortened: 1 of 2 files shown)


src/Attachment.tsx
```diff
@@ -72,19 +112,25 @@ const Video = (props: {
     onLoad?: () => void,
     onClick?: () => void,
 }) => {
-    const src = new URI(props.src);
-    const domain = src.domain();
+    const url = document.createElement('a');
+    url.href = props.src;
+
+    const queryObject = queryStringToObject(url.search);
+    const pathSegments = url.pathname.substr(1).split('/');
```
billba: what does the `substr(1)` do here?

src/Attachment.tsx
```diff
@@ -26,17 +60,20 @@ const Youtube = (props: {
...
+                    "https://",
+                    YOUTUBE_DOMAIN,
+                    "/embed/",
+                     props.embedId
+                ].join(''),
+                {
+                    "modestbranding": 1,
+                    "loop": props.loop ? 1 : 0,
+                    "autoplay": props.autoPlay ? 1 : 0
+                }
```
billba: any reason you're quoting these field names here and below? unless necessary I'd prefer to  have them unquoted.

src/Attachment.tsx
```diff
@@ -26,17 +60,20 @@ const Youtube = (props: {
...
-            })
-            .toString()
+        src={
+            buildUrl(
+                [
+                    "https://",
+                    YOUTUBE_DOMAIN,
+                    "/embed/",
+                     props.embedId
+                ].join(''),
```
billba: I'd rather use string interpolation here and below, as it's much clearer (and, in latest browsers, now as faster or faster than array join)., e.g. \``https://${YOUTUBE_DOMAIN}/embed/${props.embedId}`\`

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
...
+    for (const key in query) {
+        if (!query.hasOwnProperty(key)) {
+            continue
+        }
+        queryObject.push(key + '=' + query[key]);
+    }
+
+    builtUrl.push(queryObject.join('&'));
+    return builtUrl.join('');
+}
```
billba: Let's be more functional and also use the spread operator:
```
=> [
    src,
    '?',
    ... Object.keys(query)
        .map(key => key + '=' + query[key])
        .join('')
].join('')
```

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
...
+
+const buildUrl = (
+    src: string,
+    query: Object,
+): string => {
+    const queryObject = [];
+    const builtUrl = [src, '?'];
+
+    for (const key in query) {
+        if (!query.hasOwnProperty(key)) {
```
billba: if you use `for... of` or `Object.keys(..)` then you don't need to filter on `hasOwnProperty`

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
...
+        const keyValue = field.split('=');
+        queryObject[decodeURIComponent(keyValue[0])] = decodeURIComponent(keyValue[1] || '');
+    }
+
+    return queryObject;
+}
+
+const buildUrl = (
+    src: string,
+    query: Object,
```
billba: You can get more safety from TypeScript if you define a type for this parameter, like:

```
interface Query {
    [name: string]: string
}
```

Then you can do `query: Query`

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
...
+    const queryObject = {};
+    const fields = src.substr(1).split('&');
+
+    for (const field of fields) {
+        const keyValue = field.split('=');
+        queryObject[decodeURIComponent(keyValue[0])] = decodeURIComponent(keyValue[1] || '');
+    }
+
+    return queryObject;
+}
```
billba: would prefer this be more functional:

```
    src
    .split('&')
    .forEach(field => ... )
```

src/Attachment.tsx
```diff
@@ -72,19 +112,25 @@ const Video = (props: {
     onLoad?: () => void,
     onClick?: () => void,
 }) => {
-    const src = new URI(props.src);
-    const domain = src.domain();
+    const url = document.createElement('a');
```
billba: This was clever.

src/Attachment.tsx
```diff
@@ -72,19 +112,25 @@ const Video = (props: {
...
+    url.href = props.src;
+
+    const queryObject = queryStringToObject(url.search);
```
billba: again, let's call this something more meaningful than "queryObject"

src/Attachment.tsx
danmarshall: I believe it removes the leading slash character, so that the array from .split does not have an empty first element.

danmarshall:
Although we aren't using URI.js - I would still like to see a class called URI, with the same API as URI.js (containing only the minimal features an implementation.)

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
 const regExpCard = /\^application\/vnd\.microsoft\.card\./i;
 
 const YOUTUBE_DOMAIN = "youtube.com";
+const YOUTUBE_WWW_DOMAIN = "www.youtube.com";
 const YOUTUBE_SHORT_DOMAIN = "youtu.be";
+const YOUTUBE_WWW_SHORT_DOMAIN = "www.youtu.be";
 const VIMEO_DOMAIN = "vimeo.com";
+const VIMEO_WWW_DOMAIN = "www.vimeo.com";
+
+const queryStringToObject = (
```
bnookala: 👍

billba:
@danmarshall Why? That will inevitably end up being bigger than what's here.

src/Attachment.tsx
billba: Ah I see, pathname starts with a slash. Makes sense, thanks.

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
...
+    src: string
+): any => {
```
bnookala: 👍

src/Attachment.tsx
```diff
@@ -8,8 +7,43 @@ import { FormatState } from './Store';
...
+    const queryObject = {};
+    const fields = src.substr(1).split('&');
```
bnookala: The Location.search interface reutrns a '?' prefix: https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/search

billba: Go it, thanks.

danmarshall:
@billba I'm not saying the entrire API should be implemented, I'm suggesting that:
* only implement the stuff we actually need
* use the same "API surface" signatures for the stuff we implemented

billba:
Why?

danmarshall:
The template string `https://${YOUTUBE_DOMAIN}/embed/${props.embedId}` is nice, but it also might be misused such that a developer might skip the encoding of parameters if they `https://${YOUTUBE_DOMAIN}/embed/${props.embedId}?foo=${barValue}`

src/Attachment.tsx
bnookala: 👍

billba:
@danmarshall is that any different from doing this as a `[...].join('')`?

src/Attachment.tsx
bnookala: 👍

src/Attachment.tsx
bnookala: 👍

src/Attachment.tsx
bnookala: 👍

danmarshall:
@billba in the array.join approach, the developer will be adding `https://` and slashes surrounding segements such as `/embed/`, I was thinking that a small class could do this job, as well as the encoding of parameters. I do like the idea of having a testable URI rather that ad-hoc string building. I think it can prevent a lot of bugs.

src/Attachment.tsx
bnookala: 👍

src/Attachment.tsx
bnookala: 👍

billba:
I agree that you could eliminate a little duplication (and thus bugs) with a function. I think trying to implement a subset of the URI API is overkill. But I'm open to the idea. My proposal is to have @bnookala implement this using the simple string interpolation as I proposed. Then if you want to take a swing at extracting the URI logic into a URI subset, we can see how much extra code it adds.

src/Attachment.tsx
bnookala: @billba this would probably change the function to return a Promise as `forEach` isn't synchronous. I'm looking around at the src, and don't see any use cases of Promises - is it ok to use? Alternatively, I can use a callback, but that might be stylistically not ok.

billba: forEach is synchronous.

bnookala: 👍

bnookala:
@billba, @danmarshall: once more, with feeling 🙃

Issue #271, opened 2017-01-09
Support more action types


danmarshall:
We support a few action types in [BotConnection.ts](https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/BotConnection.ts#L19) and [History.tsx](https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/History.tsx#L55)

There are a few more ActionTypes in the SDKs ( [CSharp](https://github.com/Microsoft/BotBuilder/blob/master/CSharp/Library/Microsoft.Bot.Connector/ActionTypes.cs#L9) | [Node](https://github.com/Microsoft/BotBuilder/blob/master/Node/core/src/cards/CardAction.ts#L75) ) that have yet to be implemented:
* call
* playAudio
* playVideo
* showImage
* downloadFile

Some of these seem reasonable.

billba:
Sounds great. Need specs!

eanders-MS:
I think they should all just open the URL in a new tab.

danmarshall:
In #318 it appears that they are opening in the same tab.

eanders-MS:
Are you certain? They opened in a new tab for me in local testing.

PR #318, opened 2017-01-31
handle additional card action types

eanders-MS:


eanders-MS:
For issue https://github.com/Microsoft/BotFramework-WebChat/issues/271

eanders-MS:
My editor likes to automatically trim trailing whitespace.

src/History.tsx
```diff
@@ -64,7 +64,12 @@ class HistoryContainer extends React.Component<Props, {}> {
...
                 break;
 
+            case "call":
             case "openUrl":
+            case "playAudio":
+            case "playVideo":
+            case "showImage":
+            case "downloadFile":
             case "signin":
                 window.open(value);
```
danmarshall: Hmmm - this doesnt open in a new tab. It will open in the current window.

eanders-MS: When I tested locally all opened in a new tab.

danmarshall: Could be true for the mime types that don't resolve to ```text/*``` 
But for ```case "openUrl":``` and ```case "signin":``` these might display in the current window if they are html

eanders-MS:
According to window.open docs, it always opens in a new window (or tab, I suppose):
http://www.w3schools.com/jsref/met_win_open.asp
"The open() method opens a new browser window."

src/History.tsx
eanders-MS: According to window.open docs, it always opens in a new window (or tab, I suppose):
http://www.w3schools.com/jsref/met_win_open.asp
"The open() method opens a new browser window.

danmarshall: oic - ```_blank``` is default. tx for checking :)

PR #324, opened 2017-02-02
Shell tests

billba:


billba:
(also made it so you can set a specific value for sendTyping in case we decide to change the default)

PR #336, opened 2017-02-08
full tap support

billba:


src/Attachment.tsx
```diff
@@ -148,37 +125,56 @@ const text = (text: string) => renderIfNonempty(text, text => <p>{ text }</p>);
...
             return (
-                <div className='wc-card thumbnail'>
+                <div className='wc-card thumbnail' onClick={ onCardAction(attachment.content.tap) }>
                     { title(attachment.content.title) }
-                    { attachedImage(attachment.content.images, props.onImageLoad) }
+                    { attachedImage(attachment.content.images) }
                     { subtitle(attachment.content.subtitle) }
                     { text(attachment.content.text) }
-                    { buttons(attachment.content.buttons, props.onClickButton) }
+                    { buttons(attachment.content.buttons) }
```
danmarshall: Will a button click bubble up and fire the onClick above?

billba: No, it stops the propagation.

PR #338, opened 2017-02-10
Chat width via props

danmarshall:
Max width of carousel is determined by the History and then passed down as props.
Resize behavior is now an option which may be either 'none', 'window' or 'detect'.


billba:
Phew!

src/Carousel.tsx
```diff
@@ -23,14 +24,14 @@ export class Carousel extends React.Component<CarouselProps, CarouselState> {
...
-    private resizeListener = () => this.resize();
-    private scrollEventListener =() => this.onScroll();
+    private scrollEventListener = () => this.onScroll();
     private scrollAllowInterrupt = true;
 
     constructor(props: CarouselProps) {
         super(props);
 
         this.state = {
+            contentWidth: undefined,
```
MoisesFlores22: I merge this pr to my master branch and is broken @billba

danmarshall: Can you provide the build error?

MoisesFlores22: screen shot 2017-03-13 at 12 20 18 pm

billba: You need to `npm install` to get the latest version of TypeScript and the latest React type declarations.

MoisesFlores22: thanks

PR #344, opened 2017-02-17
reduce footprint

billba:


billba:
each rxjs operator costs us in footprint, so I got rid of `mapTo`, which is just a convenience variety of `map`. I did this in both DirectLineJS and WebChat.d

PR #352, opened 2017-02-27
fix samples

billba:


samples/backchannel/index.html
```diff
@@ -65,7 +65,7 @@ <h2 style="font-family:Segoe UI;">Type a color into the WebChat!</h2>
                 secret: params['s'],
                 token: params['t'],
                 domain: params['domain'],
-                webSocket: params['webSocket'] // defaults to true
+                webSocket: params['webSocket'] && params['webSocket'] === "true" // defaults to true
```
danmarshall: Or you can do the inline boolean conversion !!params['webSocket']

billba: No because I want to pass in `undefined` in the (common) case where the `webSocket` parameter doesn't exist.

danmarshall: Does the && convert to boolean?

billba: Anyway, `!!"false" === true` which is not what we want.

billba: No.

danmarshall: Hold up. So what if someone passes in "false" on the url? Do you get the desired behavior?

billba: Yes.

PR #375, opened 2017-03-14
CarouselAttachments should not update during resize

danmarshall:
props.format contains UI stuff like chatWidth, which changes in a resize.


billba:
The whole point of `props.format` is that it contains properties which affect the rendering of the content. For instance, if the locale changes, the strings change, and we should re-render the attachments.

At the time I thought you had said that carouselAttachments should re-render on resize because of media queries, but now I think maybe media queries don't affect re-rendering, just measuring.

After sitting on this for a few weeks I think we should take `chatHeight` and `chatWidth` out of props.format and give them their own prop (I'm thinking `size: { height, width }`). That will make it easier to decide when to re-render. Probably most components won't even take that prop.

I'm working on a refactor of Carousel right now that simplifies its logic quite a bit. It would make sense to include this proposed change in that refactor, if you're agreeable.

If so, let's pull that part of this PR and just merge in the SVG changes.

danmarshall:
Before you refactor Carousel, I've already had to do the same. SuggestedActions needs to be horizontally scrollable. I have a decent working implementation my [actions branch](https://github.com/danmarshall/BotFramework-WebChat/tree/actions). I've refactored into an HScroll component.

Issue #374, opened 2017-03-14
Outgoing text is rendered as Markdown


maba4891:
Hi,

in my node.js bot I provide a simple builder.Prompts.text to ask the user for a date (I know that Prompts.time exists but since it does not seem to work with German dates yet I use Prompts.text for now and parse the date with the datejs module).

The problem is that if the user enters the date in a German format (e.g. "12. Februar 2017") and presses enter he sees "1. Februar 2017" as his input. The same is true for other dates (e.g. "20. August 2016") which are printed in a similar way like "1. August 2016". This also happens when I provide a date with a dot as an input to any other type of prompt.

Somehow the dot seems to be the problem here, 12 Februar 2017 works just fine (as does 12.2.2017). This problem seems to be only affecting the Emulator and the WebChat, on Skype and Facebook Messenger everything seems to be fine.

The funny thing is also that internally the right date is saved (even in the Emulator and WebChat), e.g. 12. Februar 2017 even if 1. Februar 2017 is printed (I save the input in session.userData). So it seems the problem is just the way the input is printed.

Any clue what is happening here?

Originally posted at https://github.com/Microsoft/BotBuilder/issues/2409 but moved to WebChat

billba:
Weird. This has nothing to do with `Prompts` - it happens whenever you enter text in any context. I have changed the title of this issue accordingly.

This is almost certainly a bug in our Markdown parser.

billba:
I realized what is happening. Markdown considers any text of the form "number. text" to be a numbered list. It will actually ignore the number and just start a `<ul>`.

I think this means we need to treat outgoing text as plain text, but we should run this up the flagpole.

billba:
Let's treat outgoing text as plain text. We could certainly add `textFormat` as a prop to `Chat`, defaulting to 'plain', but I would hate to add a proliferation of options when no one is asking for it.

billba:
Fixed by https://github.com/Microsoft/BotFramework-WebChat/pull/377

alok-kumar:
I am using latest botchat CDN but unable to render HTML with iFrame in output. Example: Expected output is "Domestic travel related document</br><iframe src='https://onedrive.live.com/embed?resid=SDFSFSDSDF%211062' width='200' height='346' frameborder='0' scrolling='no'></iframe>"

[...]

PR #377, opened 2017-03-15
Format parameter

GeekTrainer:
Added a parameter to handle [markdown issue](https://github.com/Microsoft/BotFramework-WebChat/issues/374) for sent messages. Set the default to be plain rather than markdown.


src/Store.ts
```diff
@@ -56,6 +56,7 @@ export const shell: Reducer<ShellState> = (
 
 export interface FormatState {
     locale: string,
+    messageTextFormat: string,
```
billba: this should go in `ConnectionState`, with `user`.

src/Store.ts
```diff
@@ -76,11 +77,15 @@ export type FormatAction = {
 } | {
     type: 'Set_Measurements',
     carouselMargin: number
+} | {
+    type: 'Set_Text_Format_Option',
+    messageTextFormat: string
```
billba: if it goes in `ConnectionState` I think you can piggyback on an existing action.

GeekTrainer:
Changed the default to be plain.

GeekTrainer:
And again

PR #383, opened 2017-03-17
Master

billba:


billba:
This introduces a new React component called `Slider` which contains `History` and `SuggestedActions`, thus separating concerns between those two. `Slider` owns all the `CardAction` handling for both.

`Slider` is a connected component, which is to say that it reacts to changes to app state and passes properties to `History` (which becomes strictly a rendering component) and `SuggestedActions` (which already was).

I also took the opportunity to clean up history.tsx a smidge by liberal use of the JSX object spread operator.

This paves the way to potentially moving the suggested actions state machine into Redux, but it is worthwhile just by itself.

src/Slider.tsx
```diff
@@ -0,0 +1,99 @@
...
+function suggestedActions(activities: Activity[]) {
+    if (!activities || activities.length === 0)
+        return;
+    const lastActivity = activities[activities.length - 1];
+    if (!lastActivity || lastActivity.type !== 'message' || !lastActivity.suggestedActions || lastActivity.suggestedActions.length === 0)
+        return;
+    return lastActivity.suggestedActions;
+}
+
+export const Slider = connect(
```
danmarshall: The name Slider has a UI control connotation. But I don't have a better suggestion...

billba: Agreed it is not a great name. I also couldn't think of a better name.

src/Slider.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const onCardAction = props.onCardAction(props.sendMessage);
+    return (
+        <div className={ props.actions ? 'show-actions' : '' }>
+            <History { ... props } onCardAction={ onCardAction }/>
+            <SuggestedActions actions={ props.actions } onCardAction={ onCardAction }/>
+        </div>
+    );
+}
+
+function suggestedActions(activities: Activity[]) {
```
danmarshall: I thought you'd be using a fancier way to get these by now ;)

billba: Baby steps.

src/Slider.tsx
danmarshall: How about MessageTrack ?

danmarshall: or MessageTrail

billba: Christopher and I have voted for HistoryAndSuggestedActions which is uninspiring but descriptive. Last chance for you to veto...

billba:
This update is a better factoring of the same idea. Now `SuggestedActions` includes both the containing `div` and the actions bar itself, and renders `History` using `props.children`, which combines the clean separation of concerns of the previous approach with the efficiency of the original implementation. @danmarshall if you approve go ahead and merge, after which I will take on the Reduxification of SuggestedActions.

PR #394, opened 2017-03-21
Markdown renderer fixes

eanders-MS:
- Unescape &quot in inline-code blocks.
- Don't use fancy punctuation for things like quotes.

### DIFF
--- a/src/FormattedText.tsx
+++ b/src/FormattedText.tsx
@@ -118,7 +118,7 @@ class ReactRenderer implements MarkedRenderer {
     /// MarkedRenderer overrides
 
     code(code: string, language: string): string {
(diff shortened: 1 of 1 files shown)


src/FormattedText.tsx
```diff
@@ -41,7 +41,7 @@ const renderMarkdown = (
         sanitize: false,
         smartLists: true,
         silent: false,
-        smartypants: true,
+        smartypants: false,
```
danmarshall: not so smart, that smartypants!

eanders-MS: Too smart for its own good.

billba: story of my life

PR #396, opened 2017-03-23
cleaner React/Redux

billba:


billba:
Two real changes here:

* If you use `React.PureComponent` instead of `React.Component` it provides its own `shouldComponentUpdate` which returns true if any of the props have changed. So I replaced two instances of `React.Component`.
* I converted `sendMessage` and `sendFiles` to Action Creators, and that simplified the react-redux code in `History`, `Shell`, and `MessagePane`

danmarshall:
Nice.

danmarshall:
So, one thing is that dispatch is now hidden away. I was needing to dispatch and now I need to add the plumbing back in.

Issue #380, opened 2017-03-16
watch should rebuild .css files


billba:
Right now `npm run watch` won't notice if a .scss file changes. Folks like me who always use `watch` instead of `build` get errors like I did today when I updated to the latest sources.

billba:
Fixed by #397.

PR #397, opened 2017-03-24
Hide suggestedActions when clicked for any actionType

danmarshall:
and a little SCSS watch script


src/Store.ts
```diff
@@ -333,6 +340,14 @@ export const history: Reducer<HistoryState> = (
...
             };
 
+        case 'Clicked_SuggestedActions':
+
+            action.clickedActivity.clicked = true;
+
+            return {
+                ... state
+            };
+
```
billba: No no no no no no no no no no no no no no no - the first rule of Redux Club is you don't mutate state. It's also the second rule. (Yes, I did just tweet that.) We need to generate a new activity here, and also update `activities` and maybe `selectedActivity`. The closest model is `'Send_Message_Succeed'`. I think this is right:

```typescript
        case 'Take_SuggestedAction': {
            const i = state.activities.findIndex(activity => activity === action.trackedMessage);
            const activity = state.activities[i];
            const newActivity = {
                ... activity,
                suggestedActionTaken: true
            };
            return {
                ... state,
                activities: [
                    ... state.activities.slice(0, i),
                    newActivity,
                    ... state.activities.slice(i + 1)
                ],
                selectedActivity: state.selectedActivity === activity ? newActivity : state.selectedActivity
            }
```

You'll also need to add `'Take_SuggestedAction'` to the list of action types in `updateSelectedActivity` below.

Yeah, it's a lot more work but it's an important rule.

(I could imagine adding a helper function for the array manipulation since we do it in a few places in this reducer.)

src/MessagePane.tsx
```diff
@@ -58,20 +62,28 @@ class SuggestedActions extends React.Component<MessagePaneProps, {}> {
 
 }
 
-function suggestedActions(activities: Activity[]) {
+function activityWithSuggestedActions(activities: Activity[]) {
```
billba: for consistency, let's either call this `activityWithActions` or call the property `activityWithSuggestedActions`.

src/MessagePane.tsx
```diff
@@ -1,19 +1,20 @@
 import * as React from 'react';
-import { Activity, CardAction, User } from 'botframework-directlinejs';
-import { ChatState } from './Store';
-import { Dispatch, connect } from 'react-redux';
+import { Activity, CardAction, User, Message } from 'botframework-directlinejs';
```
danmarshall: au contraire

src/MessagePane.tsx
```diff
@@ -58,20 +62,28 @@ class SuggestedActions extends React.Component<MessagePaneProps, {}> {
...
+        activityWithActions: activityWithSuggestedActions(state.history.activities),
         doCardAction: doCardAction(state.connection.botConnection, state.connection.user, state.format.locale),
-    }), {
-        sendMessage
+    }),
+    {
+        onClickSuggestedActions: (activity: TrackedActivity) => {
+            return { type: 'Clicked_SuggestedActions', clickedActivity: activity } as HistoryAction;
+        },
+        sendMessage: sendMessage
```
danmarshall: I'm still learning these crazy new spread operators

src/MessagePane.tsx
billba: Listen you can't expect me to look with my eyes.

danmarshall:
fixes #380

src/MessagePane.tsx
danmarshall: How does GitHub know that we're done reviewing?

billba: When it's merged, I guess.

src/MessagePane.tsx
```diff
@@ -25,15 +27,8 @@ class SuggestedActions extends React.Component<MessagePaneProps, {}> {
     }
 
     actionClick(e: React.MouseEvent<HTMLButtonElement>, cardAction: CardAction) {
-
-        //click is only valid if there are props.actions
```
danmarshall: This was needed to "debounce" the handler, so that the button couldn't be clicked twice.

src/MessagePane.tsx
```diff
@@ -16,7 +16,9 @@ export interface MessagePaneProps {
 const MessagePaneView = (props: MessagePaneProps) =>
     <div className={ classList('wc-message-pane', props.activityWithSuggestedActions && 'show-actions' ) }>
         { props.children }
-        <SuggestedActions { ... props }/>
+        <div className="wc-suggested-actions">
+            <SuggestedActions { ... props }/>
```
danmarshall: Architecturally, I'd like components to own their own class name.

Issue #363, opened 2017-03-08
focus should return to the text entry box.


LarsLiden:
After the user clicks on a button the focus should return to the text entry box.  Currently after you type on a button you have to click on the text entry box to type your next input.  It’s a bit clunky and often makes buttons less convenient than just typing the response as you have to reach for the mouse and click twice

billba:
Fixed by #401

danmarshall:
This behavior as prescribed causes issues on mobile devices, where focusing the textbox causes the native keyboard to slide up.

Also, usability/accessibility may be violated because the tab order is interrupted.

When a button has focus, the user may trigger a click by:
* pressing Enter
* pressing the spacebar

The user may de-focus by pressing Tab.

So I am proposing that we have a keypress listener in the MessagePane, that if a key was pressed that is not Enter, Space or Tab, we send this keypress event character to the text entry box, and then focus it. This would by nature exclude mobile device's on-screen keyboards.

compulim:
I am going to disable JavaScript-based `setFocus` on these scenarios:
* Click/tap send button
* Click/tap action button
* Click/tap upload file button, then select a file

It will not affect the following scenario:
* Press ENTER after typing a message, the focus will continue to stay inside the textbox

And then implement @danmarshall suggestion by adding a `onKeyPress`/`onKeyDown` and then putting the focus back to the textbox.

danmarshall:
@compulim - I found a new regression with the key handling. Looks like <kbd>HOME</kbd>, <kbd>END</kbd>, page up/down and arrow up/down keys do not scroll the chat.

PR #401, opened 2017-03-28
Roll back "cleanup"

billba:


billba:
Also (sorry) includes really hacky fix for #363

PR #407, opened 2017-03-30
id -> class for text input

billba:


src/History.tsx
```diff
@@ -99,6 +101,11 @@ export class HistoryView extends React.Component<HistoryProps, {}> {
     // 2. To determine the margins of any given carousel (we just render one mock activity so that we can measure it)
     // 3. (this is also the normal re-render case) To render without the mock activity
 
+    private doCardAction(type: string, value: string) {
```
danmarshall: Every time we flatten this, it expands again! :P

billba: Truth

PR #412, opened 2017-04-06
Added es-es (Spanish) language

SantiEspada:
Hi! I've added Spanish language and also changed the tab to 4 spaces (to be like the rest of the file) on ru-ru language.


billba:
Thank you!

PR #414, opened 2017-04-07
update dependencies & build

billba:
Now uses Webpack 2.

Webpack build is now separate from the regular TypeScript build,
opening the door for tree shaking.

Resultant bundle is a little smaller too!

### DIFF
--- a/webpack.production.config.js
+++ b/webpack.production.config.js
@@ -1,13 +1,18 @@
 var webpack = require('webpack');
 
 module.exports = {
-    entry: {
-        app: ["./built/BotChat.js"],
-    },
+    entry: "./src/BotChat.ts",
     output: {
         libraryTarget: "umd",
         library: "BotChat",
         filename: "./botchat.js",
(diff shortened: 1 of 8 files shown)


webpack.production.config.js
```diff
@@ -22,20 +27,15 @@ module.exports = {
...
     ],
 
-    resolve: {
-        extensions: ["", ".webpack.js", ".web.js", ".js"]
-    },
-
     module: {
+        rules: [
+            // All files with a '.ts' or '.tsx' extension will be handled by 'awesome-typescript-loader'.
+            { test: /\.tsx?$/, loader: "awesome-typescript-loader" },
```
danmarshall: why should there be a TypeScript loader in production?

billba: Honestly I'm not sure I understand webpack enough to answer your question. This is copied from some references in how to get webpack to bundle TypeScript directly from sources (instead of from the built files).

PR #420, opened 2017-04-10
set noImplicitAny to true

danmarshall:


danmarshall:
This could / should help us from untyped bugs.

billba:
I keep meaning to do this!

PR #424, opened 2017-04-11
Cleanup typing

billba:
Generic type applies, so DRY!


danmarshall:
As long as the compiler is happy :)

PR #433, opened 2017-04-21
Using Markdown-it library

reyesrico:


src/FormattedText.tsx
```diff
@@ -29,206 +28,18 @@ const renderPlainText = (text: string) => {
...
-    }
-
-    image(href: string, title: string, text: string): string {
-        href = this.unescapeAndSanitizeLink(href);
-        if (!href)
-            return '';
-        return this.addElement(<img key={this.key++} onLoad={ () => this.onImageLoad() } {...{ src: href, title, alt: text }} />);
+namespace MarkdownItSing{
+    export function render(markdownOptions, src){
+        const md = new MarkdownIt(markdownOptions);
```
danmarshall: looks like this is still creating a new one for every render

PR #443, opened 2017-05-02
use IDoCardAction everywhere

danmarshall:
IDoCardAction, YouDoCardAction, WeAllDoCardAction


billba:
Wait how is this different from #442?

danmarshall:
It's the same, just found more places

danmarshall:
ok, the rabbit hole just went deeper, I added CardActionTypes to the signature. Massive intellisense ensues.

billba:
It's rabbits all the way down.

PR #456, opened 2017-05-10
add option to enable / disable autofocus of input when a button is clicked

macrozone:
if a user clicks on a card button, the input get focused, which opens the keyboard on mobile device.
This can be annoying if you use a lot of confirms and choices.

you can now set setFocusOnCardActionClick to false on the `<Chat>`-Component to disable this behaviour.

The caveat is that the chat no longer scrolls to bottom, when a button is clicked when this option is false.

Feedback welcome!

### DIFF
--- a/src/Chat.tsx
+++ b/src/Chat.tsx
@@ -26,7 +26,8 @@ export interface ChatProps {
     selectedActivity?: BehaviorSubject<ActivityOrID>,
     sendTyping?: boolean,
     formatOptions?: FormatOptions,
-    resize?: 'none' | 'window' | 'detect'
+    resize?: 'none' | 'window' | 'detect',
+    setFocusOnCardActionClick?: boolean, // defaults to true
 }
 
 export const sendMessage = (text: string, from: User, locale: string) => ({
@@ -54,7 +55,9 @@ import { MessagePane } from './MessagePane';
 import { Shell } from './Shell';
 
 export class Chat extends React.Component<ChatProps, {}> {
-
+    public static defaultProps: Partial<ChatProps> = {
+         setFocusOnCardActionClick: true
+    };
     private store = createStore();
 
     private botConnection: IBotConnection;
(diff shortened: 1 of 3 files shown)


src/History.tsx
```diff
@@ -98,7 +99,9 @@ export class HistoryView extends React.Component<HistoryProps, {}> {
     // 3. (this is also the normal re-render case) To render without the mock activity
 
     private doCardAction(type: CardActionTypes, value: string) {
-        this.props.setFocus();
+        if(this.props.setFocusOnCardActionClick) {
```
danmarshall: Perhaps this conditional can be moved above

src/Chat.tsx
```diff
@@ -176,10 +179,14 @@ export class Chat extends React.Component<ChatProps, {}> {
 
         return (
             <Provider store={ this.store }>
+
                 <div className="wc-chatview-panel" ref={ div => this.chatviewPanel = div }>
                     { header }
                     <MessagePane setFocus={ () => this.setFocus() }>
-                        <History setFocus={ () => this.setFocus() }/>
+                        <History
+                          setFocusOnCardActionClick={this.props.setFocusOnCardActionClick}
```
danmarshall: If we add the conditional here, we won't need the additional property on History

macrozone: I also thought about that, but from looking at this piece of code, you don't know whether History might need this.setFocus for someting else. That's why I dediced to use a rather verbose property that does exactly what it says.

billba:
I'm wary of introducing too many "behave this way" flags into properties. Is there really no best practice on how to handle this?

danmarshall:
I have a few ideas on how we might re-implement this feature, can we discuss on the original issue? https://github.com/Microsoft/BotFramework-WebChat/issues/363

macrozone:
@danmarshall the original issue focuses more on desktop interaction, so maybe we should open another issue about the problems on mobile devices. What i would expect as user when clicking on a button (on mobile device):

1. scroll to bottom
2. do not open keyboard

My option prevents the keyboard from opening, but unfortunatly does not scroll down on button click anymore

@billba totally agree, it's a workaround and does not solve the underlying issue (scrolling down while not opening the keyboard)

danmarshall:
Hi @macrozone , we are going to revisit how we want to approach the features for autoscroll and focus. I'm going to close this PR for now, but thanks for the contribution!

nupurkamble:
@macrozone hey I'm new to using bot and I want to disable the autofocus on input. i.e. i don't want the keyboard every time as it takes up too much space and the bot is barely visible. Can you please tell me how i can implement it in my bot. I've created a bot using .Net.

macrozone:
@nupurkamble I forked the project (node.js version) and added this option myself ;-) You will have to wait for the contributors to implement that functionallity.

Please follow this issue: https://github.com/Microsoft/BotFramework-WebChat/issues/623

PR #486, opened 2017-06-05
Test support for Adaptive Cards

photodesignch:
Adding adaptive cards support

### DIFF
--- a/test/cards/AdaptiveCardStyles.xaml
+++ b/test/cards/AdaptiveCardStyles.xaml
@@ -1,0 +1,305 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:core="clr-namespace:System;assembly=mscorlib"
+>
+
+    <Style x:Key="Error" TargetType="TextBlock">
+        <Setter Property="FontSize" Value="16"/>
+        <Setter Property="Foreground" Value="DarkRed"/>
+    </Style>
+
+    <Style x:Key="Hidden" TargetType="FrameworkElement">
+        <Setter Property="Visibility" Value="Collapsed"/>
+    </Style>
+
+    <!-- TextBlock definitions -->
+    <Style x:Key="Adaptive.TextBlock" TargetType="TextBlock" >
+    </Style>
+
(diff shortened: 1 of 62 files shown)


test/commands_map.ts
```diff
@@ -193,6 +195,23 @@ var commands_map: CommandValuesMap = {
             sendActivity(res, server_content.suggested_actions_card);
         }
     },
+    // CHRIS : TODO : taking out
```
photodesignch: forgot to remove these comments before I commit. I will remove them.

danmarshall:
Hi Chris, can you also rename the PR ?

danmarshall:
Perhaps it should be names "Adding Adaptive Cards tests."

test/mock_dl/index.ts
```diff
@@ -5,6 +5,7 @@ import bodyParser = require('body-parser');
 import * as path from 'path';
 
 const app = express();
+const fs = require('fs');
```
eanders-MS: Preferred to import like this: `import * as fs from 'fs';`

test/commands_map.ts
```diff
@@ -4,13 +4,15 @@ import * as Nightmare from 'nightmare';
 import * as express from 'express';
 declare let module: any;
 
+interface jsonObj {}
```
eanders-MS: This seems unnecessary.

photodesignch: removed

test/mock_dl/index.ts
photodesignch: changed to import * as fs from 'fs'

photodesignch: done

test/cards/AdaptiveCardStyles.xaml
```diff
@@ -0,0 +1,305 @@
+﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
```
danmarshall: can you remove this file entirely - unless there is some reason we need it.

photodesignch: @dandriscoll Will consolidate this folder to remove unnecessary files. Currently, this is just showing I copied all the files from AdaptiveCards folder.

reyesrico:
I suggest to add only the 5 cards that we are using in Channel Inspector instead of the 65 files. Lets use: Weather, BingSports, Simple, FlightUpdate and CalendarReminder.

Issue #195, opened 2016-12-05
Postback action writes to the chat window


eanders-MS:
The postback action should not display anything in the chat window. This is what imBack is for.

billba:
This was working and then I broke it. I'm special that way.

billba:
a test case!

mcsee:
Please do add a Test Case.
We detected the bug in our Regression Bot which has no functionality but to display all available controls

mcsee:
@billba Bug is still open

whenever a user presses an imback button the value is echoed in the window
can you check it out again ?

billba:
@mcsee that is correct behavior for imback

mcsee:
sorry. button specified as postBack	

document states
'postBack	Text of message which client will post to bot. Client applications will not display this message.'

but on webchat echo of the value field is shown instead of title

mcsee:
Sorry. apparently our old version was cached in chrome

Keep the bug open until the test case is added by  @palindromed

PR #488, opened 2017-06-08
#195 Postback action writes to the chat window

photodesignch:
The imBack vs postBack tests were involved only test/server_content.ts and test/commands_map.ts these two files. Since there were bunch of commits earlier didn't get merged to the base fork master, so the pull request showed all the files instead... If anyone knows how to exclude other files from the pull-request (code-review) please let me know.


danmarshall:
oh, because your commits are in your master branch. We may have to wait until after your other PR comes in.

PR #492, opened 2017-06-08
Test for #195  Postback action writes to the chat window

photodesignch:
Adding tests for imBack and postBack buttons.. Notice that postBack
button test currently will fail because we are expecting no echo: for
postBack, but there is a bug postBack will return echo:


danmarshall:
linked to #195

PR #494, opened 2017-06-09
fix carousel card scroll issue

thirteenflt:
fix carousel card scroll issue: scroll amount of width of the current card instead of always the first card's width


danmarshall:
Thanks for the contribution! However there may be side effects of resizing the window horizontally which may shift the "currentIndex" out of the viewport. So I think that currentIndex may need to be something calculated on the fly based on a measurement of the container.

GeekTrainer:
Closed pending further review.

PR #512, opened 2017-06-16
copy editing

DeniseMak:
This doc looked good! Just making a few minor copy edits. Please check if the **Optional: Create using SCSS** section is still correct since I rearranged it a bit.


danmarshall:
Lovely!

PR #517, opened 2017-06-19
Use rimraf package instead of rm -rf commands

pvanaarde:
Update the build-css and clean commands to be cross-platform capable rather than UNIX specific. Changed to use "rimraf" rather than "rm -rf".

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,13 +7,13 @@
   "scripts": {
     "build-test": "tsc -p ./test/mock_dl && tsc -p ./test",
     "build-test-watch": "concurrently \"tsc -w -p ./test/mock_dl\" \"tsc -w -p ./test\"",
-    "build-css": "node-sass ./src/scss/ -o . && rm -rf ./includes/",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "react-redux": "^5.0.3",
     "redux": "^3.6.0",
     "redux-observable": "^0.13.0",
+    "rimraf": "^2.6.1",
```
danmarshall: can you move this to the devDependencies section below?

GeekTrainer:
Bumping @pvanaarde WRT the update requested by @danmarshall.

Although I am curious the impact of leaving it with `rm -rf`. I personally run everything via a Git shell, so the command executes just fine. I have this thing about trying to avoid additional dependencies. :-)

billba:
@GeekTrainer what is a "Git shell"?

GeekTrainer:
@billba A PowerShell plugin provided with the GitHub client on Windows.

PR #525, opened 2017-06-22
speech unit test

w-Yuan:
1. Add speech options in the test.html and toggle speech capability by parameters in the test url
2. All the existing tests should be passed both speech enabled and speech disabled environment
3. Speech specific unit tests should be passed under speech enabled environment
    a. When speech is enabled, there should be a mic button
    b. After clicking on mic button, it should enter listening mode
    c. After clicking on the mic button, and click it again, it should exit listening mode
    d. After clicking on the mic button, and then start typing, it should exit listening mode
4. A bug fix due to the change of one package update (express)
5. Delete unnecessary file (AudioRecorder.ts)


test/mock_dl/index.ts
```diff
@@ -174,7 +174,7 @@ const processCommand = (req: express.Request, res: express.Response, cmd: string
     } else {
         switch (cmd) {
             case 'end':
-                current_uitests++;
+                current_uitests += 0.5; // For each end command, we will excute twice
```
danmarshall: Do we still want to execute twice?

test/uitest.js
```diff
@@ -61,14 +65,38 @@ describe('nightmare UI tests', function () {
...
+						// Non speech specific tests should also be passed under speech disabled environment
+						testUrl = `${url}&t=${cmd}&speech=disabled/ui`;
+						result = "";
+
+						if (cmd_index == 0) {
+							result = yield nightmare.goto(testUrl)
+								.viewport(width, 768);
+						}
+
+						result = yield nightmare.goto(testUrl)
```
danmarshall: will it work to have a common function for this code block?

test/mock_dl/index.ts
w-Yuan: yes. One for speech enabled and one for speech disabled.

test/uitest.js
w-Yuan: yes. Updated.

PR #531, opened 2017-06-23
fix regression: thumbnail card without images

shahidkhuram:


src/Attachment.tsx
```diff
@@ -177,7 +177,7 @@ export const AttachmentView = (props: {
             if (!attachment.content)
                 return null;
             const thumbnailCardBuilder = new CardBuilder.AdaptiveCardBuilder();
-            if (attachment.content.images) {
+            if (attachment.content.images.length > 0) {
```
danmarshall: can images be null? perhaps we need both `attachment.content.images && attachment.content.images.length > 0`

shahidkhuram: Agreed.

PR #539, opened 2017-06-30
Updating /test attachments as objects for server content.

reyesrico:


src/History.tsx
```diff
@@ -66,30 +66,35 @@ export class HistoryView extends React.Component<HistoryProps, {}> {
...
     // In order to do their cool horizontal scrolling thing, Carousels need to know how wide they can be.
     // So, at startup, we create this mock Carousel activity and measure it. 
     private measurableCarousel = () =>
         // find the largest possible message size by forcing a width larger than the chat itself
-        <WrappedActivity 
-            ref={ x => this.carouselActivity = x }
-            activity={ {
+        <WrappedActivity
+            ref={x => this.carouselActivity = x}
+            activity={{
```
billba: This checkin got rid of leading and trailing spaces for all objects. This is bad for two reasons:

* those spaces part of the implicit style guide for WebChat. They're in every file.
* it makes it very difficult to understand what actually changed in this checkin.

photodesignch: @billba I believed that @carlosscastro was trying to consolidate some of what he was working on.

danmarshall:
Let's just copy the desired changes into a new PR.

PR #559, opened 2017-07-08
Linking URLs with unencoded query strings

GeekTrainer:
There's an issue with the updated version of MarkdownIt that doesn't properly create links if the URL's querystring isn't encoded. I added a little replace code detecting URLs and encoding them properly.


src/FormattedText.tsx
```diff
@@ -53,7 +53,9 @@ const renderMarkdown = (
     text: string,
     onImageLoad: () => void
 ) => {
-    const src = text.replace(/<br\s*\/?>/ig, '\r\n\r\n');
+    const src = text
+                 .replace(/<br\s*\/?>/ig, '\r\n\r\n')
+                 .replace(/\[(.*)\]\((.*)\)/ig, (match, text, url) => `[${text}](${markdownIt.normalizeLink(url)})`);
```
danmarshall: can you add a comment about what this regex is searching for? For the regex-challenged folks like myself :-)

test/server_content.ts
```diff
@@ -180,7 +180,7 @@ export var mar_card: dl.Message = {
...
... are used.\r\n\r\n# h1\r\n## h2\r\n### h3\r\n#### h4\r\n\r\n### Headings *can* also contain **formatting**\r\n\r\n## URLs\r\n\r\nURLs can be made in a handful of ways:\r\n\r\n* A named link to [MarkItDown][3]. The easiest way to do these is to select what you want to make a link and hit `Ctrl+L`.\r\n* Another named link to [MarkItDown](http://www.markitdown.net/)\r\n* Some links have [query strings that need encoding](https://bing.com?q=some value)\r\n* Sometimes you just want a URL like <http://www.markitdown.net/>.\r\n\r\n## Images\r\nThis is an image\r\n\r\n![Image of Yaktocat](https://octodex.github.com/images/yaktocat.png)\r\n\r\n\r\n## Horizontal rule\r\n\r\nA horizontal rule is a line that goes across the middle of the page.\r\n\r\n---\r\n\r\nIt's sometimes handy for breaking things up.\r\n\r\n\r\n## Table\r\n|header1|header 2|\r\n|----|----|\r\n| cell 1 | cell 2|\r\n| cell three | cell four|\r\n\r\n## Whitespace\r\n\r\nHere's a line.\r\n\r\nThis has the standard two newlines before it.\r\n\r\n\r\n\r\nThis has four newlines before it.\r\n\r\n\r\n\r\n\r\n\r\nThis has six newlines before it.\r\n\r\n<br/><br/><br/><br/>This has two newlines and four &lt;br/&gt; tags before it."
```
danmarshall: actually it would be awesome to see a new test specifically for testing the regex instead of the one-test-for-all-markdown

@reyesrico - Can you help Christopher with this perhaps?

GeekTrainer: I was OK with it being part of the existing test, but I'm happy to make the new one.

src/FormattedText.tsx
GeekTrainer: Will do. :-)

test/server_content.ts
danmarshall: Yes I suppose it can be part of the existing test. Then we just need to have the test validate that both urls are correctly processed.

danmarshall:
This is even broken here on GitHub itself:

[query strings that need encoding](https://bing.com?q=some value)

danmarshall:
PR to this PR: https://github.com/GeekTrainer/BotFramework-WebChat/pull/1

danmarshall:
@GeekTrainer it seems that a customer has found an issue, when the link is within outer parentheses in the text. Can you take a look at updating the RegEx?

Here is a sample:
```
Some links are [already encoded](https://bing.com?q=some%20value) and some have [query strings that need encoding](https://bing.com?q=some value) and these can occur [multiple](https://bing.com?q=some%20value) [times](https://bing.com?q=some value) in one message. (Links can even appear [within](https://bing.com?q=some%20value) [parentheses](https://bing.com?q=some value) too.)
```

danmarshall:
@GeekTrainer - nvm, I was able to get a fix quickly. Fix was to make it non-greedy. PR coming.

PR #612, opened 2017-08-01
Fail Mocha test when UI test fails

danmarshall:
Also added ability for each test to append URL parameters.


danmarshall:
thanks for the review @eanders-MS

Issue #353, opened 2017-03-01
Continue previous conversation


baconstrudel:
Is there an easy way to reconnect to an existing conversation?

TheTobias:
Assuming the changes to DirectLineJS are made that would still mean one would have to store the conversation history and insert the history of the conversation before connecting back to the conversation to receive new messages.

billba:
@TheTobias Yes. Direct Line is not meant to be a full-fledged messaging system like e.g. Skype, Slack, or Facebook Messenger. It does not have an identity system nor message storage. It is just a simple protocol for exchanging messages with a bot. The more you want your client to work like a full-fledged messaging system, the more work you will have to do yourself.

sgellock:
The team doesn't believe it will address this issue at this time.

PR #615, opened 2017-08-03
provide past history conversation

### LINKED ISSUE #365: Supply history on startup
Some folks are interested in initializing WebChat with a history.

Since DirectLine doesn't archive messages, the bot would have to archive them separately.

Here are a couple of possible approaches:
* The web page could get the archived message out of band, e.g. by calling an endpoint provided by the bot. Then passes those to a new property and/or API in WebChat.
* The Bot Framework could enable a hook on the channel side where the bot supplies an `archiveProvider` of some kind. If it exists, DirectLine sends those messages to the WebChat. This has the advantage of working with any deployment of WebChat out of the box, but requires a lot of new plumbing in the Bot Framework.

This is clearly half-baked so I'm opening up a dialog.

danmarshall:
Fixes #365, #353  
@billba @GeekTrainer @eanders-MS - can you guys give this a review and let me know your thoughts. I think it's a simple approach.


src/Store.ts
```diff
@@ -273,6 +276,11 @@ export const history: Reducer<HistoryState> = (
 ) => {
     konsole.log("history action", action);
     switch (action.type) {
+        case 'Past_History':
+            return {
+                ... state,
+                activities: action.activities.slice()
```
billba: let's instead do:

    activities: ... action.activities

src/Chat.tsx
```diff
@@ -26,6 +26,7 @@ export interface ChatProps {
     directLine?: DirectLineOptions,
     speechOptions?: SpeechOptions,
     locale?: string,
+    pastHistory?: Activity[],
```
billba: "past" and "history" mean the same thing - how about just `history`?

src/Store.ts
```diff
@@ -236,14 +236,14 @@ export interface HistoryState {
 }
 
 export type HistoryAction = {
-    type: 'Receive_Message' | 'Send_Message' | 'Show_Typing' | 'Receive_Sent_Message'
+    type: 'Receive_Message' | 'Send_Message' | 'Show_Typing' | 'Receive_Sent_Message',
```
billba: This is a good fix but it doesn't belong with this PR. Should be a separate "cleanup" PR so that the diffs are clear on what exactly is involved to the task of setting history.

src/Store.ts
```diff
@@ -236,14 +236,14 @@ export interface HistoryState {
...
     activity: Activity
 } | {
     type: 'Send_Message_Try' | 'Send_Message_Fail' | 'Send_Message_Retry',
     clientActivityId: string
 } | {
-    type: 'Send_Message_Succeed'
-    clientActivityId: string
+    type: 'Send_Message_Succeed',
+    clientActivityId: string,
```
billba: ditto

src/Chat.tsx
```diff
@@ -74,6 +75,13 @@ export class Chat extends React.Component<ChatProps, {}> {
 
         konsole.log("BotChat.Chat props", props);
 
+        if (props.pastHistory) {
+            this.store.dispatch<HistoryAction>({
+                type: 'Past_History',
```
billba: See above -- I'd call this `Set_History`

README.md
```diff
@@ -220,6 +220,13 @@ WebChat currently defaults to *not* sending 'typing' activities to the bot when
 
 You can supply WebChat with the id (and, optionally, a friendly name) of the current user by passing `user: { id: user_id, name: user_name }` to `App`/`Chat`. This object is passed with every activity sent from WebChat to the bot, which means it is not available to the bot *before* any activities are sent. See [The Backchannel](#the-backchannel) to find out how your web page can programmatically send non-message activities to the bot.
 
+### Provide Chat History
+
+You can initialize the WebChat conversation by passing an array of activity objects in `pastHistory: [ activity_object1, activity_object2, ... ]` to `App`/`Chat`. Each activity object should have a unique `id` value, and the `from` property should match either a user identity (above) or the bot's identity. An example activity message might look like:
+```
```
billba: I'm worried about identity. You say the user and bot identities should match up, but you don't check for it. Also, this history might include identities of other folk in a conversation. Here's one idea: what if in this list of activities you have special tokens for identities, e.g. from: { id: "$BOT$" } and from : {ID: "$USER$"} which we would automatically replace with the current bot and user identities as supplied (including the name)? Others would be left alone.

If you don't do this, I think you should supply a little more guidance on how they should go about supplying a well-formed history, which should probably go into how to store the history. It is at the point the history is originally created that they know who is who.

src/Chat.tsx
```diff
@@ -74,6 +75,13 @@ export class Chat extends React.Component<ChatProps, {}> {
 
         konsole.log("BotChat.Chat props", props);
 
+        if (props.pastHistory) {
+            this.store.dispatch<HistoryAction>({
```
billba: we've been switching to using `ChatActions` (which includes `HistoryAction`), since a given action might actually be handled by multiple reducers. One day I might get rid of the individual `{reducer}Action` types for this reason.

test/commands_map.ts
```diff
@@ -35,7 +35,31 @@ interface CommandValuesMap {
...
+                    type: 'message',
+                    from: server_content.config.bot,
+                    text: 'The bot said this in the past.'
+                }
+            ]
+        },
+        client: function () {
+            //look for 5 messages - the 2 above, greeting, history, and bot response
+            return document.querySelectorAll('.wc-message').length === 5;
         }
```
billba: I'm not familiar with the test code, but elsewhere in WebChat we use the more concise fat arrow functions when possible. Here it would be:

    client: () => document.querySelectorAll('.wc-message').length === 5,

reyesrico: I suggest to add this urlAppend code as part of the server_content file.

danmarshall: @billba good suggestion, let's also do this in a separate cleanup PR.

README.md
danmarshall: Yes, this also opens the door for group conversation (more than 2 people) so perhaps there is also a way to indicate the $OTHER$ users?

danmarshall: So, in effect, they don't actually need to match?

danmarshall: @billba agree that identity can be a big issue. Should we solve this prior to accepting this PR? Or let this go through and apply solutions later?

danmarshall: @billba ping

LarsLiden:
Any thoughts on if/when this will get pulled into the main branch?  It's a super useful feature!

danmarshall:
@GeekTrainer @billba - any thoughts on this PR? Would like it to go through now, and we can work on the identity issue later. 👍 or 👎?

LarsLiden:
Wan we close this one out?  Anything blocking here?  Bill said he's fine with it.

src/Chat.tsx
danmarshall: I agree. I wanted to disambiguate from the history component.

eanders-MS:
@LarsLiden Bill is concerned that we haven't established a process for approving API changes, and that "looks good to me" isn't sufficient approval here. I'm in agreement with that. At the same time, I think this change is straightforward enough to allow through in the meantime. @billba are you willing to compromise in this instance?

billba:
I still have concerns about this one. Has anyone looked into collisions between the various supplied IDs from the past conversation, and the incoming IDs of the future conversation?

danmarshall:
@billba - we could certainly add a unit test that provides a colliding ID. What would be the expected behavior?

billba:
I have no idea. This is my point - I think this is not entirely straightforward.

danmarshall:
True, but probably the best way to work through them is to take the PR and let the community present issues. Really the only thing at stake is the API surface, which wouldn't likely change. We can always unique-ify stuff under the hood.

billba:
Wow I totally disagree with that.

danmarshall:
Web Chat implicitly already allows "injection" through the use of `IBotConnection`. We do our best to render activity objects that are provided.

billba:
True, in fact you could implement this feature with a custom IBotConnector that pumps in the history before connecting to the bot. That's another good reason to think deeply about the state machine. I firmly do not believe that the right approach is to just turn on the functionality and wait for the bugs to pour in.

LarsLiden:
Anything you can do to unblock us would be appreciated.  We’re still working from Dan’s private branch which means we aren’t getting updates.

billba:
This example of client middleware should provide the exact same functionality as this PR.

```ts
// assumes that you have already imported Observable.from and Observable.concat from rxjs

const dl = new BotChat.DirectLine({ secret });

const history: Activity[] = [/* Array of activities to load before showing new activities from the bot */];

BotChat.Chat({
    botConnection: {
        ... dl,
        activity$: Observable.from(history).concat(dl.activity$)
    },
    // other Chat props go here
});
```

If you just want too use WebChat in "read-only" mode to display a history without connecting to a bot, create your own `IBotConnection` object from scratch:

```ts
// assumes that you have already imported Observable.from, Observable.of, and BehaviorSubject from rxjs

const history: Activity[] = [/* Array of activities to load before showing new activities from the bot */];

BotChat.Chat({
    botConnection: {
        connectionStatus$: new BehaviorSubject(ConnectionStatus.Online),
        activity$: Observable.from(history),
        postActivity: () => Observable.of('id'),
        end: () => {},
    },
    // other Chat props go here
});
```

LarsLiden:
Hi Bill -
Finally getting around to trying this, and I'm finding that only one side
of the conversation is showing up (in particular the activities send by the
Bot).   The user side is missing.   Any thoughts?

[image: Inline image 1]

compulim:
The past history feature is now done on the server-side. We decided to do it that way because:
- Cleaning up conversation (a.k.a. privacy stuff) is easier on the server-side
- Even we have the client-side solution ready, there are a good chance we still need a server-side solution
- Signing in/out might lose conversation, or risk exposing privacy information

Given the points above, we are doing it on the server-side now.

kienct89:
@compulim thanks for your information,

May I know how to config it from now on? or should it work out of the box?

smit-1999:
@compulim What's the update for the server side?How are you managing to show previous chats from the server side?
 I am trying out on the server side as well. Currently I am fetching all messages to be shown from the database and then sending each of them as an activity to the client when he connects.

PR #617, opened 2017-08-03
Fix #607 Remove jQuery from sample code

### LINKED ISSUE #607: Speech sample is downloading Google jquery code from non-secure URL
in samples/speech/index.html, there is a download of jquery code from Google at http://ajax.googleapis.com/ajax/libs/jquery/1.9.0/jquery.min.js

Unfortunately, if one hosts the webchat on a secure website, this leads to Chrome complaining about "unsafe scripts", and subsequently to a failure of the speech recognition.

I changed this to download from http**s**, which is also the URL that Google is giving in their docs. This resolves the issue.

compulim:
Fix #607

### DIFF
--- a/samples/speech/index.html
+++ b/samples/speech/index.html
@@ -1,5 +1,5 @@
 ﻿<!DOCTYPE html>
-<!-- 
+<!--
     This is a sample HTML file which shows how to use speech in the WebChat.
     1. Build the project: "npm run build"
     2. Start a web server: "npm run start"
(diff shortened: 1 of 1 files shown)


compulim:
I should fix more `samples/*.html`. But want to see if you guys are okay on this coding style.

eanders-MS:
Nice clean up.

compulim:
@billba @GeekTrainer @eanders-MS @danmarshall feel free to comment on.

samples/speech/index.html
```diff
@@ -46,94 +42,99 @@
...
-                id: params['botid'] || 'botid',
-                name: params["botname"] || 'botname'
+                id: params.botid || 'botid',
+                name: params.botname || 'botname'
             };
-            
-            window['botchatDebug'] = params['debug'] && params['debug'] === "true";
 
-            // // Option 1 : No Speech
+            window.botchatDebug = params.debug && params.debug === 'true';
```
billba: This is obviously correct JavaScript, but the reason I prefer the `[]` style is that this will choke in TypeScript. This is not currently TypeScript, but I'd like the option for people (including myself) to copy and past this into TypeScript code.

compulim: Sure, will have another PR in a moment.

PR #620, opened 2017-08-04
Bracket accessors for params

compulim:
@billba moving away from dot accessor for `params` for compatibility with TypeScript


billba:
Thank you!

PR #630, opened 2017-08-09
Fix #363 Remove auto-focus and add onKeyDown focus

### LINKED ISSUE #363: focus should return to the text entry box.

compulim:
Removed all `focus()` code, which is in these 3 areas:

* Click/tap send button
* Click/tap action button
* Click/tap upload file button, then select a file

Then add a new one:

* When a user is "`onKeyDown`" on `<Chat>`, we will do `focus()`
  * Except when the user pressed <kbd>ENTER</kbd>, <kbd>TAB</kbd> and <kbd>SPACEBAR</kbd>
  * Also except when the user typing inside the sendbox (using `evt.stopPropagation()` to achieve that)


src/Chat.tsx
```diff
@@ -74,6 +75,8 @@ export class Chat extends React.Component<ChatProps, {}> {
 
         konsole.log("BotChat.Chat props", props);
 
+        this.handleKeyDown = this.handleKeyDown.bind(this);
```
danmarshall: Can this be bound to a React component? Doesn't it need a DOM element?

src/Chat.tsx
```diff
@@ -113,6 +116,17 @@ export class Chat extends React.Component<ChatProps, {}> {
         });
     }
 
+    private handleKeyDown(evt: any) {
```
danmarshall: I think we can strongly-type the event

src/Chat.tsx
```diff
@@ -113,6 +116,17 @@ export class Chat extends React.Component<ChatProps, {}> {
...
+        // Chat textbox has "onKeyDown" handler that will call evt.stopPropagation, preventing the handled event to bubble up to this point
+        if (
```
danmarshall: Remember that Adaptive Cards may also have input elements (like textboxes, checkboxes, etc.) These may not have stopPropagation handlers on them. Can you test with a card like [this](http://adaptivecards.io/samples/#InputForm)?

test/commands_map.ts
```diff
@@ -406,25 +406,36 @@ var commands_map: CommandValuesMap = {
         client: function () {
             return (((document.querySelector('.wc-shellinput') as HTMLInputElement).placeholder === 'Type your message...'));
         }
+    },
+    "focus on type": {
```
danmarshall: Very cool :)
Can we also add a test for non-focus on type, when typing in an AdaptiveCard textbox?

compulim:
Since we have inputs in Adaptive Cards, looks like <kbd>ENTER</kbd>, <kbd>SPACEBAR</kbd>, and <kbd>TAB</kbd> is not a sufficient check.

I am turning the check into `tabIndex >= 0`. But hit a browser compatibility issue: IE11 will return `0` for elements without `tabIndex`. But other browsers return either `0` for focusable elements (e.g. `<input>`), and `-1` for unfocusable elements (e.g. `<div>`).

This [article](https://nemisj.com/focusable/) talks about focus in-depth. Will do some investigations based on it.

p.s. Bing.com homepage also have similar "type to focus" behavior.

danmarshall:
@compulim can you review the feedback?

compulim:
I thought I might need to redo large part of the PR because of the `tabIndex` quirks in IE11. Will look at your comments and see.

src/Chat.tsx
compulim: We should do `.bind()` in constructor instead of inside `render()` because React do reference equals on props to find out if it need to be updated. If we do `.bind()` in `render()` loop, it'd result in a wasted render (i.e. a re-render that render nothing new, `React.addons.Perf.printWasted()`).

src/Chat.tsx
compulim: Done.

src/Chat.tsx
compulim: Tested with Adaptive Cards.

compulim:
@danmarshall done, could you take a look?

test/commands_map.ts
compulim: Done.

src/Chat.tsx
```diff
@@ -74,6 +78,8 @@ export class Chat extends React.Component<ChatProps, {}> {
 
         konsole.log("BotChat.Chat props", props);
 
+        this.handleKeyDownCapture = this.handleKeyDownCapture.bind(this);
```
danmarshall: is this necessary?

src/Chat.tsx
```diff
@@ -113,6 +119,21 @@ export class Chat extends React.Component<ChatProps, {}> {
         });
     }
 
+    private handleKeyDownCapture(evt: React.KeyboardEvent<HTMLDivElement>) {
+        const target = evt.target as HTMLElement;
+        const chatviewPanelElement = findDOMNode(this.chatviewPanel);
+        const tabIndex = getTabIndex(target);
+
+        if (
+            target === findDOMNode(this.chatviewPanel)
```
danmarshall: target === chatviewPanelElement

src/Chat.tsx
compulim: Yes, look at this [tutorial](https://facebook.github.io/react/docs/handling-events.html) from React. For ES6 class, we should call `.bind()` to make sure `this` points to the component itself.

Some people do it in `render()`, but it hurts performance.

src/Chat.tsx
compulim: Thanks for catching this, I would remove the `chatviewPanelElement` line instead, so people won't look at 2 places to find out what that `===` is doing.

src/Chat.tsx
danmarshall: oh yes of course. Can you use a lambda instead? We already have the lambda style in https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/MessagePane.tsx#L59

test/commands_map.ts
```diff
@@ -375,6 +375,14 @@ var commands_map: CommandValuesMap = {
             sendActivity(res, server_content.adaptive_cardsFn(json));
         }
     },
+    "card Inputs": {
+        client: function () {
+            return true;
```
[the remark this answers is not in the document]
compulim: Done. +1

src/Chat.tsx
compulim: I could use that, but it's still a perf hit. Because every time it `render()`, we will create a new lambda function, and default `shouldComponentUpdate()` won't be able to figure out that both lambda functions actually land on the same event handler. Thus, a wasted render will be detected thru `React.addons.Perf.printWasted()` call.

danmarshall: oh, well then we should probably convert the existing lambdas... but not in this PR.

compulim:
@danmarshall, done updating for your comments and added Edge quirk-fix.

src/Chat.tsx
```diff
@@ -86,6 +91,30 @@ export class Chat extends React.Component<ChatProps, {}> {
...
+            || tabIndex < 0
+        ) {
+            evt.stopPropagation();
+
+            let key: string;
+
+            // Quirks: onKeyDown we re-focus, but the newly focused element does not receive the subsequent onKeyPress event
+            //         It is working in Chrome/Firefox/IE, confirmed not working in Edge/16
+            //         So we are manually appending the key if they can be inputted in the box
+            if (/(^|\s)Edge\//.test(navigator.userAgent)) {
```
danmarshall: Super cool. I hope that this behavior does not get "fixed" in a future version of Edge. - then we would have 2 keys.

src/Chat.tsx
```diff
@@ -254,3 +282,15 @@ const ResizeDetector = (props: {
                 frame.contentWindow.onresize = props.onresize;
         } }
     />;
+
```
danmarshall: Can you add a comment as to these special keys

src/Chat.tsx
compulim: Now target Edge 16 only. Also filed a bug to Edge team.

src/Chat.tsx
compulim: Done.

src/Chat.tsx
billba: Reassigning the instance variable using `bind` is both clever and ugly in the extreme. I much prefer creating a second instance variable:

```ts
_handleKeyDownCapture = (evt: React.KeyboardEvent<HTMLDivElement>) => this.handleKeyDownCapture(evt);

...

onKeyDownCapture={ this._handleKeyDownCapture }
```

Prettier, with the same perf. (We do this pattern elsewhere, or at least we did.)

More importantly, @compulim, I think you have an important misconception here. There is no such thing as "default `shouldComponentUpdate()`". By default there is no `shouldComponentUpdate` - you have to add it yourself. It is true that a simple and common implementation is to just compare the value of `props`, in which case you're right, you want to avoid creating a new lambda on every call.

danmarshall: Agree that we should just go with an arrow function. Less code and less confusing. 👍

compulim: @billba I was using `PureComponent` for my other projects so I didn't notice we wasn't using it.

Will change to arrow second variable instead of `.bind()` to existing.

billba: You're right - if you're using PureComponent then you *do* get a default `shouldComponentUpdate` that just looks at props - and we do use this for `Carousel` and `CarouselAttachment`.

compulim:
@danmarshall I have incorporated all changes, could you take a look?

src/Chat.tsx
```diff
@@ -86,6 +92,34 @@ export class Chat extends React.Component<ChatProps, {}> {
...
+            || tabIndex < 0
+        ) {
+            evt.stopPropagation();
+
+            let key: string;
+
+            // Quirks: onKeyDown we re-focus, but the newly focused element does not receive the subsequent onKeyPress event
+            //         It is working in Chrome/Firefox/IE, confirmed not working in Edge/16
+            //         So we are manually appending the key if they can be inputted in the box
+            if (/(^|\s)Edge\/16\./.test(navigator.userAgent)) {
```
danmarshall: are we going to revisit this in Edge 17?

compulim: Yes. I have file a bug to Edge team.

PR #649, opened 2017-08-24
Add Markdown features section

danmarshall:


danmarshall:
@GeekTrainer @billba does it make sense to expose the markdown options out to the main ChatOptions?

```
const markdownIt = new MarkdownIt({ html: false, linkify: true, typographer: true });
```

GeekTrainer:
I poked around looking for an issue, and didn't see one. What is the goal here?

danmarshall:
@GeekTrainer this PR as of right now, is just changes to the readme, which specify the files containing the markdown-it options.
So, a developer would need a rebuild if they wanted to change these options, such as HTML rendering capability. We *could* promote these options to the main API, so that a developer could modify these options without a rebuild.

GeekTrainer:
Ahhhh... OK. I don't think we should worry too much about promoting the options, as I think we want to keep those relatively limited. But I'm OK with adding this to the docs.

danmarshall:
@GeekTrainer cool. Do you approve the readme changes?

README.md
```diff
@@ -172,14 +172,18 @@ You can alter or add localized strings in `/src/Strings.ts`:
...
-### Architecture
+#### Architecture
 
 * `Chat` is the top-level React component
 * `App` creates a React application consists solely of `Chat`
 * `Chat` largely follows the Redux architecture laid out in [this video series](https://egghead.io/lessons/javascript-redux-the-single-immutable-state-tree)
 * To handle async side effects of Redux actions, `Chat` uses `Epic` from [redux-observable](https://redux-observable.js.org) - here's a [video introduction](https://www.youtube.com/watch?v=sF5-V-Szo0c)
 * Underlying `redux-observable` (and also [DirectLineJS](https://github.com/microsoft/botframework-directlinejs)) is the `RxJS` library, which implements the Observable pattern for wrangling async. A minimal grasp of `RxJS` is key to understanding WebChat's plumbing.
 
+#### Markdown features
```
GeekTrainer: I'd leave this as simply Markdown as the header

README.md
```diff
@@ -172,14 +172,18 @@ You can alter or add localized strings in `/src/Strings.ts`:
...
+
+WebChat uses the [markdown-it](https://markdown-it.github.io/) library for markdown rendering. Markdown-it offers many rendering [options](https://github.com/markdown-it/markdown-it#init-with-presets-and-options), such as HTML rendering within markdown. You can change these options [here](https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/FormattedText.tsx#L28) in your own build of WebChat.
```
GeekTrainer: WebChat uses markdown-it (drop the **the**)

Update the link text to not use the word *here* (accessibility)

README.md
```diff
@@ -172,14 +172,18 @@ You can alter or add localized strings in `/src/Strings.ts`:
 
 Behavioral customization will require changing the TypeScript files in `/src`. A full description of the architecture of WebChat is beyond the scope of this document, but here are a few starters:
 
-### Architecture
+#### Architecture
```
GeekTrainer: Why the change in header level?

danmarshall: The previous header was 3 levels. Given that this paragraph ended with a colon, it seemed like it should be a sub-section. Thoughts?

PR #652, opened 2017-08-28
Added Norwegian Bokmål (and making nn default to nb)

taarskog:


danmarshall:
Hello, thanks for the translation! We have just added a [new field for VAT](https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/Strings.ts#L27), can you also add a translation of this new field?

danmarshall:
Thank you!!

PR #657, opened 2017-08-30
throttle consecutive speech reco start calls to 500ms. Fixes #654

### LINKED ISSUE #654: Uncaught exception in Speech re initialization
If we click on the Mic multiple times by mistake , there is an uncaught exception which is causing issues. Due to this issue , bot is getting in to a loop and never coming out with out a refresh page.

botchat.js:1426 Uncaught DOMException: Failed to execute 'start' on 'SpeechRecognition': recognition has already started.
    at BrowserSpeechRecognizer.startRecognizing (http://IP/sample_files/botchat.js:5348:29)
    at Function.SpeechRecognizer.startRecognizing (http://IP/sample_files/botchat.js:5253:39)
    at SafeSubscriber._next (http://IP/sample_files/botchat.js:17787:48)
    at SafeSubscriber.__tryOrSetError (http://IP/sample_files/botchat.js:1466:16)
    at SafeSubscriber.next (http://IP/sample_files/botchat.js:1406:27)
    at Subscriber._next (http://IP/sample_files/botchat.js:1344:26)
    at Subscriber.next (http://IP/sample_files/botchat.js:1308:18)
    at DoSubscriber._next (http://IP/sample_files/botchat.js:44464:24)
    at DoSubscriber.Subscriber.next (http://IP/sample_files/botchat.js:1308:18)
    at FilterSubscriber._next (http://IP/sample_files/botchat.js:16301:30)

shahidkhuram:


src/Store.ts
```diff
@@ -533,6 +533,7 @@ const stopListening: Epic<ChatActions, ChatState> = (action$) =>
 
 const startListening:Epic<ChatActions, ChatState> = (action$, store) =>
     action$.ofType('Listening_Starting')
+    .throttleTime(500)
```
billba: can you explain this? Where did the 500ms number come from? Is this working around a latency problem, and if so what happens if that thing we're working around takes 501ms?

shahidkhuram: The issue is that the following operations on the recognizer aren't supported in quick succession:
Start, Stop, Start

Stop should ideally be sync (or async with completion) but isn't. The mitigation is to throttle the signal to start listening (if the user is spamming the button).
The 500ms value is derived empirically based on a mid-ranged laptop (with buffer built in).

Since this is only a mitigation, the worst case scenario theoretically remains the same as today.

compulim: Look like we need to update [`SpeechModule.ts`](https://github.com/Microsoft/BotFramework-WebChat/blob/master/src/SpeechModule.ts#L104) and making the stop more definitive by signaling when speech is fully stopped, either:
1. Turning `stop()` into a Promise, or,
2. A new event handler

danmarshall: @shahidkhuram @compulim what's the consensus on this?

shahidkhuram: Unfortunately it's unlikely that i'll be able to look at this again any time soon. @compulim can you drive any further changes necessary/alternate fixes? If not i suggest taking the change as is at least as a mitigation.

danmarshall: @w-Yuan is there any test you can add to simulate this?

compulim:
@shahidkhuram I have made a PR #881 that actually look at `webkitSpeechRecognition.onend` event. So we don't need to sleep for 500ms. I will credit you over there.

shahidkhuram:
Thanks!

PR #665, opened 2017-09-06
Translated receiptVat in cs-CZ

msimecek:
Having two string about taxes also brought the need to change receiptTax to make it disctinct.


danmarshall:
Thanks!

PR #679, opened 2017-09-07
Fix #678: mock_dl to support watermark ACK and multiple conversations

### LINKED ISSUE #678: Mock_dl server is not reliable because missing ACK feature
Me and @reyesrico hit some issues when writing new tests and found that mock_dl does not implement Direct Line watermark/ACK feature correctly, leading to dropped messages and unreliable tests in Chrome.

I am proposing to revamp the message queue to support ACK and multiple conversation ID so it works with multiple browsers opened against the single server.

compulim:
Current `mock_dl` call [`queue.shift()`](../blob/master/test/mock_dl/index.ts#L245) for retrieving latest messages and we found that messages could be lost during intensive testing.

This PR improve test reliability, includes:
* ACK-ing thru `watermark` will purge messages with older watermark, won't delete current one
* Support multiple conversations thru `POST /mock/conversations` and real `conversationId`
  * As we tested, it seems `/mock/tokens/generate` and `/mock/tokens/refresh` is not called during `npm test`, so we are leaving them with a dummy conversation ID for now

### DIFF
--- a/test/mock_dl/conversations.ts
+++ b/test/mock_dl/conversations.ts
@@ -54,6 +54,6 @@ export function pushMessage(conversationID: string, activity: any) {
 }
 
 export function endConversation(conversationID: string) {
-  console.log(`Ending conversation ${ id }`);
+  console.log(`Ending conversation ${ conversationID }`);
   delete conversations[conversationID];
 }
--- a/test/mock_dl/conversations.ts
+++ b/test/mock_dl/conversations.ts
@@ -1,3 +1,5 @@
+import { Activity } from 'botframework-directlinejs';
+
 interface Conversations {
   [key: string]: Conversation
 }
@@ -32,7 +34,7 @@ class Conversation {
 export class Message {
   constructor(
     public watermark: number,
-    public activity: any
+    public activity: Activity
   ) {}
 }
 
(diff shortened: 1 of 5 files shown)


test/mock_dl/conversations.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.messages.push(new Message(this.nextWatermark, content));
+
+    return this.nextWatermark++;
+  }
+}
+
+export class Message {
+  constructor(
+    public watermark: number,
+    public activity: any
```
danmarshall: activity: Activity ?

test/mock_dl/index.ts
```diff
@@ -47,7 +47,8 @@ app.post('/mock/tokens/generate', (req, res) => {
     const token = get_token(req);
 
     res.send({
-        conversationId,
+        // We don't have a working mock on token generate/refresh yet
+        conversationId: 'DUMMY',
```
danmarshall: probably won't pass poliCheck :) can you change these?

test/mock_dl/conversations.ts
compulim: Fixed.

test/mock_dl/index.ts
compulim: Done.

compulim:
@danmarshall Done incorporating all changes, can you take a look?

Issue #598, opened 2017-07-24
Move React, react-dom and @types to npm peerDependencies


prannayb:
Our parent project uses react 15.6.1 and latest @types/react 15.0.39. But the botframework-webchat package lists react and the types as dependencies. This causes react and the 15.0.38 typings to get installed under the bot package and conflict with the typings in my project, and doesn't compile. Please move react and its typings package to peerDependencies so that it can use the ones already installed via my project.

danmarshall:
Hi @prannayb thank you for reporting this. I'm not certain that we can move our typings to peerDependencies. Would this force others to manually install each of the peer dependencies manually?

FYI we just recently locked to exact version numbers because some updates were breaking our build. We may be able to revisit this for React.

corinagum:
Fixed in v4

PR #705, opened 2017-09-30
Move @types/react to dev dependency to avoid conflict with the consuming applications' react types.

mattmazzola:
Since this package includes @types as dependency the consuming project will include these in it's compilation step and conflict with the root level @types/react either as duplicate definition of same type or as conflicts with different definitions for the same type.

Looks like all the other @types are properly in the devDependencies section.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
   "author": "Microsoft Corp",
   "license": "MIT",
   "dependencies": {
-    "@types/react": "15.0.38",
     "botframework-directlinejs": "0.9.11",
     "core-js": "2.4.1",
     "markdown-it": "8.3.1",
@@ -50,6 +49,7 @@
     "@types/express": "4.0.35",
     "@types/markdown-it": "0.0.2",
     "@types/nightmare": "1.6.30",
+    "@types/react": "^15.0.38",
     "@types/react-dom": "0.14.23",
     "@types/react-redux": "4.4.38",
     "awesome-typescript-loader": "3.1.2",


eanders-MS:
Nice... this an occasional source of build breaks in the botframework-emulator project as well. @danmarshall can we accept this?

eanders-MS:
Seems like we should move all `@types/*` dependencies to devDependencies, not just this one.

danmarshall:
There are 2 scenarios which seem to compete with each other:
1. Consuming project has a duplicate definition of React.
2. Consuming project has *no* definition of React.

Fixing one of these breaks the other :\

danmarshall:
Perhaps we need to move to peer dependencies as suggested in #598.

See https://stackoverflow.com/questions/26737819/why-use-peer-dependencies-in-npm-for-plugins

danmarshall:
Hi Matt - this is an interesting approach. So, you are suggesting that React-based consumers use `npm install --production` so that they do not download devDependencies?

mattmazzola:
No, devDependencies are only installed for the project in folder where `npm install` was run. For consumers of this package it will not be the case so they do not need to change behavior (They should not have to run `npm install --production` )

I believe #598 using peerDependencies is actually the correct solution because this project still does require at least some version of react to exist to function since it uses / exposes those types as well.  Either way, this project does not know what version of react the consuming project will choose to use and should not install it's own which was causing conflicts.

danmarshall:
Thanks @mattmazzola . I also think the peerDependencies is the correct approach, but it does present additional challenges:

Some people are here to just build this repo as-is. They are not consuming it in another project. They will not be able to build until they choose a version of React & RxJS and then install them. What does this process look like, and where can we specify the "recommended" version number.

I would love it if they can just continue doing `npm install` without having to do this extra step.

delight-by:
Why can't we use both peerDependencies and devDependencies at the same time? I've seen library authors doing that with normal dependencies and can't really imagine why it might not work with @types dependencies. This way we'll cover both cases: people building the repo will get devDeps while people consuming the library will get peerDeps warnings.

corinagum:
Sorry we didn't get this PR in, but in good news this has been addressed in v4!

compulim:
@delight-by we now have `react` in both `devDependencies` and `peerDependenices`.

PR #727, opened 2017-10-11
Make upload and speak icon tabbable

compulim:
# Make upload and speak button tabbable

Make sure all buttons are tabbable, and make sure <kbd>SPACEBAR</kbd> and <kbd>ENTER</kbd> also works.

# Focus indicator for all buttons

Add focus indicator (or CSS outline) to all clickable elements.

image

### DIFF
--- a/src/Shell.tsx
+++ b/src/Shell.tsx
@@ -28,9 +28,6 @@ class ShellContainer extends React.Component<Props> implements ShellFunctions {
     private textInput: HTMLInputElement;
     private fileInput: HTMLInputElement;
 
-    private _handleSendButtonKeyPress = this.handleSendButtonKeyPress.bind(this);
-    private _handleUploadButtonKeyPress = this.handleUploadButtonKeyPress.bind(this);
-
     private sendMessage() {
         if (this.props.inputText.trim().length > 0) {
             this.props.sendMessage(this.props.inputText);
@@ -116,7 +113,7 @@ class ShellContainer extends React.Component<Props> implements ShellFunctions {
             <div className={ className }>
                 <label
                     className="wc-upload"
-                    onKeyPress={ this._handleUploadButtonKeyPress }
+                    onKeyPress={ evt => this.handleUploadButtonKeyPress(evt) }
                     tabIndex={ 0 }
                 >
                     <svg>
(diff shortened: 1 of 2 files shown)


src/Shell.tsx
```diff
@@ -89,16 +107,27 @@ class ShellContainer extends React.Component<Props, {}> implements ShellFunction
...
-            !this.props.listening && 'inactive',
+            !this.props.listening && 'inactive'
         );
 
         return (
-            <div className={className}>
-                <input id="wc-upload-input" type="file" ref={ input => this.fileInput = input } multiple onChange={ () => this.onChangeFile() } />
-                <label className="wc-upload" htmlFor="wc-upload-input">
+            <div className={ className }>
+                <label
```
compulim: We have to move DOM around to show the focus indicator correctly. In short, `<label>` should wrap around `<input>` so we don't need `htmlFor` prop and ID-ing everything.

danmarshall: OIC. The label trick gives us a free "click" event on the file uploader.

compulim: Yes, until we figured out the "free click" does not works with "spacebar" or "enter". :(

src/Shell.tsx
```diff
@@ -24,22 +24,38 @@ export interface ShellFunctions {
...
 
-class ShellContainer extends React.Component<Props, {}> implements ShellFunctions {
+class ShellContainer extends React.Component<Props> implements ShellFunctions {
     private textInput: HTMLInputElement;
     private fileInput: HTMLInputElement;
 
-    constructor(props: Props) {
-        super(props);
-    }
+    private _handleSendButtonKeyPress = this.handleSendButtonKeyPress.bind(this);
```
danmarshall: can we use lambdas for these?

src/scss/botchat.scss
```diff
@@ -568,78 +560,105 @@
...
-        vertical-align: middle;
-    }
+        right: 0;
 
-    .wc-console label {
-        cursor: pointer;
-        display: inline-block;
-        height: 40px;
-    }
+        > * {
```
danmarshall: can we not use star?

src/Shell.tsx
billba: Yes please.

compulim: Moved to Lambda.

src/scss/botchat.scss
compulim: Fixed.

PR #738, opened 2017-10-17
Updating brand color for more contrast ratio

compulim:
# Updating brand color

To match 4.5:1 contrast ratio, the following colors were updated:

* Update brand color
* Update timestamp color
* Remove `$c06` because it was not longer used

image

### DIFF
--- a/src/scss/includes/colors.scss
+++ b/src/scss/includes/colors.scss
@@ -5,13 +5,13 @@ $c02: rgba(0, 0, 0, 0.2);
 $c03: #8a8a8a;
 $c04: #999999;
 $c05: #cccccc;
-$c06: #0063b1;
-$c07: #3a96dd;
+$c07: #0078d7;
 $c08: #808c95;
 $c09: #eceff1;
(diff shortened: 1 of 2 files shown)


src/scss/includes/colors.scss
```diff
@@ -5,19 +5,20 @@ $c02: rgba(0, 0, 0, 0.2);
...
 $c05: #cccccc;
-$c06: #0063b1;
-$c07: #3a96dd;
+$c07: #0078d7;
 $c08: #808c95;
 $c09: #eceff1;
 $c10: #dbdee1;
 $c11: #d2dde5;
 $c12: #ffa333;
+$c13: #767676; // used for timestamp under chat bubbles
```
danmarshall: can you repurpose c06 so there is no gap?

compulim:
Updated to remove `$c06` gap. We obsoleted `$c06`, will replace with `$c_brand` with `0.8` instead.

danmarshall:
Ship it!

PR #747, opened 2017-10-23
Enable override locale

compulim:
Locale was grabbed from `navigator`. Instructions added on how to override default locale.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,12 +172,6 @@ You can alter or add localized strings in [/src/Strings.ts](src/Strings.ts):
 * Add logic to map the requested locale to the supported locale in `strings`
 * Please help the community by submitting a [pull request](https://github.com/Microsoft/BotFramework-WebChat/pulls).
 
-You can specify locale to use in your website depends on your usage:
-
-* IFRAME: append `&locale=en-us` in query string
-* Inline JavaScript: add `{ locale: 'en-us' }` to options in `BotChat.App()` call
-* React: add `locale="en-us"` as a props to `<Chat>` component
-
 ### Behaviors
 
 Behavioral customization will require changing the TypeScript files in `/src`. A full description of the architecture of Web Chat is beyond the scope of this document, but here are a few starters:
--- a/README.md
+++ b/README.md
@@ -166,11 +166,17 @@ Web Chat strives to use responsive design when possible. As part of this, Web Ch
 
 ### Strings
 
-You can alter or add localized strings in `/src/Strings.ts`:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -166,12 +166,18 @@ Web Chat strives to use responsive design when possible. As part of this, Web Ch
...
-You can alter or add localized strings in `/src/Strings.ts`:
+You can alter or add localized strings in [/src/Strings.ts](src/Strings.ts):
 
 * Add one or more locales (with associated localized strings) to `localizedStrings`
 * Add logic to map the requested locale to the supported locale in `strings`
 * Please help the community by submitting a [pull request](https://github.com/Microsoft/BotFramework-WebChat/pulls).
 
+You can specify locale to use in your website depends on your usage:
+
+* IFRAME: append `&locale=en-us` in query string
```
danmarshall: Is this the Bot Framework embed Iframe?

compulim: Yes, adding new feature inside the IFRAME host.

billba: I'm confused by this. Are you talking about customizing locale inside of the WebChat Bot Framework channel? This repo is not oriented towards that channel.

danmarshall: I think we should remove this, because we are refraining from using URL parameters in the Iframe.

billba: Again I'm confused by the intention here. Is this about the channel or just our IFRAME sample? If the latter, that's just a sample and it shouldn't be referenced in the README.

compulim: The feature is in the code already. I will update one of the sample in the README.md to surface the feature.

samples/sidebar/index.html
```diff
@@ -101,6 +101,7 @@
         resize: 'detect',
         // sendTyping: true,    // defaults to false. set to true to send 'typing' activities to bot (and other users) when user is typing
         user: user,
+        locale: 'es-es', // override locale to Spanish
```
eanders-MS: This feels jammed in here because this isn't a sample made to highlight locale features. It also feels random.. why only this sample and not all of them? I agree it is a good idea to communicate through the samples that this parameter exists. I think it would be more consistent to add it to all samples, but commented out:
```ts
// locale: 'es-es', // optional: override locale
```

compulim: Agree, fixed. 😄

samples/backchannel/index.html
```diff
@@ -84,6 +84,7 @@
         bot: bot,
         botConnection: botConnection,
         user: user
```
eanders-MS: comma after `user` here :)

PR #753, opened 2017-10-26
Add Japanese locale

bigplants:
Hi,

I added Japanese locale strings.
Please check the strings and merge it.

### DIFF
--- a/src/Strings.ts
+++ b/src/Strings.ts
@@ -34,6 +34,21 @@ const localizedStrings: LocalizedStrings = {
         consolePlaceholder: "Type your message...",
         listeningIndicator: "Listening..."
     },
+    'jp': {
+        title: "チャット",
+        send: "送信",
+        unknownFile: "[ファイルタイプ '%1']",
+        unknownCard: "[不明なカード '%1']",
+        receiptVat: "消費税",
+        receiptTax: "税",
+        receiptTotal: "合計",
+        messageRetry: "再送",
+        messageFailed: "送信できませんでした。",
+        messageSending: "送信中",
+        timeSent: " %1",
+        consolePlaceholder: "メッセージを入力してください...",
+        listeningIndicator: "聴いてます..."
+    },
     'nb-no': {
         title: "Chat",
         send: "Send",
@@ -315,6 +330,8 @@ export const strings = (locale: string) => {
         locale = 'zh-hans';
     else if (locale.startsWith('cs'))
         locale = 'cs-cz';
+    else if (locale === 'jp')
+        locale = 'jp';
     else
         locale = 'en-us';
 
--- a/src/Strings.ts
+++ b/src/Strings.ts
@@ -322,17 +322,13 @@ export const strings = (locale: string) => {
         locale = 'el-gr';
     else if (locale.startsWith('it'))
         locale = 'it-it';
-    else if (locale === 'zh-yue')
-        locale = 'zh-yue';
     else if (locale === 'zh-hant' || locale === 'zh-hk' || locale === 'zh-mo' || locale === 'zh-tw')
         locale = 'zh-hant';
     else if (locale.startsWith('zh'))
         locale = 'zh-hans';
     else if (locale.startsWith('cs'))
         locale = 'cs-cz';
-    else if (locale === 'ja')
-        locale = 'ja';
-    else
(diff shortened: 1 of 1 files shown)


bigplants:
I fixed locale code.

src/Strings.ts
```diff
@@ -315,6 +330,8 @@ export const strings = (locale: string) => {
         locale = 'zh-hans';
     else if (locale.startsWith('cs'))
         locale = 'cs-cz';
+    else if (locale === 'ja')
+        locale = 'ja';
```
billba: This code literally does nothing, so there's no reason to add it.

Many other locales have regional variations. For instance, American English has minor spelling variations from British English. This pattern allows us start out returning a single set of strings for a locale, then add regional variations later if necessary.

If Japanese has no regional variations, then you can just provide a single set of strings and not add this pattern.

bigplants: @billba 
Thank you for reviewing my changes. I understood that I don't need to add `ja`.
But, I found a problem. If locale is not explicitly checked, it will always be `en-us`.
So, I'll fix the problem.

src/Strings.ts
```diff
@@ -307,15 +322,13 @@ export const strings = (locale: string) => {
         locale = 'el-gr';
     else if (locale.startsWith('it'))
         locale = 'it-it';
-    else if (locale === 'zh-yue')
-        locale = 'zh-yue';
```
bigplants: I found same case. So, I removed it.

src/Strings.ts
```diff
@@ -307,15 +322,13 @@ export const strings = (locale: string) => {
...
     else if (locale === 'zh-hant' || locale === 'zh-hk' || locale === 'zh-mo' || locale === 'zh-tw')
         locale = 'zh-hant';
     else if (locale.startsWith('zh'))
         locale = 'zh-hans';
     else if (locale.startsWith('cs'))
         locale = 'cs-cz';
-    else
+    else if (locale in localizedStrings === false)
         locale = 'en-us';
```
bigplants: If locale is not defined, It will be `en-us` as default.
Please let me know if it's not expected behavior.

bigplants:
@billba How do you think about my changes?

src/Strings.ts
billba: This is a better fix, thank you.

src/Strings.ts
compulim: Is there any reason why `zh-yue` is removed?

billba: My fault for approving this. I was complaining about code of this form:

```ts 
if (locale === 'foo')
    locale = 'foo';
```

because I found it confusing that the action was to assign locale to its current value. But obviously its important to catch this case.

I think a better version of this part of the logic is:

```ts
    ...
    else if (locale === 'zh-hant' || locale === 'zh-yue') {
        // leave alone
    } else if (locale === 'zh-hk' || locale === 'zh-mo' || locale === 'zh-tw')
        locale = 'zh-hant';
    else if (locale.startsWith('zh'))
        locale = 'zh-hans';
    ...
```

That way it's obvious what's going on.

However I think an even better fix is to break out the locale mapping logic into its own function so that the mapping is handled by a return rather than an assignment, which is more functional and (to my eye) clearer:

```ts
export const mapLocale = (locale: string) => {
    if (locale.startsWith('da'))
        return 'da-dk';
    if (locale.startsWith('de'))
        return 'de-de';
    ...
    if locale.startsWith('zh')) {
        if (locale === 'zh-hant' || locale === 'zh-yue')
            return locale;
        if (locale === 'zh-hk' || locale === 'zh-mo' || locale === 'zh-tw')
            return 'zh-hant';
        return 'zh-hans';
    }
    ...
    if (locale in localizedStrings)
        return locale;
    return 'en-us';
}

export const strings = (locale: string) => localizedStrings[mapLocale(locale)];
```

compulim: 💗 to make it functional programming. Let me make a PR for this.

compulim: #804 is the PR which fix this, along with your proposed `mapLocale` function.

bigplants: @compulim @billba I missed the conversation.. Thank you for fixing the fault. It has changed to be more readable 😃

PR #771, opened 2017-11-07
Fix suggested actions pushing behind header

### LINKED ISSUE #640: SuggestedActions pushes the top of the chat behind the header bar
When the div for suggested actions buttons is present, the message group ~~content~~ div is displaced upwards but doesn't take into account the header being there. So the top part of the first message will be stuck behind it, even if the user scrolls to the very top. It returns to normal once the suggested actions have disappeared.

- [Before Suggested Actions, scrolled to top](https://user-images.githubusercontent.com/11547914/29302493-b79d8abe-81d8-11e7-9295-191d0685995a.png)

- [During Suggested Actions, scrolled to top (first line of top message is stuck behind the header)](https://user-images.githubusercontent.com/11547914/29302489-b79a1938-81d8-11e7-8bca-6a37712d8e12.png)

edited for clarity

message group in correct position

message group displaced by suggested actions, so the top is blocked by the header

Paveltarno:
resolve #640

### DIFF
--- a/src/History.tsx
+++ b/src/History.tsx
@@ -24,6 +24,8 @@ export interface HistoryProps {
     doCardAction: IDoCardAction
 }
 
+const BOTTOM_SCROLL_TOLERANCE = 50; // This value should be at least as big as the histories container :before element
+
 export class HistoryView extends React.Component<HistoryProps, {}> {
     private scrollMe: HTMLDivElement;
     private scrollContent: HTMLDivElement;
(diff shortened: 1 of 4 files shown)


danmarshall:
Hi @Paveltarno - thanks for your contribution! :)
Can you also test this scenario when the Web Chat is initialized with the `{ showHeader: false }` option?

Paveltarno:
@danmarshall Iv'e tested it in both header configurations, works nicely

compulim:
Thanks for the fix @Paveltarno.

I updated your PR. Instead of using `40px` as "scroll to bottom" detection tolerance. We will look at whether there is a suggested actions bar just show up `hasActivityWithSuggestedActions`. If a suggested actions bar just show up, we will use `40px` as the tolerance. Otherwise, we will keep it as `1px`.

PR #786, opened 2017-11-16
Update to AdaptiveCards 1.0

danmarshall:
- [x] Update code to use latest package
- [ ] Run tests


package.json
```diff
@@ -29,10 +29,10 @@
   "license": "MIT",
   "dependencies": {
     "@types/react": "15.0.38",
+    "adaptivecards": "^1.0.0-beta2",
```
eanders-MS: Should this be non-carat so that it locks to the exact version?

danmarshall: fixed 👍

danmarshall:
PR sent to another branch.

PR #792, opened 2017-11-21
Disabling br Tags rendering

### LINKED ISSUE #658: Should <br/> be rendered in Markdown mode in Emulator?
When textFormat is set to "markdown", the code converts br tags to "\n\r"...to keep things straight, XML tags should not be accepted when textFormat is set to "markdown". Here is where that replace happens in the code:

https://github.com/Microsoft/BotFramework-WebChat/blob/ac769d76121134c9280d70909ffaf1b0e1acaf48/src/FormattedText.tsx#L57

----

Also - a related issue is that multiple line breaks are being replaced with just one. Meaning, if I sent "<br/><br/>" or "\n\n" then I expect to see two line breaks. Right now, it doesn't matter how many line breaks I put in, I'm only getting 1 line break in the Emulator rendering.

reyesrico:
Disables `<br/>` tag rendering.
Fixes #658


reyesrico:
Do not merge this PR until users are being notified.

compulim:
We have a way for the customer to provide their own setting of Markdown-It and even their own implementation of Markdown renderer. So this PR no longer applies.

PR #801, opened 2017-12-05
Add null check to shellRef callback.

marrobins:
Fix for issue Microsoft#782

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -230,6 +230,12 @@
         }
       }
     },
+    "append-field": {
+      "version": "0.1.0",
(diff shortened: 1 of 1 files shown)


src/Chat.tsx
```diff
@@ -117,6 +117,10 @@ export class Chat extends React.Component<ChatProps, {}> {
     }
 
     private saveShellRef(shellWrapper: any) {
+        if (!shellWrapper) {
+            this.shellRef = null;
+            return;
+        }
         this.shellRef = shellWrapper.getWrappedInstance();
```
compulim: IMHO, `this.shellRef = shellWrapper && shellWrapper.getWrappedInstance();` is a better solution. But I will accept this one.

PR #802, opened 2017-12-05
 Implementing CardImage tap actions and test. Fixes #663

### LINKED ISSUE #663: Need "tap" property on card images
CardImage objects should have a "tap" property which is a CardAction. 
This applies to any image in a card.

reyesrico:
- Includes call action.
- Includes openUrl and call action tests.

### DIFF
--- a/test/uitest.js
+++ b/test/uitest.js
@@ -77,9 +77,9 @@ describe('nightmare UI tests', function () {
 					}
 				});
 
-			let result = commands[cmd].client? 
-				yield nightmare.evaluate(commands[cmd].client) :
-				yield commands[cmd].evalOtherWindow(nightmare);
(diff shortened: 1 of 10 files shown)


test/uitest.js
```diff
@@ -73,8 +75,12 @@ describe('nightmare UI tests', function () {
...
 						}
 					}
-				})
-				.evaluate(commands[cmd].client);
+				});
+
+			let result = commands[cmd].client? 
+				yield nightmare.evaluate(commands[cmd].client) :
+				yield commands[cmd].evalOtherWindow(nightmare);
+
```
compulim: Try this.

```
let result = !commands[cmd].client || yield nightmare.evaluate(commands[cmd].client);

result &= !commands[cmd].evalOtherWindow || yield commands[cmd].evalOtherWindow(nightmare);
```

PR #804, opened 2017-12-07
Adding zh-yue back to the list

compulim:
Deleted in PR #753.

### DIFF
--- a/src/Strings.ts
+++ b/src/Strings.ts
@@ -420,7 +420,7 @@ export const defaultStrings = localizedStrings['en-us'];
 // e.g. if 'en-us' is the only supported English locale, then
 // strings('en') should return localizedStrings('en-us')
 
-export const mapLocale = (locale: string) => {
+function mapLocale(locale: string) {
     locale = locale && locale.toLowerCase();
 
     if (locale in localizedStrings) {
--- a/src/Strings.ts
+++ b/src/Strings.ts
@@ -420,53 +420,60 @@ export const defaultStrings = localizedStrings['en-us'];
 // e.g. if 'en-us' is the only supported English locale, then
 // strings('en') should return localizedStrings('en-us')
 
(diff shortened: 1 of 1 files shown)


src/Strings.ts
```diff
@@ -420,51 +420,60 @@ export const defaultStrings = localizedStrings['en-us'];
...
+        } else {
+            return 'pt-pt';
+        }
+    } else if (locale.startsWith('ru')) {
+        return 'ru-ru';
+    } else if (locale.startsWith('sv')) {
+        return 'sv-se';
+    } else if (locale.startsWith('tr')) {
+        return 'tr-tr';
+    } else if (locale.startsWith('zh')) {
```
billba: Pretty sure this logic isn't right yet.

billba: Need to make sure `zh-hant` returns `zh-hant`

compulim: I actually moved the `if (locale in localizedStrings)` check to the top, because it looks more efficient to do it there. And other cases listed here are just fallback.

billba: Ah, smart!

src/Strings.ts
```diff
@@ -420,51 +420,60 @@ export const defaultStrings = localizedStrings['en-us'];
...
-        locale = 'ja-jp';
-    else if (locale.startsWith('hu'))
-        locale = 'hu-hu';
-    else if (locale.startsWith('sv'))
-        locale = 'sv-se';
-    else if (locale.startsWith('tr'))
-        locale = 'tr-tr';
-    else if (locale in localizedStrings === false)
-        locale = 'en-us';
+export const mapLocale = (locale: string) => {
```
billba: This helper doesn't need to be exported.

compulim: Fixed.

Issue #811, opened 2017-12-12
src directory should be published in botframework-webchat package


billba:
When we switched from blacklist (`.npmignore`) to whitelist (`files` in `package.json`), the `src` folder got left out. The reason to have it there is to be able to debug using source maps.

compulim:
I think we should keep the `/src` folder away, but enable inline source code in the source map files. So practically, source code in `.ts` files will be inlined directly in the `.js.map` files.

What do you think?

billba:
Oh cool, show me how to do that and I'll make it so in DirectLineJS as well.

compulim:
I didn't test it out, but you should able to get it by modifying `tsconfig.json`. Change `sourceMap: true` to `inlineSourceMap: true`. You can look up at options in TypeScript doc [here](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

After you recompile, you can verify `*.js.map` files will contains source code from the original `*.ts` files, in addition to those `AACA;AACA` VLQ + Base64 code.

PiWiBardy:
hello i tried to remove sourcemap from the project, it's ok for the webchat project but i've the same issue with the adaptive cards project..

anyone is getting an idea?

billba:
@PiWiBardy if you have made this work for Web Chat please submit a PR. Adaptive Cards is just one of many dependencies of Web Chat, and we have no control over how they run their project.

PR #842, opened 2018-01-17
Replace SourceMap by InlineSourceMap for Production Package, and remo…

PiWiBardy:
…ve postinstall script that's break the install while using the package as a local dependency
#811

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "webpack": "webpack",
     "webpack-watch": "webpack --progress --watch",
     "clean": "rimraf built",
-    "postinstall": "cd test && npm install",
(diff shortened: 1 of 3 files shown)


compulim:
We cannot inline sources in `BotChat.js`, otherwise, it will bloat it up to > 5 MB.

We are going to include source map in 2 scenarios:
* `BotChat.js.map` will have TS source code inlined
* `built/*.js.map` will have TS source code inlined

In this way, we can make sure the user does not load big source map files when they do not need them.

I will credit you in the PR #878. Thanks for your effort for bringing this in.

I will also bring in your `postinstall` fix on another PR.

compulim:
PR #879 is the `postinstall` fix.

PR #845, opened 2018-01-19
Move @types to devDependencies

LarsLiden:
Prevents compile error when using different version of react

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
   "author": "Microsoft Corp",
   "license": "MIT",
   "dependencies": {
-    "@types/react": "15.0.38",
     "botframework-directlinejs": "0.9.13",
     "core-js": "2.4.1",
     "markdown-it": "8.3.1",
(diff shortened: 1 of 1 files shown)


LarsLiden:
@eanders-MS - look reasonable?

eanders-MS:
Change looks fine, but please preserve alphabetical order.

eanders-MS:
@danmarshall does this look ok to you?

danmarshall:
I am going to guess not. Here's what you should do to verify:
1. Get a clean clone of this.
2. Try to build.

I surmise that you wouldnt be able to build, because the TS compiler won't know what React is. So then, the dev is forced into installing React on their own.

danmarshall:
Actually, I correct myself: 
Create another project that uses Web Chat as a dependency. This new project would not build.

danmarshall:
See #598 and #705 which is a dupe of this PR

danmarshall:
I do believe React should be a Peer Dependency. There will probably be some tweaking to the build steps if we decide to do that.

compulim:
`peerDependencies` is the only way to do that. NPM will warn if the user did not select a version of React while they do `npm install`.

For your "create another project that uses Web Chat...", I will check it out but it should be fine. Because as we `npm install botframework-webchat` in, we are bringing the pre-built version.

corinagum:
Sorry that we never got this PR in. But in good news, this is resolved in v4! :)

PR #849, opened 2018-01-22
Adaptive Cards 1.0

compulim:
Bringing Adaptive Cards 1.0 including customizable and *live-reloading* `HostConfig.json`.

### DIFF
--- a/adaptivecards-hostconfig.json
+++ b/adaptivecards-hostconfig.json
@@ -90,7 +90,7 @@
       "actionMode": "inline",
       "inlineTopMargin": 8
     },
-    "actionsOrientation": "horizontal",
+    "actionsOrientation": "vertical",
     "actionAlignment": "stretch"
   },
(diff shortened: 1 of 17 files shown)


EduardTruuvaart:
:thumbsup: Any ideas on when that will be merged?

adaptivecards-hostconfig.json
```diff
@@ -1,174 +1,121 @@
...
+  "containerStyles": {
+    "default": {
+      "backgroundColor": "#00000000",
+      "foregroundColors": {
+        "default": {
+          "default": "#000000",
+          "subtle": "#808c95"
+        },
+        "accent": {
+          "default": "#2e89fc",
```
EduardTruuvaart: adaptivecards beta09 still seem to use "normal" instead of "default". It was changed in later versions of adaptive cards. Colours in current scheme will be defaulted to black as as properties will not match.

PR #865, opened 2018-02-19
Fixed speech synthesises on Safari for MacOS but not on iOS

DerPate2010:


compulim:
Looks good. I did some minor indentation formatting.

PR #810, opened 2017-12-12
Add customizable title

shade33:
#754

PR #875, opened 2018-03-06
Customizable title

compulim:
Credit @shade33 for the original PR #810.

Instead of having two props `showHeader?: boolean` and `title?: string`. We prefer a single one named `title?: boolean | string`.

Rule for `title`:
- `undefined` or `true` is passed, we will show the header with default text (locale-sensitive)
- `false` (or falsy except `undefined`) is passed, we will remove the header
- a string is passed, we will show the header with the text

A deprecation warning on `formatOptions` and `showHeader` is added.


compulim:
Deprecate whole `formatOptions` and move `title` to `ChatProps`.

PR #889, opened 2018-03-09
Fix #874 improve test facility

### LINKED ISSUE #874: Mock_dl server should echo activity of type "event"
This makes our backchannel sample more useful.

The sample page could also link to https://docs.microsoft.com/en-us/bot-framework/bot-service-design-pattern-embed-web-site, so people can dig around documentations about backchannel.

compulim:
Fix #874.

* Root `package.json` should run test scripts under `test/package.json`
* Localizing some dependencies (`/test` should not depends on root dependencies), still need more work here
* Use `cors` package, instead of making our own CORS because it don't work with pre-flight
* MOCK_DL will now echo back `EventActivity`

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -4,7 +4,8 @@
   "description": "> ## Disclaimer: This test framework is intended for Web Chat and DirectLine testing ONLY. This is NOT intended for testing Bots.",
   "main": "mock_dl/index.js",
   "scripts": {
-    "postinstall": "tsc -p ./mock_dl && tsc -p . && tsc -p ./mock_speech",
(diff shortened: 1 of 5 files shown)


test/package.json
```diff
@@ -19,11 +20,16 @@
     "@types/nightmare": "^2.10.0",
     "@types/node": "^8.0.56",
     "body-parser": "^1.18.2",
-    "botframework-directlinejs": "0.9.13",
+    "botframework-directlinejs": "^0.9.13",
+    "cors": "^2.8.4",
     "dotenv": "^4.0.0",
-    "express": "^4.16.2",
+    "exnode press": "^4.16.2",
```
danmarshall: typo?

compulim: Good catch. Type too fast on VSCode. 😜

PR #911, opened 2018-03-21
Polyfill Promise as needed

compulim:
We need Promise polyfill in IE11. Otherwise, it will fail `imBack` and `postBack` in `suggestedactions` tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ webpacked
 built
 botchat.js
 botchat.js.map
+botchat-es5.js
+botchat-es5.js.map
 CognitiveServices.js
 CognitiveServices.js.map
 *.css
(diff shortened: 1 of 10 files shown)


danmarshall:
How large is Bluebird? It is added to the bundle even if not used...

compulim:
Good call. It’s 80 KB minified.

https://unpkg.com/bluebird@3.5.1/js/browser/

What do you think?


Sent from my Windows 10 phone

________________________________

danmarshall:
I think we should document that `Promise` is a requirement, and that developers should add this as needed. We can provide a sample for polyfill.

eanders-MS:
I agree with Dan on that point.

billba:
That doesn't help people running the WebChat control in IE11.

Which continues to lead us towards that control being a separate project which depends on this one.

eanders-MS:
Oh that's an interesting option. `botframework-webchat-ie11`, or more generally `botframework-webchat-withpolyfills`.

billba:
Actually I just mean `botframework-webchat-control` which is the "just deploy it" version of the WebChat control. This is what would bundle in all necessary polyfills, the speech stuff, etc, and create the thing that is published to the CDN.

Then this repo can relax a bit and only do one job instead of two.

compulim:
Let me clear up our thoughts. Dev have few ways to use Web Chat:

* `<script src="botchat.js" />`: this is @billba `botframework-webchat-control` "just deploy it" version. This will include Promise polyfill
* `<iframe src="https://webchat.botframework.com/" />`: this use the same `botchat.js` approach, which is also polyfilled
* `npm install botframework-webchat`: this one does not polyfill, dev will need to bring their Promise implementation

Is that correct?

compulim:
Talked to @eanders-MS offline. We can do this:

* We will publish `botchat.js` (without polyfill) and `botchat-es5.js` (with polyfill) to our CDN
  * People using `<script>` can choose to have polyfill or not
* We will use `botchat-es5.js` on our embed page

Since users on NPM (i.e. `import 'botframework-webchat'`) are very likely building their website using Babel (or kinda), they should have polyfills. We will only offer non-polyfilled version, thus, we will not have a separate NPM package or Git repos.

danmarshall:
> We will publish botchat.js (without polyfill) and botchat-es5.js (with polyfill) to our CDN 

👍 Just need to add that to the `readme`.

billba:
Nice.

PR #937, opened 2018-04-14
Add listenFor speech recognizer hint

compulim:
`listenFor` is a hint for speech recognizer, suggesting possible wordings to improve accuracy. We support this feature for both Cognitive Services and Web Speech API + JSRF.

Note: DirectLineJS currently do not support `listenFor`.

### DIFF
--- a/src/Store.ts
+++ b/src/Store.ts
@@ -628,18 +628,28 @@ const stopListeningEpic: Epic<ChatActions, ChatState> = (action$, store) =>
 const startListeningEpic: Epic<ChatActions, ChatState> = (action$, store) =>
     action$.ofType('Listening_Starting')
     .do(async (action : ShellAction) => {
-        var locale = store.getState().format.locale;
-        var onIntermediateResult = (srText : string) => { store.dispatch({ type: 'Update_Input', input: srText, source:"speech" })};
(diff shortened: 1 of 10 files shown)


src/Store.ts
```diff
@@ -628,18 +628,28 @@ const stopListeningEpic: Epic<ChatActions, ChatState> = (action$, store) =>
...
-                onIntermediateResult(srText);
-                store.dispatch({ type: 'Listening_Stopping' });
-                store.dispatch(sendMessage(srText, store.getState().connection.user, locale));
-            };
-        var onAudioStreamStart = () => { store.dispatch({ type: 'Listening_Start' }) };
-        var onRecognitionFailed = () => { store.dispatch({ type: 'Listening_Stopping' })};
-
-        await Speech.SpeechRecognizer.startRecognizing(locale, onIntermediateResult, onFinalResult, onAudioStreamStart, onRecognitionFailed);
+        const { history: { activities }, format: { locale } } = store.getState();
+        const lastActivity = activities[activities.length - 1];
```
eanders-MS: lastActivity might be a trace or debug activity. I think you need to find the last activity of type `message`.

compulim:
Fixed, now using activity with `type === 'message'`, instead of any type of activity.

PR #938, opened 2018-04-14
Fix #924 upload button not show by default

### LINKED ISSUE #924: Upload Button
Hi, I've just pulled the last version of chat and can't find the upload button.
Is there any issue about it?

compulim:


src/Chat.tsx
```diff
@@ -86,7 +86,7 @@ export class Chat extends React.Component<ChatProps, {}> {
             this.store.dispatch<ChatActions>({ type: 'Set_Chat_Title', chatTitle });
         }
 
-        this.store.dispatch<ChatActions>({ type: 'Toggle_Upload_Button', showUploadButton: props.showUploadButton });
+        this.store.dispatch<ChatActions>({ type: 'Toggle_Upload_Button', showUploadButton: props.showUploadButton !== false });
```
danmarshall: so, if `showUploadButton` is `undefined`, you want it to be shown?

compulim: Yes, the default behavior should show upload button.

PR #954, opened 2018-04-25
Add support for OAuthCards to WebChat

Jeffders:
Adds support for OAuthCards to WebChat:

1. Add support for rendering OAuthCard attachments in a similar way to SignInCards, however, using Submit actions instead of just opening the URL because of #2
2. Add support for calling getSessionId from IBotConnection just prior to opening the OAuthCard link. When used with the DirectLine service, this method sets a security token in the browser to secure WebChat during OAuthCard sign-ins; basically this helps ensure the webchat browser where the OAuthCard was received is the same one where the link was traversed.

### DIFF
--- a/src/Chat.tsx
+++ b/src/Chat.tsx
@@ -327,7 +327,7 @@ export const doCardAction = (
             break;
         case "signin":
             let loginWindow =  window.open();
-            if(botConnection.getSessionId)  {
+            if (botConnection.getSessionId)  {
                 botConnection.getSessionId().subscribe(sessionId => {
                     konsole.log("received sessionId: " + sessionId);
                     loginWindow.location.href = text + encodeURIComponent('&code_challenge=' + sessionId);
--- a/src/Attachment.tsx
+++ b/src/Attachment.tsx
@@ -263,7 +263,7 @@ export const AttachmentView = (props: {
             if (!attachment.content)
                 return null;
             return (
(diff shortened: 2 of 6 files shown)


compulim:
Can you add some tests for your feature?

You can start by:
* Modify `test/commands_map.ts`
  * For quick turnaround time, temporary modify the very last lines `var testOnly` to include your new tests only
  * Add new commands
    * `server` code is what the server should send when the command is received
    * `do` code should use [Nightmare](http://www.nightmarejs.org/) to do extra client UI stuff
    * `client` code run inside Chrome (sandboxed), to assert on the UI
    * You can learn more from the command `hostconfig`, it's the simplest amongst them all
* `npm run build-test`
* `npm run test`

If you want to run your command manually:
* `npm run build-test`
* `npm run mock`
* Browse to [http://localhost:3000/?domain=http://localhost:3000/mock](http://localhost:3000/?domain=http://localhost:3000/mock)

src/Chat.tsx
```diff
@@ -323,9 +323,22 @@ export const doCardAction = (
         case "playVideo":
         case "showImage":
         case "downloadFile":
-        case "signin":
             window.open(text);
             break;
+        case "signin":
+            let loginWindow =  window.open();
+            if(botConnection.getSessionId)  {
```
billba: space between if and ( please :-)

Jeffders: No problem.

PR #966, opened 2018-04-30
Use tsconfig.json for definition files

compulim:


tsconfig.json
```diff
@@ -17,6 +17,7 @@
     },
     "files": [
         "src/adaptivecards-hostconfig.d.ts",
-        "src/BotChat.ts"
+        "src/BotChat.ts",
+        "src/jspeech.d.ts"
```
eanders-MS: Filename casing: Should be `JSpeech.d.ts`

eanders-MS: Oh I see now!

PR #976, opened 2018-05-09
Masking password input with backchannel event

enessoylu:
If you send the following event to WebChat through the backchannel, the input box will become a password input until you send a message.

Event:
```ts
{
    name: 'inputTypeChange',
    type: 'event',
    value: 'password'
}
```
Demo:
demo

### DIFF
--- a/src/ActivityView.tsx
+++ b/src/ActivityView.tsx
@@ -67,7 +67,7 @@ export class ActivityView extends React.Component<ActivityViewProps, {}> {
                 return (
                     <div>
                         <FormattedText
-                            text={ activity.text }
+                            text={ activity.inputHint === 'password' ? '•'.repeat(activity.text.length) : activity.text }
                             format={ activity.textFormat }
                             onImageLoad={ props.onImageLoad }
                         />
--- a/src/Chat.tsx
+++ b/src/Chat.tsx
@@ -6,7 +6,7 @@ import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
 
 import { Activity, IBotConnection, User, DirectLine, DirectLineOptions, CardActionTypes } from 'botframework-directlinejs';
-import { createStore, ChatActions, sendMessage } from './Store';
+import { createStore, ChatActions, ShellAction, sendMessage } from './Store';
 import { Provider } from 'react-redux';
 import { SpeechOptions } from './SpeechOptions';
 import { Speech } from './SpeechModule';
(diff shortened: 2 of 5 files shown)


billba:
Generally we encourage bot developers *not* to use bots for passwords, PII, and other secure data. Although Web Chat and Direct Line are secure, sending things like passwords through the same channel as the rest of the text tends to lead to bad situations. You wouldn't want passwords in LUIS utterances, for example, which are often made available widely to people in an organization.

But in this PR is a hint of a way forward. Just as you send a special type of activity *from* the server to initiate a password, you could also send one *to* the server with the password. This would allow you to create a secure pathway separate from the normal text pathway. This would have to follow all the best practices for passwords.

Do other channels have special affordances for passwords?

I would think of the server->client activity as an extension of Message indicating that the current message is requesting a password. But on the way back I'd either use a new 'password' Activity type or an 'event' Activity with name `password`.

(I'd want an actual security expert to look at this and chime in.)

Some notes about your implementation:
* Instead of a broad notion of "input types" I'd just make this about passwords. If it needs to be generalized later, we'll do it later.
* Instead of hacking `sendMessage` I'd add a `sendPassword` method.
* In terms of UX state I'd follow the lead of Suggested Actions, they're both similarly modal

@compulim @vishwacsena I think this idea is worth considering.

billba:
(Also, it’s better to raise an issue than go straight to a PR.)

compulim:
I understand the need to send secret over. A few things on UX:
1. How about sending secrets thru an Adaptive Card (or similar UI)?
  1. Separate UI make the user feel more secure than the send box
  2. We could hide/delete the UI after the secrets is sent
2. The user should be notified, before typing, that what they send will be masked
3. How about if the last message was activated by speech? The microphone icon will be pushed by default, need to handle that

We could use [`jsencrypt`](https://github.com/travist/jsencrypt) and use a public/private key to encrypt the data, just in case if they are not secure enough

compulim:
The feature did not go into Direct Line as a standard event.

Until we get this event into Direct Line, I prefer it to be done as a sample so people can "opt-in".

You can look at one of the samples that [customize activity](https://github.com/Microsoft/BotFramework-WebChat/tree/master/samples/activity-decorator-highlight). And implementing a new activity that render a password box.

Since the new sample would be very different from this PR, I am closing this one.

Please feel free to submit another PR for this work.

compulim:
This is the tracking issue for the sample, https://github.com/Microsoft/BotFramework-WebChat/issues/1413.

Issue #986, opened 2018-05-21
Props to turn off interactivity


compulim:
Turn off interactivity:
* [x] Hide send box
* [x] Hide suggested actions
* [x] Disable interactivity in all activities
   * [x] `<Media>`
   * [x] `<AdaptiveCardContainer>`

```js
<Chat interactive={ false } />
```

compulim:
Fixed in #988.

PR #988, opened 2018-05-22
Props to disable interactivity

compulim:
Fix for #986.

### DIFF
--- a/src/AdaptiveCardContainer.tsx
+++ b/src/AdaptiveCardContainer.tsx
@@ -13,8 +13,8 @@ import { AdaptiveCardsState, ChatState } from './Store';
 
 export interface Props {
     className?: string;
+    disabled?: boolean;
     hostConfig: HostConfig;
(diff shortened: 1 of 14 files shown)


src/AdaptiveCardContainer.tsx
```diff
@@ -173,8 +179,25 @@ class AdaptiveCardContainer extends React.Component<Props, State> {
             }
 
             if (renderedCard) {
+                if (!this.props.interactive) {
+                    const hyperlinks = renderedCard.querySelectorAll('a');
+                    const inputs = renderedCard.querySelectorAll('button, input, select, textarea');
+
+                    [].forEach.call(inputs, (input: HTMLInputElement) => {
```
tonyanziano: For consistency's sake:

on line 203 you use the convention `Array.prototype.forEach.call()`, but here you use `[].forEach.call()`

It would be nice to stick to one convention / pattern

PR #1008, opened 2018-06-07
Put back into Travis

compulim:


compulim:
Config change, self-approving.

PR #1010, opened 2018-06-07
Move version tag to before_install to prevent it run twice

compulim:


compulim:
Self-approving script changes.

PR #1012, opened 2018-06-07
Update README.md for CDN features

compulim:


compulim:
Self-approving README changes.

PR #1016, opened 2018-06-08
Add support for html text format

FrancisRD91:
Added support for text messages in HTML format, the functionality can be improved but it is a good basic start.


billba:
At present we have specifically and intentionally excluded HTML from messages (it's a long story).

In general, rather than submitting a PR for a new feature, it's good to file an issue for that feature so that it can be discussed first.

PR #1022, opened 2018-06-25
Typo Issue in timsent locale string with correct german fixed

jKelio:
I changed only one typo issue in the lcoale strings for german language at the timesent key.

In german for only time is "um" more correct than "am". If there is a date before the time, then "am" would be correct.


compulim:
Web Chat [0.14.1](https://www.npmjs.com/package/botframework-webchat/v/0.14.1) is out with the fix #1034.

PR #1054, opened 2018-07-22
Update NPM token

compulim:
Update NPM token due to ESLint security breach


compulim:
Self-approving non-code change.

PR #1063, opened 2018-07-31
Add AC Markdown tests and fix some tests

compulim:
- Added tests to make sure Markdown in Adaptive Cards are rendered correctly.
- Fixed some tests broken by Adaptive Cards.


compulim:
Self-approving non-production code and non-architectural changes.

PR #1078, opened 2018-08-14
Fixing code snippet

SherifElMahdi:
Fixing code snippet under "Easy: In your non-React website, run Web Chat inline" section. Replaced "secret" with "token" as writing it as "secret" never works.


corinagum:
Thank you for your submission. I'm sorry we didn't give feedback in a timely manner. Unfortunately this PR is outdated from the v4 code base, so I am closing.

PR #1107, opened 2018-08-27
Sort the activities by timestamp

ckkashyap:
Sort the activities by timestamp so that historic activities would show up above the newer activities.

### DIFF
--- a/src/Store.ts
+++ b/src/Store.ts
@@ -352,12 +352,21 @@ export const history: Reducer<HistoryState> = (
         }
         case 'Receive_Message':
             if (state.activities.find(a => a.id === action.activity.id)) { return state; } // don't allow duplicate messages
+            const activities = state.activities;
+            activities.push(action.activity);
+
+            const allActivities = activities.sort((l, r) => {
+                const ld = Date.parse(l.timestamp);
+                const rd = Date.parse(r.timestamp);
+
+                if (ld === rd) { return 0; }
+                if (ld < rd) { return -1; }
+                return 1; });
 
             return {
                 ...state,
                 activities: [
-                    ...state.activities.filter(activity => activity.type !== 'typing'),
-                    action.activity,
+                    ...allActivities.filter(activity => activity.type !== 'typing'),
                     ...state.activities.filter(activity => activity.from.id !== action.activity.from.id && activity.type === 'typing')
                 ]
             };


src/Store.ts
```diff
@@ -353,13 +353,20 @@ export const history: Reducer<HistoryState> = (
         case 'Receive_Message':
             if (state.activities.find(a => a.id === action.activity.id)) { return state; } // don't allow duplicate messages
 
+            const activities = [...state.activities];
+
+            if (activities.length > 0 && activities[activities.length - 1].type === 'typing') {
```
compulim:
```js
activities = activities.filter(existingActivity =>
  existingActivity.from.id !== action.activity.from.id
  && existingActivity.type === 'typing'
);
```

compulim:
Need to add some unit tests. I will do it, or if you want to give it a try, just let me know.

oflahero:
I hate to drag up old issues, but I encountered a problem with messages appearing out-of-order in my bot and narrowed it down to this commit.

I've built two versions of WebChat - one of the latest on master (d1a161a) and one of a local branch, from this commit, that reverts commit "Sort activities by timestamp" (a60d301).

View the former on https://ai-dev.acumy.com/Home/NotOutOfOrder, and the latter on https://ai-dev.acumy.com/Home/OutOfOrder. I'll try to keep them up as long as I can (dev website).

You should be able to reproduce within ten 'gambits' on the latter. I waited for a bot response each time before entering a new message.

Sample:

image

oflahero:
Note below that the activity IDs in the DOM are actually out of order:

image

compulim:
@oflahero due to #1127, we will revert the fix in `master` branch for now. We need to come up with a better fix for it.

compulim:
The problem is if there is a clock skew between the client and Direct Line service, the message sent by the user will be OOO.

The way to solve this: when we do insertion sort, ignore messages that are sending by the user. But right now in v3, we don't have a flag to see the status of the message (we do have that flag in v4). And thus, in order to fix this, there is a bigger cost than what we think.

Given "preserve history" feature is not going to production at Ignite, we are reverting the sorting part of code. Today, we will have "preserve history" feature in v4, and it will be in public preview pretty soon.

Thanks for debugging for us, @oflahero.

oflahero:
Thanks for all the info, William. Happy to help. I'll deactivate those debugging links soon. 'Preserve history' will be a great feature to have, though I appreciate how much work is probably involved.

compulim:
Thanks @oflahero.

We thought about few approaches:
* Client-side preserve history (the developer decide where to save it, probably `localStorage`)
   * This requires the developer have some experiences on how coding JavaScript
* Server-side upload history
   * The bot has some ability to upload past conversation and Direct Line will send it to Web Chat
   * Web Chat need to sort/dedupe the activity in correct order (that's the bug you are seeing today)

At the end of day, we decided to go server-side. And there are some rules on inactivity:
* 0 to 24 hours: Direct Line will automatically redownload 24-hour history from Direct Line
* 24 hours to 2 weeks: Bot need to upload past conversation manually
* After 2 weeks: conversation ID is gone, resume is not possible

We are still working on it. Our target is an end-to-end PoC that developers can understand, try out, and copy.

oflahero:
I'm no expert here but that sounds like the correct approach to me.
Essentially it will mimic what the other channels are doing - for example,
Skype and Messenger clients (app or desktop) currently sync with their
conversation data stores on Microsoft's and Facebook's servers,
respectively, in order to preserve conversation history. So users'
DirectLine clients will similarly sync with their own configured DirectLine
conversation history storage.

I imagine then that this will expose another interface for DirectLine
conversation history storage, the way IBotData does for BotFramework state.
The endpoint directline.botframework.com is just enabling infrastructure -
it can't be expected to host users' conversation history at production
level. E.g. IDirectLineConversationStore, and classes
SqlDirectLineConversationStore, AzureTableDirectLineConversationStore, etc.
- but configured at bot registration level instead of in bot code.

Anyway this is probably the wrong place for spouting off about it, I'll
leave you to it! Thanks.

PR #1132, opened 2018-09-04
Speech Recognizer Result Formats

ryanvolum:
This PR adds the ability to specify the result format that a developer wants from the Speech Recognizer service. Until now the result format has automatically just been ITN (Inverse Text Normalization). 

Supported result formats are now: 
* **MaskedITN** - automatically hides curse words: 
maskeditn

* **Lexical** - gives fully lexical representations of numbers: 
lexical

* **ITN** - gives number version of numbers: 
itn

* **Display** - Automatically punctuates transcription: 
image

The form still defaults to ITN, which calls the service with the "Simple" format, but using a different response form simply requires setting the resultForm property that now sits in the SpeechRecognizer interface:

```
    const speechOptions = {
      speechRecognizer: new CognitiveServices.SpeechRecognizer({
        subscriptionKey: 'YOUR_COGNITIVE_SPEECH_API_KEY',
        resultForm: 'MaskedITN'
      }),
```

This PR also adds comments in the speech sample to exemplify how to use a different result form.

### DIFF
--- a/samples/speech/index.html
+++ b/samples/speech/index.html
@@ -23,133 +23,136 @@
   
 -->
 <html>
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
-    <title>Bot Chat</title>
+<head>
+  <meta charset="UTF-8" />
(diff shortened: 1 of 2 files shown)


ryanvolum:
@compulim FYI!

compulim:
We talked offline, we are moving to the new SDK and it will shows result with Lexical, ITN, and Masked ITN. I will update you once it get there, very soon.

PR #1162, opened 2018-09-15
Hide post back activities by flag

compulim:
We were using `channelData.clientActivityId` to hide post back.

But in Chatdown, the transcript generated does not have `clientActivityId`, which is normal.

In Web Chat, we should change it to be a `postBack` flag to hide activities sent by the user instead.


compulim:
@tonyanziano this is the fix we talked about before, add a flag in Web Chat so it understand Chatdown correctly.

PR #1163, opened 2018-09-17
Fix #1127 Revert sort activities

### LINKED ISSUE #1127: Activity insertion sort should append before message from user due to clock skew
Code from v4 that works correctly:

```js
function upsertActivityWithSort(activities, nextActivity) {
  const {
    channelData: { clientActivityID: nextClientActivityID } = {},
    from: { id: nextFromID } = {}
  } = nextActivity;

  const nextTimestamp = Date.parse(nextActivity.timestamp);
  const nextActivities = activities.filter(({ channelData: { clientActivityID } = {}, from, type }) =>
    // We will remove all "typing" and "sending messages" activities
    // "clientActivityID" is unique and used to track if the message has been sent and echoed back from the server
    !(
      (type === 'typing' && from.id === nextFromID)
      || (nextClientActivityID && clientActivityID === nextClientActivityID)
    )
  );

  // Then, find the right (sorted) place to insert the new activity at, based on timestamp, and must be before "typing"
  // Since clockskew might happen, we will ignore timestamp on messages that are sending
  // If we are inserting "typing", we will always append it
  const indexToInsert = nextActivity.type === 'typing' ? -1 : nextActivities.findIndex(({ channelData: { state } = {}, timestamp, type }) =>
    (Date.parse(timestamp) > nextTimestamp && state !== 'sending' && state !== 'send failed') || type === 'typing'
  );

  // If no right place are found, append it
  nextActivities.splice(~indexToInsert ? indexToInsert : nextActivities.length, 0, nextActivity);

  return nextActivities;
}
```

compulim:
Due to #1127 and the complications of the fix, we are reverting #1107 for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Translations
   - `pt-BR`, by [Diego Castro](https://github.com/dfdcastro) in PR [#1074](https://github.com/Microsoft/BotFramework-WebChat/pull/1074)
-- Activities are now sorted by timestamp, by [CK Kashyap](https://github.com/ckkashyap) in PR [#1107](https://github.com/Microsoft/BotFramework-WebChat/pull/1107)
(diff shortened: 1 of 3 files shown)


compulim:
Self-approving reverts.

PR #1168, opened 2018-09-22
Update dependencies for samples

compulim:


compulim:
Approving non-code change

Issue #1125, opened 2018-08-30
Trying to render invalid AdaptiveCards blows up WebChat/Emulator


stevengum:
Trying to render improper AdaptiveCard attachments results in the following (using the v4 emulator, build 4.0.218-nightly):

![explode](https://user-images.githubusercontent.com/14935595/44883519-4a6e7b00-ac6c-11e8-9aed-2b53de0b9316.PNG)

The attachment looks like this:

```json
{
  "contentType": "application/vnd.microsoft.card.adaptive",
  "content": 5
}
```

This is the error from the dev tools console:

```bash
react-dom.production.min.js:151 TypeError: Cannot create property 'version' on number '5'
    at t.mountAdaptiveCards (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:1:117091)
    at t.componentDidMount (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:1:116352)
    at commitLifeCycles (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:72264)
    at x (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:60072)
    at w (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:58882)
    at g (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:58413)
    at _ (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:58250)
    at b (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:57629)
    at c (file:///C:/Program%20Files%20(x86)/Bot%20Framework%20Emulator/resources/app.asar/node_modules/@bfemulator/client/public/vendors.js:43:57091)
[...]

PR #1176, opened 2018-09-27
Add more error handling to mountAdaptiveCards

corinagum:
#1125 
@compulim could you let me know if this is what you were looking for for this fix? (also could you provide guidance on appropriate tests?)

### DIFF
--- a/src/AdaptiveCardContainer.tsx
+++ b/src/AdaptiveCardContainer.tsx
@@ -164,12 +164,16 @@ class AdaptiveCardContainer extends React.Component<Props, State> {
 
         adaptiveCard.hostConfig = this.props.hostConfig || defaultHostConfig;
 
-        let errors: IValidationError[] = [];
+        let errors: Array<IValidationError|Error> = [];
 
         if (!this.props.nativeCard && this.props.jsonCard) {
-            this.props.jsonCard.version = this.props.jsonCard.version || '0.5';
-            adaptiveCard.parse(cardWithoutHttpActions(this.props.jsonCard));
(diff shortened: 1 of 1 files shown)


src/AdaptiveCardContainer.tsx
```diff
@@ -164,12 +164,16 @@ class AdaptiveCardContainer extends React.Component<Props, State> {
...
         if (!this.props.nativeCard && this.props.jsonCard) {
-            this.props.jsonCard.version = this.props.jsonCard.version || '0.5';
-            adaptiveCard.parse(cardWithoutHttpActions(this.props.jsonCard));
-            errors = adaptiveCard.validate();
+            try {
+                this.props.jsonCard.version = this.props.jsonCard.version || '0.5';
+                adaptiveCard.parse(cardWithoutHttpActions(this.props.jsonCard));
+                errors = adaptiveCard.validate();
+            } catch (error) {
+                errors.push(error);
```
compulim: Can you verify how `errors` array get displayed on the screen?

Think about the upstream and downstream when changing the code:
- Downstream
   - Can the screen display `Error`?
   - Can it display `IValidationError` and `Error` at the same time?
   - What's the code over there? There is a command at MockBot v4 to force `IValidationError`, The command is called `card broken` and `card broken:1`
   - Also add a case to MockBot to make sure we capture this regression later
      - MockBot might already have this captured or not, I am not quite sure
- Upstream: there is no upstreamer in this case, so it's good

Also, try it in v4 by visiting https://webchat-playground.azure-websites.net/. If there is a bug over there, file another issue.

PR #1190, opened 2018-10-04
Sample: Speech thru Cognitive Services and browser support

compulim:
- Headless dictation component moved away from `<MicrophoneButton>` to `<Composer>`
   - Simplifying components
      - `<MicrophoneButton>` is a button and it should not know about interims
   - Let other components to trigger dictation thru Redux actions
      - Dev can start dictation programmatically
- Break `<DictationInterims>` out of `<MicrophoneButton>`
   - Dev can put `<DictationInterims>` at place they want (they can put up multiple of them)
- Sample for Cognitive Services
- Sample for browser-supported speech recognition


packages/component/src/Dictation.js
```diff
@@ -0,0 +1,97 @@
...
+  render() {
+    const {
+      props: { dictateState, disabled, webSpeechPonyfill }
+    } = this;
+
+    return (
+      <DictateComposer
+        onDictate={ this.handleDictate }
+        onError={ this.handleError }
+        onProgress={ this.handleDictating }
```
tonyanziano: Suggestion:

```
const {
  props: ...,
  handleDictate,
  handleError,
  handleDictating
}
...

<DictateComposer
  onDictate={ handleDictate }
  ...
```

compulim: Thanks for the suggestion. I will keep it as of today right now because there are no performance benefit if we change it in `Dictation.js`.

I will look at other code and see if we could improve it with this coding pattern.

PR #1197, opened 2018-10-10
Fix playground not building on Docker

compulim:
- Playground should build under `prepublishOnly`
- Docker build only after `prepublishOnly`


compulim:

PR #1203, opened 2018-10-12
Add nb-NO

compulim:
Added translation
- Norwegian Bokmål (Norway)


packages/playground/src/App.js
```diff
@@ -260,31 +260,32 @@ export default class extends React.Component {
                 value={ state.language }
               >
                 <option value="">Default ({ window.navigator.language })</option>
+                <option value="zh-HK">Chinese (Hong Kong)</option>
```
compulim: Re-ordering. For UX reason, we re-order based on the text, instead of the value, which is a little bit painful, but better for the user.

PR #1211, opened 2018-10-17
Bump test dependencies

compulim:
Fix vulnerability issues.


compulim:

PR #1213, opened 2018-10-17
Fix URL parsing in modern browsers

compulim:
Regression from recent fix.


compulim:
Self-approving this change. It broke video attachment now and the bug was a regression from bad code refactoring.

PR #1219, opened 2018-10-19
Lerna version should use exact version

compulim:
If `--exact` is not specified:
- `bundle` depends on `component@^4.0.0-preview.newer` and `core@^4.0.0-preview.newer`
- But since both are caret-ed version
- NPM will prefer version with `latest` tag, which means `@4.0.0-preview.veryold` instead


compulim:

PR #1224, opened 2018-10-21
Prefer user ID from Direct Line token

compulim:
- JWT generated from Direct Line will now contains "user" field, indicating the user ID associated with this token. More details can be found in this [article](https://blog.botframework.com/2018/09/01/using-webchat-with-azure-bot-services-authentication/)
   - If user ID is available from JWT, we should use it because it's more authoritative (not to mention its security nature)
      - JWT is from the hosted page, given it's developed by the same developer, it should be secure enough to trust the embedded user ID without verifying it
   - If there is a conflict between `props.userID` vs. user ID in JWT, we will prefer the one from JWT and give a warning
- Also, we are removing `username` because it was not used
   - For avatar initials, we are using values from `botAvatarInitials` and `userAvatarInitials`

(@mingweiw Web Chat will now prefer the user ID in JWT, than the one the user passed in. This is done automatically)

### DIFF
--- a/packages/component/src/Composer.js
+++ b/packages/component/src/Composer.js
@@ -35,9 +35,6 @@ import shallowEquals from './Utils/shallowEquals';
 const EMPTY_ARRAY = [];
 const NULL_FUNCTION = () => ({});
 
-const DEFAULT_USER_ID = 'default-user';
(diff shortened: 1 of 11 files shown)


mingweiw:
awesome! thanks!

PR #1225, opened 2018-10-21
Minor update to pl-Pl localization

peterblazejewicz:
- chat title translation fixed
- new key translated (x minutes ago)

Thanks!

### DIFF
--- a/packages/component/src/Localization/pl-PL.js
+++ b/packages/component/src/Localization/pl-PL.js
@@ -1,5 +1,5 @@
 export default {
-  'Chat': 'Chat',
+  'Chat': 'Czat',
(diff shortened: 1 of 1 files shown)


compulim:
@peterblazejewicz thanks for committing the PR.

I have commented out "X minutes ago", because it requires an implementation like [this](https://github.com/Microsoft/BotFramework-WebChat/blob/preview/packages/component/src/Localization/en-US.js#L1).

The reason why it need to be a custom implementations and cannot be generalized:
* Some languages have multiple plural form, e.g. Polish
* Some language prefer "a minute" over "1 minute"

I am merging this one in. Feel free to add another PR for the "X minutes ago". 😉

PR #1226, opened 2018-10-22
Revisit all breakaway components to make sure they are easy to recompose

compulim:
# Background

> Prerequisite: PR #1222

This PR is to revisit all UI components in Web Chat. We will determine which components to expose and how the user will use our code to recompose a new UI.

We have `Basic*` components in React, which is a "snap kit", containing multiple parts. They are similar to "[prototyping snap kit](https://www.sparkfun.com/products/14346)" or "snap model kits", but they works even before you snap them off.

Advanced user may want to rebuild the UI to fit their need. They snap parts off our kits and start building their own UI. They may use our parts, and they are also free to build their own. It requires minimal wiring code because we use [React context](https://reactjs.org/docs/context.html) to automatically wire things up.

We also provide business logic in some common components, e.g. `connectMicrophoneButton` will help the user build a new microphone button by providing the logic behind the "click" event, e.g. click to record, and click again to stop. The user can also build their own business logic using our [React context](https://github.com/Microsoft/BotFramework-WebChat/blob/preview/packages/component/src/Composer.js) directly.

## Transcripts

Currently, we baked too much logic in `BasicTranscript.js` and `activity/createCoreMiddleware.js` around "which activities to show" and "how to group timestamp". It is not trivial for the user to build a new UI with a totally new transcript interface.

We are splitting up some logic behind "which activities to show":
- Logic in `BasicTranscript.js` (more difficult to change and unlikely needed to change)
   - We are only showing "message" (non-empty/non-postback) and "typing" activities
      - We are also showing "self typing" activities, this will be filtered up to the middleware (i.e. the user)
- Logic in our default activity middleware (easy to change, and simpler to change)
   - We are hiding "self typing" activity

This will keep a better balance between code complexity and simplicity.

We could put all logics in our middleware so the user can change whatever they want, but the user will need to dupe most of the logic (some are uninterested to them) into their middleware. Instead of going this route, we prefer to separate out into two groups and put them into `BasicTranscript.js` and default middleware, accordingly.

## Connecting components

We refactored `connectXXX` function out of the component and exposed them. These functions expose the business logic behind the UI component, while designed carefully to be UI-agnostic. They are designed to help user to build new UI without the need to build business logic.

For example, `connectMicrophoneButton` will expose `dictating`, `disabled`, and, `onClick`. These are critical for `<MicrophoneButton>`. And it did not expose `styleSet` which is for UI.

If the user need more than we expose, they are free to pass in additional `selectors` to select more data from `<Composer>`. Or they can build their own UI using `<Composer>` directly.

They paved a way to bring in React Native components later.

## Things to do

- [ ] ~Update README.md for the new sample (copy from this PR)~
   - Not critical, we have a link in the README.md to refer to this article
   - When we clean up doc later, we will visit it
- [ ] (Optional) Revisit components again to make sure the architecture looks good
   - [ ] Customizable avatar
   - [ ] Customizable send status
   - [ ] Customizable timestamp

## Samples

- Breakaway speech button, a simple React web site that use 3 custom UI
   - Microphone button: a `<button>` to start/stop microphone input
   - Dictation interims: a `<p>` to show dictation interims
   - Last bot activity: a new component which show the last message activity from bot

## Changes

- Move `BasicSendBox.js` and `BasicTranscript.js` to top-level folder
- Move some filter logic from `BasicTranscript.js` to `createCoreMiddleware.js`
- Expose UI components so the user can recompose UI with our components easily
   - Transcript components
      - `<Avatar>`
      - `<Bubble>`
      - `<CarouselLayout>`
      - `<ErrorBox>`
      - `<SendStatus>`
      - `<SpeakActivity>`
      - `<StackedLayout>`
      - `<Timestamp>`
   - Send box components
      - `<DictationInterims>`
      - `<MicrophoneButton>`
      - `<SendButton>`
      - `<SendTextBox>`
      - `<SuggestedActions>`
      - `<UploadAttachmentButton>`
- Fix `sendTyping` not exposing on `propTypes`
- Removed "speech-only button" demo from playground
   - We now have a "breakaway speech button" sample
- Rename `connectWithContext` to `connectToWebChat`
   - This is the HOC function that wrap and enable React component with Web Chat functionality, e.g. `postActivity`


samples/breakaway-speech-button/public/index.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <meta name="theme-color" content="#000000">
+    <!--
+      manifest.json provides metadata used when your web app is added to the
+      homescreen on Android. See https://developers.google.com/web/fundamentals/web-app-manifest/
```
stevengum: What's this? Is this from another template somewhere?

compulim: Yes, this is done by `create-react-app`, the official + optional way for creating React app. Many people use it.

You don't really need `create-react-app` if you know Babel + Webpack (or if you don't need them). It's just conveniences for a lot of simple case. :)

PR #1256, opened 2018-10-30
Embed version via bundle

compulim:
# Background

## Diagnostic information thru `<meta>`

For debugging and diagnostic purpose, we will include versions and variant (`full`, `full-es5`, and `minimal`) in `<meta>` tag.

The version is read from `package.json` of corresponding project via Babel.

```html
<meta name="botframework-webchat:bundle:variant" content="full" />
<meta name="botframework-webchat:bundle:version" content="4.1.0-0" />
<meta name="botframework-webchat:core:version" content="4.1.0-0" />
<meta name="botframework-webchat:ui:version" content="4.1.0-0" />
```

## Rename `botchat-core.js` to `botchat-minimal.js`

Instead of `botchat-core.js`, we prefer `botchat-minimal.js`. This is because Web Chat Core is the name for the Redux store. We prefer to use the word "core" to represent the Redux store.


packages/bundle/src/addVersion.js
```diff
@@ -0,0 +1,24 @@
...
+      meta.setAttribute('content', content);
+
+      document.head.appendChild(meta);
+    }
+  } catch (err) {}
+}
+
+export default function (variant) {
+  setMetaTag('botframework-webchat:bundle:variant', variant);
+  setMetaTag('botframework-webchat:bundle:version', VERSION);
```
tonyanziano: where does the `VERSION` constant come from?

PR #1272, opened 2018-10-31
Use lerna for samples written in React

compulim:


compulim:
Self-approving non-production code change

PR #1278, opened 2018-10-31
Bump to 4.1.0-0

compulim:
The first release of Web Chat v4 is `4.1.0-0`.


compulim:

PR #1285, opened 2018-11-01
Use webchat.js as bundle name

compulim:
To prevent colliding with Web Chat v3, which use `botchat-*.js` as bundle name.


compulim:

PR #1286, opened 2018-11-01
Send welcome event and fixes around speech sagas

### LINKED ISSUE #1328: Speech: Microphone not respecting input hint
The microphone keeps activating and listening after a message is received, ignoring the message input hint value. This behavior is happening in the latest Chrome and Edge browsers at least.

I am using the the latest full bundle:
https://cdn.botframework.com/botframework-webchat/latest/webchat.js

### LINKED ISSUE #1371: Revert synthesized "conversationUpdate" activity
## Sample information
1. Sample type: Sample
2. Sample language: All languages
3. Sample name: welcome-users

## Describe the bug
In the emulator and _most_ other channels, the welcome-users sample proactively welcomes the user when the conversation is initiated. This leads to the (desirable) behavior of the bot giving the user some context of its capabilities before the user types anything. Without a proactive welcome, users are often left in the dark as to what they can/should do.

Unfortunately, when using the same welcome-users sample with webchat (and directline) as a client, the bot's behavior is different - it doesn't send a welcome message until a user says something. 

Any divergence in channel capabilities (especially a client/channel as popular as webchat/directline) should be documented. If the emulator uses webchat, and doesn't reflect the actual behavior of webchat, what is it really emulating? Developers tend to build their bot locally, and are then confused as to why it behaves differently when deployed. 

## To Reproduce
1. Create a bot using the ```welcome-users``` sample ([node](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/javascript_nodejs/03.welcome-users), [C#](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/csharp_dotnetcore/03.welcome-user))
2. deploy the bot
3. chat with the bot through webchat/directline
4. observe that the bot will not proactively welcome you

## Expected behavior
I expect the bot to welcome the user _before_ the user sends any message to the bot in Webchat. 

## Additional context
This issue of a bot initiating a conversation over directline has existed since the first iteration of the botbuilder. Without a user initiating a conversation, the bot doesn't have a conversationId to correctly route its message. @dandriscoll can probably provide more details! In the past I've implemented a workaround, where on initialization of the directline endpoint, the client sends an "invisible" backchannel event to the bot, which the bot then replies to. Unfortunately, this workaround is not documented and requires complex code changes to both the client and the server. 

Also, fwiw, this could have been a bug in the samples repo, botbuilder repo or in webchat
[bug]

compulim:
> Fix #1371 and fix #1328.

## Background

Since `conversationUpdate` was not designed for welcome scenario, it lacks of important features required for welcome, e.g. ability to send browser language and information to the bot as the first message.

This PR was designed to add a sample for send welcome event.

But during coding the sample, we talked to `redux-saga` team because one of their [design decision](https://github.com/redux-saga/redux-saga/issues/1651) is affecting our code. We settled with their design decision but that forced us to add an interim action (`CONNECT_FULFILLING`). We revisited all sagas to make sure they can handle successive actions properly.

While revisiting all sagas, we also identified few problematic areas in our sagas:

- First incoming activity came from user will throw null reference exception in saga
- Start/stop speech (TTS) dispatches were scattered across multiple sagas
   - `sendEvent` will stop speaking, behave same as other `postActivity` helpers
   - We will consolidate all stop speaking behavior in a new saga
- `via` argument in `setSendBox` is never used, removed
- Input hint was not honored in saga (#1328)
   - This could potentially separated in another PR, but since this is a one-line change and we touched all sagas around speech, we will leave it in for now

Although this PR was originally targeted to add a sample with minor-but-broad change on all sagas, since we practically touched almost every sagas, we also revisited problematic areas to make sure they are working as expected.

### Added
- Core: Added `sendEvent`, in PR [#1286](https://github.com/Microsoft/BotFramework-WebChat/pull/1286)
- Core: Added `CONNECT_FULFILLING` action to workaround `redux-saga` [design decision](https://github.com/redux-saga/redux-saga/issues/1651), in PR [#1286](https://github.com/Microsoft/BotFramework-WebChat/pull/1286)

### Fixed
- Core: Some sagas missed handling successive actions, in PR [#1286](https://github.com/Microsoft/BotFramework-WebChat/pull/1286)
- Core: `incomingActivitySaga` may throw null-ref exception if the first activity is from user, in PR [#1286](https://github.com/Microsoft/BotFramework-WebChat/pull/1286)
- Core: Fix [#1328](https://github.com/Microsoft/BotFramework-WebChat/issues/1328). Should not start microphone if input hint is set to `ignoringInput`, in PR [#1286](https://github.com/Microsoft/BotFramework-WebChat/pull/1286)

### Samples
- Backchannel: [Send welcome event](https://microsoft.github.io/BotFramework-WebChat/15.b.backchannel-send-welcome-event/), in PR [#1286](https://github.com/Microsoft/BotFramework-WebChat/pull/1286)


packages/bundle/package-lock.json
```diff
@@ -3624,12 +3624,14 @@
 				"balanced-match": {
 					"version": "1.0.0",
 					"bundled": true,
-					"dev": true
+					"dev": true,
+					"optional": true
```
compulim: `optional` is expected to be in `package-lock.json`, but a bug in NPM platform make them gone in previous commits.

packages/component/src/Dictation.js
```diff
@@ -29,7 +29,7 @@ class Dictation extends React.Component {
     props.stopDictate();
 
     if (transcript) {
-      props.setSendBox(transcript, 'speech');
+      props.setSendBox(transcript);
```
compulim: `setSendBox` no longer accept `via` argument. It was not used before, so it get cleaned up.

packages/core/src/reducers/activities.js
```diff
@@ -25,9 +25,14 @@ function findByClientActivityID(clientActivityID) {
 function upsertActivityWithSort(activities, nextActivity) {
   const {
     channelData: { clientActivityID: nextClientActivityID } = {},
-    from: { id: nextFromID } = {}
+    from: { id: nextFromID, role: nextFromRole } = {},
+    type: nextType
   } = nextActivity;
 
+  if (nextType === 'typing' && nextFromRole === 'user') {
```
compulim: When Direct Line platform echoed back the user typing activity, we should not add it because it was not intended to show up in the UI.

packages/core/src/sagas/clearSuggestedActionsOnPostActivitySaga.js
```diff
@@ -1,17 +1,23 @@
...
 import setSuggestedActions from '../actions/setSuggestedActions';
 
 export default function* () {
   yield whileConnected(function* () {
-    for (;;) {
-      yield take(({ payload, type }) => type === POST_ACTIVITY_PENDING && payload.activity.type === 'message');
-      yield put(setSuggestedActions());
-    }
+    yield takeEvery(
+      ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/markActivityForSpeakSaga.js
```diff
@@ -1,39 +1,29 @@
...
-    const { payload: { activity } } = yield take(
-      ({ payload: { activity } = {}, type }) =>
-        type === INCOMING_ACTIVITY
-        && speakableActivity(activity, userID)
-    );
-
-    yield put(markActivity(activity, 'speak', true));
-  }
+  yield takeEvery(
+    ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/sendEventToPostActivitySaga.js
```diff
@@ -0,0 +1,27 @@
...
+
+import whileConnected from './effects/whileConnected';
+
+import { SEND_EVENT } from '../actions/sendEvent';
+import postActivity from '../actions/postActivity';
+
+export default function* () {
+  yield whileConnected(function* () {
+    yield takeEvery(
+      ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/sendFilesToPostActivitySaga.js
```diff
@@ -9,16 +9,17 @@ import whileConnected from './effects/whileConnected';
...
 const getType = mime.getType.bind(mime);
 
 export default function* () {
   yield whileConnected(function* () {
-    for (;;) {
-      const { payload: { files } } = yield take(SEND_FILES);
-
-      if (files.length) {
+    yield takeEvery(
+      ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/sendPostBackToPostActivitySaga.js
```diff
@@ -1,19 +1,20 @@
...
+import whileConnected from './effects/whileConnected';
 
 export default function* () {
   yield whileConnected(function* () {
-    for (;;) {
-      const { payload: { value } } = yield take(SEND_POST_BACK);
-
-      if (value) {
+    yield takeEvery(
+      ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/sendTypingOnSetSendBoxSaga.js
```diff
@@ -30,8 +30,12 @@ export default function* () {
 
     for (;;) {
       let lastSend = 0;
-      const task = yield takeLatest(SET_SEND_BOX, function* ({ payload: { text } }) {
-        if (text) {
+      const task = yield takeLatest(
+        ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/startDictateAfterSpeakActivitySaga.js
```diff
@@ -1,38 +1,43 @@
...
+    yield whileSpeakIncomingActivity(startDictateAfterSpeakActivitySaga);
   });
 }
 
 function* startDictateAfterSpeakActivitySaga() {
-  for (;;) {
-    const { payload: { activityID } } = yield take(({ payload, type }) => type === MARK_ACTIVITY && payload.name === 'speak' && payload.value === false);
-    const activities = yield select(({ activities }) => activities);
+  yield takeEvery(
+    ({ payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/startDictateAfterSpeakActivitySaga.js
```diff
@@ -1,38 +1,43 @@
...
+      type === MARK_ACTIVITY
+      && payload.name === 'speak'
+      && payload.value === false
+    ),
+    function* ({ payload: { activityID } }) {
+      const { activities } = yield select();
```
cwhitten: Like an earlier comment, I'm favoring a defined selector that can represent the filter below and made available for reuse.

Is something like this possible?
```javascript
export const findActivity = state = id => (state.activities.filter(activity => (activity.id === id)))
```

```javascript
const activity = select(findActivity(id))
```

packages/core/src/sagas/startSpeakActivityOnPostActivitySaga.js
```diff
@@ -0,0 +1,25 @@
...
+import whileConnected from './effects/whileConnected';
+
+import { POST_ACTIVITY_PENDING } from '../actions/postActivity';
+import startSpeakingActivity from '../actions/startSpeakingActivity';
+import stopSpeakingActivity from '../actions/stopSpeakingActivity';
+
+export default function* () {
+  yield whileConnected(function* () {
+    yield takeEvery(
+      ({ meta, payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/stopDictateOnCardAction.js
```diff
@@ -10,18 +10,17 @@ import { POST_ACTIVITY_PENDING } from '../actions/postActivity';
...
-        ({ payload, type }) => type === POST_ACTIVITY_PENDING && payload.activity.type === 'message'
-      );
-
-      yield put(stopDictate());
-    }
+    yield takeEvery(
+      // Currently, there are no actions that are related to card input
+      // For now, we are using POST_ACTIVITY of a "message" activity
+      // In the future, if we have an action for card input, we should use that instead
+      ({ payload, type }) => type === POST_ACTIVITY_PENDING && payload.activity.type === 'message',
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

packages/core/src/sagas/stopSpeakActivityOnInputSaga.js
```diff
@@ -1,39 +1,39 @@
...
+import stopSpeakingActivity from '../actions/stopSpeakingActivity';
 
 export default function* () {
   yield whileConnected(function* () {
-    for (;;) {
-      yield take(
-        ({ payload, type }) => type === START_DICTATE
-        || (type === SET_SEND_BOX && payload.text && payload.via !== 'speech')
+    yield takeEvery(
+      ({ meta, payload, type }) => (
```
cwhitten: inlining this takes away from the readability for me. can it be defined outside the function and given a good name?

samples/05.c.presentation-mode-styling/index.html
```diff
@@ -30,8 +30,17 @@
 
         const res = await fetch('https://webchat-mockbot.azurewebsites.net/directline/token', { method: 'POST' });
         const { token } = await res.json();
+        const store = window.WebChat.createStore({}, ({ dispatch }) => next => action => {
+          // The following code is for convenient when trying out this sample, not required for production.
```
cwhitten: Can this comment be written like this?

`The following code is for convenience only. It not required for production.`

samples/07.customization-timestamp-grouping/index.html
```diff
@@ -87,8 +87,14 @@
 
         const res = await fetch('https://webchat-mockbot.azurewebsites.net/directline/token', { method: 'POST' });
         const { token } = await res.json();
+        const store = window.WebChat.createStore({}, ({ dispatch }) => next => action => {
+          // The following code is for convenient when trying out this sample, not required for production.
```
cwhitten: Can this comment be written like this?

`The following code is for convenience only. It not required for production.`

packages/core/src/sagas/markAllAsSpokenOnStopSpeakActivitySaga.js
```diff
@@ -0,0 +1,20 @@
...
+  select,
+  takeEvery
+} from 'redux-saga/effects';
+
+import { STOP_SPEAKING_ACTIVITY } from '../actions/stopSpeakingActivity';
+import markActivity from '../actions/markActivity';
+
+export default function* () {
+  yield takeEvery(STOP_SPEAKING_ACTIVITY, function* () {
+    const { activities } = yield select();
```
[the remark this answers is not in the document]
compulim: Added new `speakingActivity` predicate to "definitions" directory.

Instead of selector, we have "definitions" set up for this purpose. IMO, definition is better than selector because we can select/query by combining multiple definitions (and/or).

packages/core/src/actions/postActivity.js
```diff
@@ -5,9 +5,10 @@ export const POST_ACTIVITY_FULFILLED = `${ POST_ACTIVITY }_${ FULFILLED }`;
 export const POST_ACTIVITY_PENDING = `${ POST_ACTIVITY }_${ PENDING }`;
 export const POST_ACTIVITY_REJECTED = `${ POST_ACTIVITY }_${ REJECTED }`;
 
-export default function (activity) {
+export default function (activity, via = 'keyboard') {
```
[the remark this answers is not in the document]
compulim: Renamed.

packages/core/src/sagas/markActivityForSpeakSaga.js
compulim: Refactored out into a separate function.

packages/core/src/sagas/startSpeakActivityOnPostActivitySaga.js
compulim: Refactored out into a separate function.

packages/core/src/sagas/startDictateAfterSpeakActivitySaga.js
compulim: I agree with the previous one. But this one doesn't looks like have a lots of benefits to me.

For example, `findActivityByID` is a function that find an activity in a set by ID, and given ID is unique in the set.

The original code looks like:

```js
const spokenActivity = activities.find(({ id }) => id === activityID);
```

The new code would looks like:

```js
const spokenActivity = findActivityByID(activityID);
```

And the utility/selector function `findActivityByID.js` will looks like:

```js
export default function (activityID) {
  activities.find(({ id }) => id === activityID);
}
```

The benefit of finding one activity is not significant in this case, and also given that the schema is locked down, we are very unlikely to change how we handle activity ID.

Another potential possible move is a `activityHasID` predicate.

```js
const spokenActivity = activities.find(activity => activityHasID(activity, activityID));
```

And the predicate function `activityHasID.js` would looks like:

```js
export default function (activity, activityID) {
  return activity && activity.id === activityID;
}
```

This coding style is more targeted for adopting changing schema.

What do you think?

packages/core/src/sagas/incomingActivitySaga.js
```diff
@@ -20,38 +22,45 @@ function last(array, predicate) {
...
-
-        yield put(setSuggestedActions(actions));
-      }
-    });
+  yield whileConnected(observeActivity);
+}
+
+function* observeActivity(directLine, userID) {
+  yield observeEach(directLine.activity$, function* (activity) {
+    activity = patchActivityWithFromRole(activity, userID);
```
cwhitten: this suggests that all activities stored in state does have a `from` property. Is this is true why is the condition on L41 needed?

packages/core/src/sagas/incomingActivitySaga.js
```diff
@@ -20,38 +22,45 @@ function last(array, predicate) {
...
+
+    yield put(incomingActivity(activity));
+
+    // Update suggested actions
+    const { activities } = yield select();
```
cwhitten: I will disagree and commit here but it's odd that the convention is to pull _all_ state into scope and then parse it down when redux-saga provides a mechanism to document what part of the state it intended to be inspected via a selector. I see selectors and definitions working together, not mutually exclusive.

packages/core/src/sagas/incomingActivitySaga.js
```diff
@@ -20,38 +22,45 @@ function last(array, predicate) {
...
+      && lastMessageActivity.from.role === 'bot'
+    ) {
+      const { suggestedActions: { actions } = {} } = lastMessageActivity;
+
+      yield put(setSuggestedActions(actions));
+    }
   });
 }
+
+function patchActivityWithFromRole(activity, userID) {
```
cwhitten: can you provide context why this `from` role doesn't exist natively?

packages/core/src/sagas/incomingActivitySaga.js
compulim: Oh yea, that one can be omitted. Fixed.

packages/core/src/sagas/incomingActivitySaga.js
compulim: I am unsure if the selector inside `select()` function will do any perf benefit if we scope it down, that's why I am changing it back to the bare style. Syntax-wise, it looks more readable and compact,

```js
// Bare approach
const { activities } = yield select();
```

Compare to the original selector function,

```js
// Inline selector
const activities = yield select(({ activities }) => activities);
```

Yes, if we have the selector, then we can simplify it as (shortest after uglify):

```js
// Effect
const activities = yield selectActivitiesFromStore();
```

Or
```js
// Selector factory
const activities = yield select(activitiesSelector());
```

> I prefer selector factory here because we can add additional selectors like, `activitiesSelector(({ type }) => type === 'message')` or `activitiesSelector(speakableActivity)`.

What do you think about the 4th approach (selector factory)? The implementation would looks like,

```js
export default function (predicate) {
  return ({ activities }) => predicate ? activities.filter(predicate) : activities;
}
```

But when we need to combine multiple selectors, it will start look a bit cumbersome, or we need to do 2 calls to `select()`,

```js
// With combined selectors (better perf)
const { activities, language } = yield select(state => ({
   activities: activitiesSelector()(state),
   language: state.language
}));
```

```js
// With 2 calls (more readable)
const activities = yield select(activitiesSelector);
const language = yield select(languageSelector);
```

vs. the bare approach

```js
const { activities, language } = yield select();
```

The benefit here would be ability to quickly adapt to different store shape. The downside is more code to run.

What do you think?

packages/core/src/sagas/incomingActivitySaga.js
compulim: Added comment for the reason why we need patching, copied here.

> Some activities, such as "ConversationUpdate", does not have "from" defined.
> And although "role" is defined in Direct Line spec, it was not sent over the wire.
> We normalize the activity here to simplify null-check and logic later.

packages/core/src/sagas/startDictateAfterSpeakActivitySaga.js
compulim: Resolving this one because of dupe of another comment. Both talks about selector.

packages/core/src/sagas/incomingActivitySaga.js
compulim: @cwhitten I have updated this PR with selector.

Could you help me to review the change? I really want to make the selector architecture right, simple and intuitive to use.

To jumpstart you on the selector architecture:

- [`selectors/language.js`](https://github.com/Microsoft/BotFramework-WebChat/pull/1286/commits/f6737e332f3513df686dd787acd5641c8852e9de#diff-55132b6cc5589cd51c40c50eb106ccdbR1) is the simplest selector, getting the language settings
   - [`sagas/postActivitySaga.js`](https://github.com/Microsoft/BotFramework-WebChat/pull/1286/commits/f6737e332f3513df686dd787acd5641c8852e9de#diff-e5d56925571fab8eb14108fef46d8b23R43) is a sample usage of this selector
- [`sagas/startDictateAfterSpeakActivitySaga.js`](https://github.com/Microsoft/BotFramework-WebChat/pull/1286/commits/f6737e332f3513df686dd787acd5641c8852e9de#diff-95143c2160a015d4a649b9bf088b954fR35) is getting the first activity with ID of `activityID`
   - Note the use of rest operator here to get the first activity out of an array
- [`selectors/activities.js`](https://github.com/Microsoft/BotFramework-WebChat/pull/1286/commits/f6737e332f3513df686dd787acd5641c8852e9de#diff-25ad76dafa72bbef0076b54dd5195217R1) is our most complicated selector, it support a few things
   - Simply selecting all activities
   - Selecting activities using a predicate
   - Selecting activities of a specified type
   - Selecting activities of a specified ID (will return array)
      - Although ID is consider unique and should return a single result, prefer returning array for uniform result
- [`sagas/markAllAsSpokenOnStopSpeakActivitySaga.js`](https://github.com/Microsoft/BotFramework-WebChat/pull/1286/commits/f6737e332f3513df686dd787acd5641c8852e9de#diff-e220fa2015b4cb57f5fe472e3683a34aR19) is a sample usage of combining selector and definition

p.s. this is the last comment for this PR, but feel free to open more.

PR #1309, opened 2018-11-05
added heb support

geea-develop:
Added he-il lang support


compulim:
Looks good. Will merge after Travis greenlight.

PR #1327, opened 2018-11-08
French localization and suggestedActionBorderRadius style option

tao1:


packages/component/src/Localization/fr-FR.js
```diff
@@ -1,16 +1,41 @@
+function xMinutesAgo(date) {
```
compulim: Nice implementation.

Please use 2-space indent. 😉

packages/component/src/Styles/StyleSet/SuggestedAction.js
```diff
@@ -28,12 +29,14 @@ export default function createSuggestedActionStyle({
       '&:disabled': {
         background: suggestedActionDisabledBackground,
         border: suggestedActionDisabledBorder,
+        borderRadius: suggestedActionBorderRadius,
```
compulim: Maybe we should put this under `& > button` rule instead.

compulim:
Thanks. Love the border radius one, we missed that.

Comment inside. Please ignore coveralls for now, we are not guarding on our test numbers yet.

Thanks for opening the branch for editing. We will help you fix those comments if we have bandwidth.

PR #1357, opened 2018-11-14
RTL support && Hebrew language

geea-develop:


compulim:
@geea-develop the PR overall looks good, love to see RTL and Hebrew support.

I saw you added `direction` to the Redux store. But I didn't see where did you use it. Could you tell me more about it?

corinagum:
@compulim should we go ahead and remove direction & approve?

corinagum:
After speaking with @compulim offline, we'd like to wait to hear back from @geea-develop before proceeding -- it would be great to learn more about the requirements for rtl support.

packages/component/src/Composer.js
```diff
@@ -218,7 +219,11 @@ class Composer extends React.Component {
   }
 
   setLanguageFromProps(props) {
-    props.dispatch(setLanguage(props.locale || window.navigator.language || 'en-US'));
+    const lang = props.locale || window.navigator.language || 'en-US';
+    props.dispatch(setLanguage(lang));
+    if (['he', 'he-IL'].indexOf(lang) !== -1) {
+        props.dispatch(setDirection('rtl'));
```
geea-develop: @compulim here I added the dispatch for set Direction since it is only relevant for specific languages. 
Now that I think of it this RTL languages array can be moved to a utility, with an isRtl() method.
But otherwise, you can dispatch the setDirection from anywhere in your code.

geea-develop:
HI @compulim, thanks for the quick reply.
I actually added the setDirection in Composer.js under the setLanguage so its automatically updated base on the language you are in.
https://github.com/Microsoft/BotFramework-WebChat/pull/1357/files#r237997995

Obviously, more CSS changes will be needed as I added in my forked repos. But let's do that in another pull request.

packages/playground/src/App.js
```diff
@@ -140,8 +142,9 @@ export default class extends React.Component {
   }
 
   handleLanguageChange({ target: { value } }) {
-    this.setState(() => ({ language: value }), () => {
-      document.querySelector('html').setAttribute('lang', value || window.navigator.language);
+    const lang = value || window.navigator.language;
+    this.setState(() => ({ language: value, direction: this.getDirection(lang) }), () => {
```
compulim: Empty line after `const` and before `this.setState`.

compulim: Sorting by name, and split into multiple lines.

```suggestion
    this.setState(() => ({
      direction: this.getDirection(lang),
      language: value
    }), () => {
```

packages/core/src/reducers/direction.js
```diff
@@ -0,0 +1,17 @@
+import { SET_DIRECTION } from '../actions/setDirection';
+
+const DEFAULT_STATE = 'ltr';
+
+const DIRECTIONS = ['ltr', 'rtl'];
```
compulim: Probably can remove this one due to the comment below.

compulim:
I think some components might need to know about LTR/RTL, so that's why we have `rtl` as a Redux reducer.

corinagum:
@geea-develop hi thank you for the PR! Could you do a few things? 
Please add your username to the `LOCALIZATION.md` file on the Hebrew line, and add the following line to `CHANGELOG.md` under Unreleased Changes:

`- Add Hebrew localization by [@geeas-develop](https://github.com/geeas-develop) in PR [#1435](https://github.com/Microsoft/BotFramework-WebChat/pull/1435)`

compulim:
I have updated your PR with a few changes. Since the change is a bit big, I submitted a PR https://github.com/geea-develop/BotFramework-WebChat/pull/9 to your branch. Could you take some time to look if they looks good to you?

- Bump to Web Chat 4.2.0
- Instead of using an action/reducer for `setDirection`, I am exposing the `direction` property via React context (exposed by `Composer.js`).
   - The value is now computed out of `language` props, instead of setting into the Redux store.
   - I assume `direction` will be used by UI components inside Web Chat, so they can calculate correct layout.
   - UI components are recommended to use React context (exposed by `Composer.js`), instead of Redux store.
      - React context is the formal API of Web Chat, Redux store is the internal data representation for Web Chat.
      - React context is more-or-less computed from Redux store
      - Redux store may change from time to time to fit the internal need, but React context will be controlled
   - There are exceptions
      - If Redux saga (Web Chat logic code) need to know about the property, we should set it in Redux store for visibility
- Removed `state.direction` in playground
   - It is unused now.
   - For upcoming PR, if it's being used, please add it back.

compulim:
Sorry click wrong "Close and comment" button, reopening it.

geea-develop:
I am reviewing the changes and will get started on including the CSS changes required for the RTL to be usable.

corinagum:
@geea-develop Hey Guy, it's been a little while since we've had updates on this PR -- do you need anything from us or have any questions? Is there something we can help you with? Please let us know!

geea-develop:
@corinagum Hi I added the changes. might be a bit more than expected but had to change some of the buttons as well. Hope this helps.

geea-develop:
opening a new Pull request

PR #1406, opened 2018-11-28
Deploy with "legacy" tag

compulim:
NPM cannot use `v3` as a tag (it looks too similar to semver). We are changing that to `legacy`.


compulim:

PR #1438, opened 2018-12-05
Fix script to build GitHub pages

compulim:
We changed sample names but forgot to change the deployment script.


compulim:

PR #1451, opened 2018-12-07
Add playground to prepublishOnly

compulim:
Fix: add `playground` back to `prepublishOnly` and it would deploy correctly to https://webchat-playground.azurewebsites.net/


compulim:

Issue #1411, opened 2018-11-29
 [MAS1.4.3] Luminosity Contrast Ratio of 'Mic' button icons is less than required.


priysn:
Actual Result:
Luminosity Contrast Ratio of 'Send' button icon is 2.4:1.

Expected Result:
Luminosity Contrast Ratio of 'Send' button should be greater than or equal to 4.5:1.

Repro Steps:

1. Open URL [https://apac01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmicrosoft.github.io%2FBotFramework-WebChat%2Fspeech-cognitive-services-bing-speech%2F%3Fs%3D7c7dacfe04044f4f99a1edcf9b0e2ab9&data=02%7C01%7Cv-saugup%40microsoft.com%7C0f6f5e63167449daa46d08d65570f676%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636790341026155420&sdata=gcFq8JsflCLfUKAyw8f2A19aJcEQEFxGzjASMK9S59A%3D&reserved=0](https://apac01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmicrosoft.github.io%2FBotFramework-WebChat%2Fspeech-cognitive-services-bing-speech%2F%3Fs%3D7c7dacfe04044f4f99a1edcf9b0e2ab9&data=02%7C01%7Cv-saugup%40microsoft.com%7C0f6f5e63167449daa46d08d65570f676%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636790341026155420&sdata=gcFq8JsflCLfUKAyw8f2A19aJcEQEFxGzjASMK9S59A%3D&reserved=0) in Edge browser.
2. Notice 'Upload File' and 'Mic' button icons present in bottom of screen.
3. Open 'Colour Contrast Analyser' and select Foreground & Background color of mentioned elements one by one.
4. Notice value of Luminosity Contrast Ratio for Mic button.

User Impact:
Users with low vision or Color blindness will find it difficult to access content with low luminosity contrast ratio.

**Test Environment:**
**OS:** Windows 10
**OS Build:** 17134.345
**OS Version:** 1803
**Browser:** Edge
**Tool:** Color Contrast Analyser

**Attachment:**
![mas1 4 3_cca for mic button](https://user-images.githubusercontent.com/41668935/49214625-192d5480-f3ed-11e8-9faf-09f172e9acf2.png)

[...]

Issue #1341, opened 2018-11-13
[MAS1.4.3]Luminosity Contrast Ratio of 'Heading text' which appears after ''Adaptive cards that is broken because of invalid language' button is less than required. 


priysn:
**Actual Result:**
Luminosity Contrast Ratio of  'Text'  which appears after ''Adaptive cards that is broken because of invalid language' button is 4.0:1.

**Note** - Same issue repro at -
1. 'Text'  which appear after ''Adaptive cards that is broken because of invalid Version' button is 4.0:1.
2. Text 'No renderer for this attachment' appear after activating "Show an attachment of unknown type" button.

**Expected Result:**
Luminosity Contrast Ratio  for 'Heading Text - No renderer for this attachment'  should be greater than or equal to 4.5:1.

**Repro Steps:**

1. Open URL [https://microsoft.github.io/BotFramework-WebChat/full-bundle/](https://microsoft.github.io/BotFramework-WebChat/full-bundle/) in Edge browser.
2. Notice 'Upload File' and 'Send' button icons present in bottom of screen.
3. Type 'Help' in the input control at the bottom of the screen and hit enter (or click on send button).
4. Content will upload, Navigate to any "Adaptive cards that is broken because of invalid language" button by using Tab key and press Enter on it.
5. New data will upload.
6. Open 'Colour Contrast Analyser' and select Foreground & Background color of mentioned elements one by one.
7. Observe that Luminosity Contrast Ratio of text is less then 4.5:1.

**User Impact:**
Users with low vision or Color blindness will find it difficult to access content with low luminosity contrast ratio.

**MAS Reference:**
[MAS 1.4.3 – Contrast (Minimum) (17)](https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={a73546c7-765f-489c-b18f-afb659fe99e6})

**Test Environment:**
**OS:** Windows 10
**OS Build:** 17134.345
**OS Version:** 1803
**Browser:** Edge
**Tool:** Color Contrast Analyser

**Attachment:**
![mas1 4 3_contrast ratio for text](https://user-images.githubusercontent.com/41668935/48403206-6e156d80-e753-11e8-8b15-e0588d89df48.png)

compulim:
Please change the red color to `#EF0000`.

corinagum:
Related to #1354

Amit8527510735:
[...]

Issue #1316, opened 2018-11-05
[Usable]Name for 'Download icon' button is not properly defined.


priysn:
**Actual Result:**
When focus lands on 'download icon' in scan mode then narrator is reading "Empty paragraph".
Name property is not properly defined for 'Download icons'.

**Note -** Same issue repro at -
1. URL -> Help -> Document Data URI button/File attachments Button -> Download button icon - When focus lands on 'download icon' forward navigation of scan mode then narrator is reading "Empty paragraph", narrator not reading the name for download icon.

**Expected Result:**
When focus lands on 'download icon' in scan mode then narrator should not read "Empty paragraph".
Name property should be properly defined for 'download icons'

**Repro Steps:**

1. Open URL [https://microsoft.github.io/BotFramework-WebChat/full-bundle/](https://microsoft.github.io/BotFramework-WebChat/full-bundle/) in Edge browser.
2. Notice 'Upload File' and 'Send' button icons present in bottom of screen.
3.  Type 'accessibility' in the input control at the bottom of the screen and hit enter (or click on send button).
4.  Page will upload.
5. Open Narrator with scan mode (Caps+Space).
6. Navigate to "download icon" by using Up/Down Arrow keys.
7. Observe that Narrator is not reading name for "download icon" while navigate in the scan mode.

**User Impact:**
If Name property is not defined properly for control then the users will get confuse about the purpose of that control.

**Test Environment:**
**OS:** Windows 10
**OS Build:** 17134.345
**OS Version:** 1803
**Browser:** Edge
**Tool:** Narrator

**Attachment:**
[Usable_Name for download icon.zip](https://github.com/Microsoft/BotFramework-WebChat/files/2584552/Usable_Name.for.download.icon.zip)

Amit8527510735:
#HCL;#Accessibility;#HCL_BotFramework_WebChat;A11yUsable

PR #1494, opened 2018-12-13
Color contrast ratio fixes & downloadIcon narrator accessibility

### LINKED ISSUE #1305: [MAS1.4.3] Luminosity Contrast Ratio of 'Upload File' and 'Send' button icons is less than required.
**Actual Result:**
Luminosity Contrast Ratio of 'Upload File' button icon is 2.8:1 and 'Send' button icon is 1.7:1.

**Expected Result:**
Luminosity Contrast of 'Upload File' and 'Send' button icons should be greater than or equal to 4.5:1.

**Repro Steps:**
1. Open URL [https://microsoft.github.io/BotFramework-WebChat/full-bundle/](url) in Edge browser.
2. Notice 'Upload File' and 'Send' button icons present in bottom of screen.
3. Open 'Colour Contrast Analyser' and select Foreground & Background color of mentioned elements one by one.
4. Notice value of Luminosity Contrast Ratio.
 
**User Impact:**
Users with low vision or Color blindness will find it difficult to access content with low luminosity contrast ratio.

**MAS Reference:**
[MAS 1.4.3 – Contrast (Minimum) (17)](url)

**Test Environment:**
OS: Windows 10
OS Build: 17134.345
OS Version: 1803
Browser: Edge
Tool: Colour Contrast Analyser

**Attachment:**
mas1 4 3_botframework_webchat_uploadfile_send

corinagum:
Fix #1305, #1316,  #1341,  #1411

### DIFF
--- a/packages/component/src/Attachment/Assets/DownloadIcon.js
+++ b/packages/component/src/Attachment/Assets/DownloadIcon.js
@@ -1,7 +1,10 @@
 import React from 'react';
 
-export default ({ className, size = 1 }) =>
+import { localize } from '../../Localization/Localize';
+
+export default ({ className, language, size = 1 }) =>
   <svg
+    aria-label={ localize('Download file', language) }
     className={ className }
     height={ 20 * size }
     viewBox="0 0 12 20"
--- a/packages/component/src/Attachment/Assets/DownloadIcon.js
+++ b/packages/component/src/Attachment/Assets/DownloadIcon.js
@@ -1,10 +1,8 @@
 import React from 'react';
 
-import { localize } from '../../Localization/Localize';
-
-export default ({ className, language, size = 1 }) =>
+export default ({ className, label, size = 1 }) =>
   <svg
-    aria-label={ localize('Download file', language) }
+    aria-label={ label }
     className={ className }
     height={ 20 * size }
     viewBox="0 0 12 20"
--- a/packages/component/src/Attachment/DownloadAttachment.js
+++ b/packages/component/src/Attachment/DownloadAttachment.js
@@ -5,11 +5,12 @@ import connectToWebChat from '../connectToWebChat';
 import DownloadIcon from './Assets/DownloadIcon';
 
 export default connectToWebChat(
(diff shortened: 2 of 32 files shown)


packages/component/src/Attachment/DownloadAttachment.js
```diff
@@ -28,7 +29,7 @@ export default connectToWebChat(
                 <div className="size">{ format(size) }</div>
             }
           </div>
-          <DownloadIcon className="icon" size={ 1.5 } />
+          <DownloadIcon className="icon" language={ language } size={ 1.5 } />
```
cwhitten: instead of passing the language and letting `<DownloadIcon />` do the localization, could we just add a `label` prop and have `<DownloadAttachment />` pass in the already localized string? I like the separation there a bit better.

packages/component/src/Attachment/Assets/DownloadIcon.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
 
-export default ({ className, size = 1 }) =>
+import { localize } from '../../Localization/Localize';
+
+export default ({ className, language, size = 1 }) =>
   <svg
+    aria-label={ localize('Download file', language) }
```
a-b-r-o-w-n: Since this seems to be a presentational component, it may make more sense to do the localization up a level and just pass a `label` prop instead. That will give this component a bit more flexibility.

packages/component/src/Attachment/DownloadAttachment.js
corinagum: 👍

packages/component/src/Localization/sv-SE.js
```diff
@@ -3,6 +3,7 @@ export default {
   // 'Microphone off': '',
   // 'Microphone on': '',
   'Listening…': 'Lyssnar…',
+  'Download file': '',
```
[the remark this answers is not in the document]
corinagum: thanks for catching!

corinagum:
@cwhitten ping with PR fixes :)

PR #1495, opened 2018-12-13
Sample redirects

corinagum:


compulim:
Tested, it's working.

Now we need to cater this, https://github.com/Microsoft/BotFramework-WebChat/tree/master/samples/1.a.getting-started-full-bundle. How can it be redirected to 01.a.getting?

PR #1509, opened 2018-12-18
 Fix error with string interpolation

odysseus1973:


packages/component/src/Localization/ru-RU.js
```diff
@@ -1,16 +1,48 @@
...
+  } else if (deltaInHours < 5) {
+    return `${ deltaInHours } часа назад`;
+  } else if (deltaInHours <= 24) {
+    return 'Сегодня';
+  } else if (deltaInHours <= 48) {
+    return 'Вчера';
+  } else if (deltaInHours <= 72) {
+    return 'Позавчера';
+  } else {
+    return new Intl.DateTimeFormat('ru-RU').format(date);
```
compulim: If the browser doesn't support `Intl`, can you add code to fallback to simple format (although it might be not-quite-working), better than error out.

compulim:
👍 (Waiting for Travis CI to complete and will merge)

Thanks. I didn't realize we didn't handle peaceful fallback of `window.Intl` in `en-US`, we will add it on `en-US`.

PR #1538, opened 2018-12-22
Bump botframework-directlinejs in all packages/samples

cwhitten:


compulim:
Looks like this is missing `package-lock.json`, causing errors in `npm ci` in Travis CI.

If you have `package-lock.json`, please submit along. Otherwise, please do not modify `package.json` manually, always use either `lerna add` or `npm install` for dependency management. And also make sure `npm` is of latest version so it emit latest format of `package-lock.json`.

cwhitten:
Ah, yeah this isn't ready. The lockfile generation on Mac looks to be doing something different than lockfile generation on Windows and I was inspecting the diff a bit before pushing them up.

compulim:
Yea, it upset a lot of people when the fix is not in 6.4.1, including me.

As long as we are adding more `optional` than removing them, I am fine with it. And AFAIK, `npm ci` is fine with the missing `optional` too.

cwhitten:
closing in lieu of https://github.com/Microsoft/BotFramework-WebChat/pull/1544

PR #1544, opened 2018-12-26
Bump botframework-directlinejs in all packages/samples

cwhitten:


cwhitten:
@compulim does this LGTY?

PR #1558, opened 2019-01-08
Bump to adaptivecards@1.1.2

compulim:
## Background

Due to a security vulnerability issue found recently, we are bumping to `adaptivecards@1.1.2`.

Web Chat has 2 bundles: minimal and full. Minimal bundle does not include Adaptive Cards. And the full bundle have Adaptive Cards pre-configured with Markdown-It engine.

The security vulnerability only affect users who is using Adaptive Cards **without** any Markdown engine. By default, Web Chat ship and pre-configured with Adaptive Cards and Markdown-It, thus, the security vulnerability does not affect our default configurations. Only advanced users who manually configure Adaptive Cards explicitly without Markdown is affected.

### Changed
- `*`: Bump to [`adaptivecards@1.1.2`](https://npmjs.com/package/adaptivecards/), in [#1558](https://github.com/Microsoft/BotFramework-WebChat/pull/1558)

### Inclusion of `css-loader` and `style-loader`

Adaptive Cards 1.1.2 requires `css-loader`. The [code here](https://github.com/Microsoft/AdaptiveCards/blob/master/source/nodejs/adaptivecards/src/adaptivecards.ts#L6) read `import "./adaptivecards-default.css";`. Thus, in order to pack Adaptive Cards inside Web Chat, we have to add `css-loader` and `style-loader` in our pipeline.

Web Chat prefer bundler-independent:

- We reserve the flexibility to switch between [Webpack](https://webpack.js.org/), [rollup.js](https://rollupjs.org/) and other bundlers we see fit
- We prefer not to bundle any non-JavaScript assets in the bundle
   - Non-JavaScript assets usually assume the bundled code is running inside a browser (with a `window` object and corresponding functions to load different assets into memory)
   - Web Chat prefer to be platform-independent and reserve the flexibility to run on headless device

But since Adaptive Cards 1.1.2 requires `css-loader` and `style-loader`, we have to give up our bundler-independent flexibility. We have filed [bug #2279](https://github.com/Microsoft/AdaptiveCards/issues/2279) to Adaptive Cards team and see if they could remove the CSS or pre-compile it into JavaScript code before publishing their project to NPM to preserve their purity.

### DIFF
--- a/packages/component/package.json
+++ b/packages/component/package.json
@@ -61,7 +61,7 @@
     "typescript": "^3.1.6"
   },
   "dependencies": {
-    "adaptivecards": "1.1.2",
+    "adaptivecards": "~1.1.2",
     "botframework-webchat-core": "^0.0.0-0",
     "bytes": "^3.0.0",
     "classnames": "^2.2.6",
--- a/packages/component/package.json
+++ b/packages/component/package.json
@@ -61,7 +61,7 @@
     "typescript": "^3.1.6"
   },
   "dependencies": {
-    "adaptivecards": "^1.0.0",
+    "adaptivecards": "1.1.2",
     "botframework-webchat-core": "^0.0.0-0",
     "bytes": "^3.0.0",
     "classnames": "^2.2.6",
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -9,7 +9,7 @@
     "not op_mini all"
   ],
   "dependencies": {
-    "adaptivecards": "^1.0.0",
+    "adaptivecards": "1.1.2",
     "botframework-directlinejs": "^0.10.1",
     "botframework-webchat": "^0.0.0-0",
     "botframework-webchat-component": "^0.0.0-0",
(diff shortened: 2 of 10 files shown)


packages/playground/package.json
```diff
@@ -9,7 +9,7 @@
     "not op_mini all"
   ],
   "dependencies": {
-    "adaptivecards": "^1.0.0",
+    "adaptivecards": "1.1.2",
```
cwhitten: lets lock to the minor version

packages/component/package.json
```diff
@@ -61,7 +61,7 @@
     "typescript": "^3.1.6"
   },
   "dependencies": {
-    "adaptivecards": "^1.0.0",
+    "adaptivecards": "1.1.2",
```
cwhitten: lets lock to the minor version

packages/playground/package.json
compulim: Now on `~1.1.2`, lock to minor.

packages/component/package.json
compulim: Now on `~1.1.2`, lock to minor.

PR #1591, opened 2019-01-15
Send test image snapshots to Azure Blob for further study

compulim:
## Background

Sometimes, Travis CI fail because of difference in visual regression test.

This script will ship image snapshot results to our server.

After the build, the additional scripts will ZIP the image snapshots and send it to blob storage for further investigation. Like the one below,

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 # Do not commit binaries
 /chromedriver*
 chromedriver.exe
+
+/image-snapshots.zip
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,10 @@ script:
 - docker-compose down --rmi all
 - npm run coveralls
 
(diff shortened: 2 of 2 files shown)


corinagum:
dumb question, do we need some kind of policy that deletes the snapshots after a certain amount of time?

compulim:
I would love to and I thought that feature was there or at least in preview mode, https://feedback.azure.com/forums/217298-storage/suggestions/2474308-provide-time-to-live-feature-for-blobs.

But unfortunately, when I dig out, it is still in private preview (thru invitation only and limited data center).

So until they roll out this feature, we need to clean it up by hand, like monthly. But since this snapshots are tests and does not contains any personal data, I think it is fine right now to keep it there.

I structured the folder so it's easier for us to clean up when we need to.

PR #1604, opened 2019-01-17
[DRAFT] Allow ponyfilling window.open

### LINKED ISSUE #368: Allow ponyfilling window.open
Right now we call `window.open(url)` for card actions. The default in most browsers is to open the url in a new tab, but some might like to open it in a new window, etc.

My proposal is to add `urlTarget?:string` and `urlTitle?:string` to `ChatProps`. If supplied we'd pass them as the second and third parameter to `window.open`.

justinwilaby:
> Fix #368.

### DIFF
--- a/packages/component/src/Composer.js
+++ b/packages/component/src/Composer.js
@@ -96,9 +96,8 @@ function createCardActionLogic({ directLine, dispatch }) {
         case 'playAudio':
         case 'playVideo':
         case 'showImage':
-          // TODO: [P3] We should support ponyfill for window.open
-          //       This is as-of v3
-          window.open(value);
+          const factory = this.props.externalUrlHandlerFactory({ value });
+          factory && factory(value);
           break;
 
         case 'signin':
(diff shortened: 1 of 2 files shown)


packages/component/src/Composer.js
```diff
@@ -96,9 +96,8 @@ function createCardActionLogic({ directLine, dispatch }) {
         case 'playAudio':
         case 'playVideo':
         case 'showImage':
-          // TODO: [P3] We should support ponyfill for window.open
-          //       This is as-of v3
-          window.open(value);
+          const factory = this.props.externalUrlHandlerFactory({ value });
```
justinwilaby: @compulim - Let me know what you think should be included in the factory's function signature.

compulim: How about?

`windowOpenPonyfillFactory({ activity, attachment, cardAction }) => (url: string, windowName: string, windowFeatures: string) => void`

For the object returned from the ponyfill, we can assume we are not acting on it, until we need it. So I am leaving it `void` right now.

cwhitten:
@compulim @justinwilaby where are we at on this?

compulim:
Will fix in #1704. Closing this for now.

PR #1614, opened 2019-01-18
Pass `conversationId` to DirectLineJS constructor

neetu-das:
Added conversationId as an argument while instantiating DirectLine. Doing this will resume the conversation history related to a particular conversationId

### DIFF
--- a/packages/bundle/src/createDirectLine.js
+++ b/packages/bundle/src/createDirectLine.js
@@ -1,13 +1,13 @@
 import { DirectLine } from 'botframework-directlinejs';
 
-export default function ({ domain, fetch, secret, token, webSocket, conversationId }) {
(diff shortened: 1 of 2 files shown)


packages/bundle/src/createDirectLine.js
```diff
@@ -1,12 +1,13 @@
 import { DirectLine } from 'botframework-directlinejs';
 
-export default function ({ domain, fetch, secret, token, webSocket }) {
+export default function ({ domain, fetch, secret, token, webSocket, conversationId }) {
```
corinagum: Please sort the keys alphabetically :)

neetu-das: Sorted the keys alphabetically in Directline js constructor  and also updated CHANGELOG.md

PR #1616, opened 2019-01-18
Fix #1602: suggested-action regression 

### LINKED ISSUE #1602: Quotation marks around suggested action
Hey,

when using the "master" branch, the suggested actions are suddenly surrounded by quotation marks. The functionality is the same, it's just a style issue.

image

corinagum:
Fixes #1602 
`title` was not getting added to `postback` and `imback` action buttons.
- Also added an integration test to prevent this from happening in the future :)


packages/component/src/SendBox/SuggestedActions.js
```diff
@@ -48,8 +48,11 @@ export default connectSuggestedActions(
             buttonText={
               type === 'messageBack' ?
                 title || displayText
-              :
-                title || (typeof value !== 'string') ? JSON.stringify(value) : value
+              : title ?
+                  title
+                : (typeof value === 'string') ?
```
compulim: The indentation looks a bit weird to me.

__tests__/suggestedActions.js
```diff
@@ -9,6 +9,20 @@ function sleep(ms = 1000) {
 // https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebDriver.html
 
 describe('suggested-actions command', async () => {
+    test('should show correctly formatted buttons when suggested actions are displayed', async() => {
```
compulim: Can you change it from 4-space to 2-space? You find my VSCode extension to do the job.

packages/component/src/SendBox/SuggestedActions.js
a-b-r-o-w-n: Not only that, but a nested ternary is usually avoided. Maybe this can be extracted into a utility function and re-written as an if / else

corinagum:
ping @compulim @a-b-r-o-w-n, please let me know what you think about the changes :)

packages/component/src/SendBox/SuggestedActions.js
```diff
@@ -5,6 +5,18 @@ import React from 'react';
 import connectToWebChat from '../connectToWebChat';
 import SuggestedAction from './SuggestedAction';
 
+function renderSuggestedAction ({title, type, value}) {
+  if(type === 'messageBack') {
+    return title || displayText
```
a-b-r-o-w-n: `displayText` is undefined here right?

packages/component/src/SendBox/SuggestedActions.js
```diff
@@ -5,6 +5,18 @@ import React from 'react';
 import connectToWebChat from '../connectToWebChat';
 import SuggestedAction from './SuggestedAction';
 
+function renderSuggestedAction ({title, type, value}) {
+  if(type === 'messageBack') {
```
a-b-r-o-w-n: nit: space between `if` and `(`

packages/component/src/SendBox/SuggestedActions.js
```diff
@@ -46,10 +58,7 @@ export default connectSuggestedActions(
         ) =>
           <SuggestedAction
             buttonText={
-              type === 'messageBack' ?
-                title || displayText
-              :
-                title || (typeof value !== 'string') ? JSON.stringify(value) : value
+             renderSuggestedAction({title, type, value})
```
a-b-r-o-w-n: A comment about the naming of this function: when I see `renderBlah` I expect that function to return some jsx. In this case it just returns a string, so I would suggest renaming to something like `suggestedActionText`.

packages/component/src/SendBox/SuggestedActions.js
corinagum: oops thanks for the catch

corinagum:
@a-b-r-o-w-n @compulim ping :)

packages/component/src/SendBox/SuggestedActions.js
```diff
@@ -46,10 +58,7 @@ export default connectSuggestedActions(
         ) =>
           <SuggestedAction
             buttonText={
-              type === 'messageBack' ?
-                title || displayText
-              :
-                title || (typeof value !== 'string') ? JSON.stringify(value) : value
+             suggestedActionText({displayText, title, type, value})
```
compulim: Formatting: parenthesis between bracelets.

```suggestion
             suggestedActionText({ displayText, title, type, value })
```

packages/component/src/SendBox/SuggestedActions.js
```diff
@@ -5,6 +5,18 @@ import React from 'react';
 import connectToWebChat from '../connectToWebChat';
 import SuggestedAction from './SuggestedAction';
 
+function suggestedActionText ({displayText, title, type, value}) {
```
compulim: Code formatting.

```suggestion
function suggestedActionText({ displayText, title, type, value }) {
```

Issue #1521, opened 2018-12-20
Need Offline UI: Connection is taking longer than usual


corinagum:
Original issue is #135, which has been split into multiple issues. 

We need UI specs and implementation for:
- When the bot is offline, due to
   - Never able to connect successfully (or a definite result for not able to connect)
      - In the first 5 seconds, don't say it's not able to connect
      - UI for bad configuration (wrong secret or access token)
   - If it last for > 5 seconds , we should show an error
        - We need to find a good number, is 5 seconds 'longer than usual'?
        - UI for network error

### What we need
- [ ] Design specs
- [ ] Acceptance criteria
- [ ] T-shirt sizing

cwhitten:
@mewa1024 can we sit down and discuss this?

DesignPolice:
@cwhitten @mewa1024 I'm working today if this needs a visual. I have not experienced this error, so not sure what we are looking for, but happy to create something if we need it.

cwhitten:
Sure @DesignPolice that would be great! We need primarily three indicators:

1. A warning somewhere that WebChat is taking longer than usual to connect to a bot
2. A error saying that WebChat has failed to connect to a bot
3. A error saying the credentials used to connect to the bot are not working

corinagum:
Updated top comment with design specs and acceptance criteria. Please let me know your feedback. :)

DesignPolice:
@corinagum - It depends on how often this happens, seems like it might be nice to see it in line, vs disappearing.

mewa1024:
@corinagum @DesignPolice Hey all, quick question about the behavior: does Web Chat check for a connection periodically or only when the a message is sent? The reason I'm asking, if it's the latter case, then I think the error message should appear in place of the time stamp below the failed message, similar to the iPhone Message app and slack screens here:

![image](https://user-images.githubusercontent.com/39317542/50658983-f17f5e00-0f4f-11e9-8545-37e0ba6ec885.png)

![image](https://user-images.githubusercontent.com/39317542/50659017-0b20a580-0f50-11e9-9394-403bb4db7874.png)

If we do this, then the error message should scroll along with the message bubble. We might also want to allow the user to click on the error message to resend, similar to this feature, but not necessary:
![image](https://user-images.githubusercontent.com/39317542/50659065-360af980-0f50-11e9-9d03-81a47701724e.png)

[...]

PR #1649, opened 2019-01-27
Rework connect saga to support connectivity status and reconnection

compulim:
> This PR is pre-requisite of #1521 and related PRs.

Today, we do not support failed to connect initially and interrupted connection. Both features are also not supported in DLJS currently. In addition, we also do not support abort before a connection is made.

This PR will enable those scenarios. And we will work on support from DLJS side.

### Changed
- `core`: Reworked logic on connect/disconnect for reliability on handling corner cases, by [@compulim](https://github.com/compulim) in PR [#1649](https://github.com/Microsoft/BotFramework-WebChat/pull/1649)

### DIFF
--- a/packages/component/src/BasicWebChat.js
+++ b/packages/component/src/BasicWebChat.js
@@ -1,7 +1,5 @@
-import { createStore } from 'botframework-webchat-core';
 import { css } from 'glamor';
 import classNames from 'classnames';
-import memoize from 'memoize-one';
 import React from 'react';
 
 import BasicSendBox from './BasicSendBox';
@@ -30,7 +28,6 @@ export default class extends React.Component {
   constructor(props) {
     super(props);
 
-    this.createMemoizedStore = memoize(() => createStore());
     this.sendBoxRef = React.createRef();
 
     this.state = {
--- a/packages/core/src/sagas/connectSaga.js
+++ b/packages/core/src/sagas/connectSaga.js
@@ -131,11 +131,10 @@ function* connectSaga(directLine) {
 export default function* () {
   for (;;) {
     const { payload: { directLine, userID: userIDFromAction } } = yield take(CONNECT);
-    const userID = rectifyUserID(directLine, userIDFromAction);
     const updateConnectionStatusTask = yield fork(observeAndPutConnectionStatusUpdate, directLine);
 
     try {
-      const meta = { userID };
+      const meta = { userID: rectifyUserID(directLine, userIDFromAction) };
       let endDirectLine;
 
       yield put({ type: CONNECT_PENDING, meta });
@@ -167,8 +166,8 @@ export default function* () {
           yield take(negativeUpdateConnectionStatusAction);
         }
       } finally {
-        // It is meaningless to continue to use the Direct Line object even disconnect fail.
-        // And we will still unsubscribe to incoming activities.
+        // It is meaningless to continue to use the Direct Line object even disconnect failed.
+        // We will still unsubscribe to incoming activities and consider Direct Line object abandoned.
         yield put({ type: DISCONNECT_FULFILLED });
 
         endDirectLine();
--- a/packages/core/src/sagas/connectSaga.js
+++ b/packages/core/src/sagas/connectSaga.js
@@ -49,13 +49,9 @@ function* observeAndPutConnectionStatusUpdate(directLine) {
     for (;;) {
       const connectionStatus = yield call(connectionStatusQueue.shift);
 
-      console.log(`DirectLineJS connectionStatus updated to: ${ connectionStatus }`);
-
       yield put(updateConnectionStatus(connectionStatus));
     }
   } finally {
-    console.log(`Stop subscribing connectionStatus from DirectLineJS.`);
-
     connectionStatusSubscription.unsubscribe();
   }
 }
@@ -134,8 +130,6 @@ function* connectSaga(directLine) {
 
 export default function* () {
   for (;;) {
-    console.log('Waiting for CONNECT');
-
     const { payload: { directLine, userID: userIDFromAction } } = yield take(CONNECT);
     const userID = rectifyUserID(directLine, userIDFromAction);
     const updateConnectionStatusTask = yield fork(observeAndPutConnectionStatusUpdate, directLine);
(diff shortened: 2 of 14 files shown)


packages/component/src/BasicWebChat.js
```diff
@@ -30,7 +28,6 @@ export default class extends React.Component {
   constructor(props) {
     super(props);
 
-    this.createMemoizedStore = memoize(() => createStore());
```
a-b-r-o-w-n: Was this just no getting used?

packages/core/src/sagas/connectSaga.js
```diff
@@ -26,108 +25,155 @@ import {
...
 
 // const UNINITIALIZED = 0;
-// const CONNECTING = 1;
+const CONNECTING = 1;
 const ONLINE = 2;
-// const EXPIRED_TOKEN = 3;
-// const FAILED_TO_CONNECT = 4;
+const EXPIRED_TOKEN = 3;
+const FAILED_TO_CONNECT = 4;
 const ENDED = 5;
```
a-b-r-o-w-n: I think these are also exported from DirectLine

packages/core/src/sagas/connectSaga.js
```diff
@@ -26,108 +25,155 @@ import {
...
-        yield put({ type: CONNECT_REJECTED, error: true, meta, payload: new Error('cancelled') });
+    for (;;) {
+      const { payload: { connectionStatus } } = yield take(UPDATE_CONNECTION_STATUS);
+
+      // We will ignore DISCONNECT actions until we connect
+
+      if (connectionStatus === ONLINE) {
+        // TODO: [P2] DirectLineJS should kill the connection when we unsubscribe
+        //       But currently in v3, DirectLineJS does not have this functionality
+        //       Thus, we need to call "end()" explicitly
```
a-b-r-o-w-n: Have you logged an issue on the directline repo?

packages/core/src/sagas/connectSaga.js
```diff
@@ -26,108 +25,155 @@ import {
...
+
+        return () => {
+          activitySubscription.unsubscribe();
+
+          try {
+            // TODO: [P3] DirectLineJS will throw an error saying "conversation ended", it seems redundant
```
a-b-r-o-w-n: This was just fixed. We should publish a new version today.

packages/core/src/sagas/connectSaga.js
```diff
@@ -26,108 +25,155 @@ import {
...
+            directLine.end();
+          } catch (err) {}
+        };
+      } else if (connectionStatus === ENDED || connectionStatus === EXPIRED_TOKEN || connectionStatus === FAILED_TO_CONNECT) {
```
a-b-r-o-w-n: I prefer using `Array.includes` for this:

```js
[ENDED, EXPIRED_TOKEN, FAILED_TO_CONNECT].includes(connectionStatus)
```

It makes it easier to see the statuses we care about when they are right next to each other.

packages/playground/src/App.js
```diff
@@ -94,7 +95,7 @@ export default class extends React.Component {
         domain,
         fetch,
         token: directLineToken,
-        webSocket: webSocket === 'true' || +webSocket
+        webSocket: webSocket === 'true' || !!+webSocket
```
a-b-r-o-w-n: Could just be `Boolean(webSocket)`

packages/playground/src/App.js
```diff
@@ -147,6 +148,10 @@ export default class extends React.Component {
     this.setState(() => ({ botAvatarInitials: value }));
   }
 
+  handleDisconnectClick() {
+    this.props.store.dispatch({ type: 'DIRECT_LINE/DISCONNECT' });
```
a-b-r-o-w-n: Can you reference a constant here instead of `'DIRECT_LINE/DISCONNECT'`?

packages/playground/src/App.js
```diff
@@ -290,6 +295,12 @@ export default class extends React.Component {
           >
             Start conversation with local MockBot
           </button>
+          <button
+            onClick={ this.handleDisconnectClick }
+            type="button"
```
a-b-r-o-w-n: nit: this is redundant.

packages/component/src/BasicWebChat.js
compulim: Yes, it is dead code, not used.

packages/core/src/sagas/connectSaga.js
compulim: Yap, that's at https://github.com/Microsoft/BotFramework-DirectLineJS/issues/124.

packages/core/src/sagas/connectSaga.js
compulim: Okay, will wait for it.

packages/core/src/sagas/connectSaga.js
compulim: It looks expensive on the perf side (function call + object creation). Maybe I should line them up vertically so they are side-by-side too.

packages/playground/src/App.js
compulim: Today, it is either `?ws=true`, `?ws=0` or `?ws=1`. Otherwise, if we are treating it as boolean, I think we could just do `webSocket !== 'false'` (non-`true`/`false`, will by default turns to `true`).

packages/playground/src/App.js
compulim: Kind of, until we fix our reconnection story.

This is currently for Corina to test her connectivity status UI.

packages/core/src/sagas/connectSaga.js
compulim: Fixed.

packages/core/src/sagas/connectSaga.js
compulim: Fixed by aligning vertically for readability.

packages/playground/src/App.js
compulim: Currently, we did not expose any action type (constant).

We need a better plan on how/what to expose. Exposing everything will hurt the community than not exposing them at all.

Resolving this for now. If you think expose is important, please file a new bug and we discuss over there.

packages/core/src/sagas/connectSaga.js
compulim: Fixed, using those imported from `botframework-directlinejs`.

packages/core/src/sagas/connectSaga.js
compulim: Fixed, by removing the try-catch.

packages/playground/src/App.js
a-b-r-o-w-n: I specifically mean `!!+webSocket` can be `Boolean(webSocket)`

PR #1653, opened 2019-01-28
update docs to include selectable activity link

a-b-r-o-w-n:


compulim:
Test failed due to timeout on the bot (> 5 seconds on first message).

If you see something similar to this, try hit this URL, https://webchat-mockbot.azurewebsites.net/. If you keep <kbd>F5</kbd> and see it respond pretty quickly, that means MockBot is in great shape (and probably Direct Line platform is not).

image

We will work on test reliability in 4.4. It was working great before but Direct Line platform looks pretty flaky recently.

PR #1663, opened 2019-01-30
Fix typo polyfills

nt-7:
ponyfill -> polyfills


DaveVoyles:
Thank you for clarfying, I appreciate it! A co-worker had come across this and we thought _"this can't be right..."_

PR #1670, opened 2019-01-30
Extend style options to support basic adaptive cards style

a-b-r-o-w-n:
Currently only passes font family and font colors through

# Added
- `component`: Allow font family and adaptive cards text color to be set via styleOptions, by [@a-b-r-o-w-n](https://github.com/a-b-r-o-w-n), in PR [#1670](https://github.com/Microsoft/BotFramework-WebChat/pull/1670)


packages/component/src/Styles/adaptiveCardHostConfig.js
```diff
@@ -1,99 +1,35 @@
...
         }
       }
+    },
+    emphasis: {
+      default: {
+        default: bubbleTextColor,
+        subtle: subtle
+      }
     }
   },
```
a-b-r-o-w-n: @compulim I'd like your opinion on this.

I made these changes based on the themes in Emulator. I think that allowing users to override the host config handles the extreme case of needing to fully customize the adaptive cards config. This change only passes a few things from the style options through.

Let me know what you think. We could extend this and add more options to styleOptions, but I thought that would not be necessary since the override already exists.

a-b-r-o-w-n:
@compulim or @corinagum How should I proceed with the failing image snapshots?

corinagum:
@a-b-r-o-w-n were there slight color changes? that's the only difference I can see in the snapshots. 
https://webchattest.blob.core.windows.net/travis-result/2019/01/31/gy262/index.html
If that's the case, you can delete the failing images and run the tests locally again to push up the new png's. :)

a-b-r-o-w-n:
@corinagum ah. yes. probably because of this: https://github.com/Microsoft/BotFramework-WebChat/pull/1670/files#diff-27cc33be5272a601ab9ad5a925024b12R20. I'll revert that and see if it passes.

packages/component/src/Styles/StyleSet/DownloadAttachment.js
```diff
@@ -1,12 +1,11 @@
-import { primaryFont } from '../Fonts';
-
 export default function ({
   accent,
   bubbleTextColor,
-  paddingRegular
+  paddingRegular,
```
compulim: No need `paddingRegular` here.

packages/component/src/Styles/adaptiveCardHostConfig.js
```diff
@@ -1,99 +1,35 @@
 import { HostConfig } from 'adaptivecards';
 import defaultStyleSetOptions from './defaultStyleSetOptions';
-import { primaryFont } from './Fonts';
-
 // https://docs.microsoft.com/en-us/adaptive-cards/rendering-cards/host-config
 
-export default ({ accent } = defaultStyleSetOptions) => new HostConfig({
+export default ({
+  accent,
+  bubbleTextColor ,
```
compulim: Comma indentation.

PR #1678, opened 2019-01-31
add chat send history sample

### LINKED ISSUE #1091: Up/down arrow for send history
What should up/down arrow do when:

- [ ] Long lines (we don't support multiple lines right now)
- [ ] Suggested actions
- [ ] `imBack`

a-b-r-o-w-n:
closes #1091 

### Added
- `samples`: Added chat send history sample, by [@a-b-r-o-w-n](https://github.com/a-b-r-o-w-n), in PR [#1678](https://github.com/Microsoft/BotFramework-WebChat/pull/1678)


cwhitten:
I'm unclear why this isn't provided by the component out of the box!

a-b-r-o-w-n:
@cwhitten one thing I can think of in opposition to adding this to the bundle is concerns over security / managing the session state. Would it be by user / conversation / bot / something else? IMO this is better suited as a custom add on that suits the customer's need. Also, while this feature is helpful for testing, I am not convinced that it is useful for the majority of bot interactions.

PR #1679, opened 2019-01-31
Offline UI - indicate slow connection and failed credentials

### LINKED ISSUE #1521: Need Offline UI: Connection is taking longer than usual

corinagum:
Fix #1521 
- Update Italian localization, courtesy of andreo
- @compulim, can you go through the Chinese localization files and add to `FAILED_CONNECTION_NOTIFICATION` and `SLOW_CONNECTION_NOTIFICATION`? I'll update my PR with your changes.


cwhitten:
@corinagum can you provide some screenshots in the descriptions of the pull request? thanks.

packages/core/src/reducers/connectivityStatus.js
```diff
@@ -0,0 +1,37 @@
...
+      state = 'connectingslow';
+      break;
+
+    case DISCONNECT_FULFILLED:
+      state = 'notconnected';
+      break;
+
+    default: break;
+  }
+  return state
```
tonyanziano: missing `;`

cwhitten:
@mewa1024 @DesignPolice given that WebChat isn't a public facing name given to the Chat window, should we make the copy more generic? Take BuzzFeed's recent [Zo bot](https://www.buzzfeed.com/jasonstangelo/valentines-day-gifts-2019-microsoft-zo). It would be odd to see "WebChat" in error/warning copy here. But I could imagine more generic copy that doesn't call out WebChat specifically.

mewa1024:
@cwhitten Good point. How about changing the messages to "Taking longer than usual to connect" and "Unable to connect"? @DesignPolice, thoughts?

DesignPolice:
@cwhitten @mewa1024 @corinagum - That is true Chris, I forget that we need to be more generic with so many different uses of our product. That seems fine to me Mewa.

corinagum:
I'll update the copy shortly!

corinagum:
Update: I've updated the copy and fixed the build, but I'm having trouble getting tests to run locally so that I can update the png's for the ConnectivityStatus tests. Will check back in a bit.

PR #1682, opened 2019-02-01
Add username to activity for parity with v3

### LINKED ISSUE #1321: Set the user name in from in new version
I was able to set the user ID in the new version of web chat, but I can't figure out how to set the user's name. I've got some logic that relies on the user's name in a bot and need to set this before migrating to webchat v4. Is there a way to set this?

compulim:
> Fix #1321.

### Added
- `*`: Added `username` back to activity, fixed [#1321](https://github.com/Microsoft/BotFramework-WebChat/issues/1321), by [@compulim](https://github.com/compulim), in PR [#1682](https://github.com/Microsoft/BotFramework-DirectLineJS/pull/1682)


packages/core/src/sagas/connectionStatusUpdateSaga.js
```diff
@@ -12,7 +12,7 @@ export default function* () {
   yield whileConnected(observeConnectionStatus);
 }
 
-function* observeConnectionStatus(directLine) {
+function* observeConnectionStatus({ directLine }) {
```
cwhitten: this looks out of place. it doesn't look like the shape of the parameter change?

cwhitten: now i see it - nevermind!

PR #1685, opened 2019-02-01
Revert `conversationUpdate` and update tests

### LINKED ISSUE #1371: Revert synthesized "conversationUpdate" activity

compulim:
> Fix #1371.

## Background

This will revert the recent introduction of synthesized `conversationUpdate` activity.

Instead, we will suggest users to use our ["welcome event" sample](https://github.com/Microsoft/BotFramework-WebChat/tree/master/samples/15.d.backchannel-send-welcome-event), which offer more functionalities especially around I18N.

## Implementations

- [x] Remove `sendConversationUpdateSaga`

### Tests update

Due to a bug in Direct Line token API, although we prime the channel with an activity, Direct Line platform do not send `conversationUpdate(user)` to the bot. We are removing the "Welcome to Mockbot!" from our test snapshot. Instead, we now have a programmatic way to check if the bot is ready to chat or not.


tonyanziano:
So basically all the calls verifying the number of activities sent are being reduced by 1 because there is no longer a conversationUpdate activity, correct?

compulim:
@tonyanziano Yes, I think all of them.

Issue #1423, opened 2018-11-30
Sample: Host Web Chat with Angular5


corinagum:
@omarsourour, let's coordinate to make sure we are on the same page for how to implement this sample.

corinagum:
> Based off email discussion, it looks like this sample would be great as an addition to the currently available [samples](https://github.com/Microsoft/BotFramework-WebChat#samples-list), hosted and with a live bot for testing. I made a placeholder for it called `03.b.host-with-Angular5`

Notes from @compulim: 
We want the samples to be as simple as possible. For example, in some of our React demos, we are inlining Babel + React via `<script>` tag, instead of building a full-blown [app] using create-react-app.
-	We prefer more readability, "quick-into-code", than spending time finding the right file to look at 
	  - Measurement: under a sample project root, can the user find the right file to look at under 10 seconds?
- Even though the code may need heavy modifications in production environment, I think it is fine, because samples are of POC quality, not production quality
- Can the user copy-and-paste our code snippets into their production environment? 
     - If the user cannot copy the portion without code reviewing them, we will tell them what things that need to consider
    - For example, in one of our work-in-progress sample called "[Direct Line token server](https://github.com/compulim/BotFramework-WebChat/tree/sample-token-server/samples/token-server)", we have a section in README.md talks about what are the top security concerns if they want to make this code production ready


@omarsourour Does having the sample hosted with the other samples work with you?

omarsourour:
Hey Corina!

Yes it does. Please make sure that this being for Angular 5 (and probably down to Angular 2) is highlighted. Angular 6 uses RxJS 6, which breaks with this WebChat sample unless other considerations are taken into place.

If you require a sample for Angular 6, I can also work on providing that 😊

Thanks!
Omar Sourour

corinagum:
Thanks Omar! The goal is to have this sample available for 4.3. 
@compulim What do you think about asking for an Angular 6 sample? :)

Dongata:
Any updates?

corinagum:
Hey @omarsourour, just wondering if you have updates on this sample. :)

corinagum:
@omarsourour @compulim 
After offline discussion today, we agreed that Web Chat will need a patch for npm mime because Angular CLI needs a JSON loader to resolve the package names. Patch should be innocuous. Tracking is available here: #1596

[...]

PR #1686, opened 2019-02-01
Applying mime wrapper fix to avoid Angular CLI project build breaks

omarsourour:
Supports #1423 

### Motivation
We want the web-chat widget to be usable from inside an Angular CLI project.

### Issue
- node-mime imports two JSON files via `require` statements without specifying the `.json` extension in their path.
- Angular CLI web-pack configurations do not support loading extension-less JSON modules/files paths.
- The only way users have to support `.json` file loading is to *eject* the web-pack configurations, which is less then optimal for users who expect to come and use the Angular CLI and have it handle all configurations for them.

### Solution
- Use a custom node-mime wrapper that requires the JSON files while adding their extensions in the paths.

### DIFF
--- a/packages/core/src/sagas/sendFilesToPostActivitySaga.js
+++ b/packages/core/src/sagas/sendFilesToPostActivitySaga.js
@@ -4,6 +4,7 @@ import {
 } from 'redux-saga/effects';
 
 import mime from '../utils/mime-wrapper';
+
 import whileConnected from './effects/whileConnected';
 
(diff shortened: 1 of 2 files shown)


corinagum:
@omarsourour we're having some problems with our testing framework, so we might have to wait until that is resolved before getting your PR (and the other PR's) in :)

compulim:
Added credit/license to the original work. Will merge once it passed Travis CI.

PR #1691, opened 2019-02-04
exclude chat send activity from initial rsync

a-b-r-o-w-n:


a-b-r-o-w-n:
merging before travis reports because this just touches build.

Issue #1267, opened 2018-10-31
Bug: put width/height 100% styles in Web Chat


compulim:
![image](https://user-images.githubusercontent.com/1622400/47759744-3647f980-dc6e-11e8-9f99-1aced95ed075.png)

Can we put width/height in Web Chat to reduce boilerplate code?

compulim:
This boilerplate code is different for every website. In this sample, we are using full-screen for Web Chat, thus, it's 100%/100%. Developers may do it differently and using different values.

Will build a sample around this to show how to resize Web Chat dynamically.

corinagum:
also: need to update samples

corinagum:
I've seen at least 3 or 4 issues related to this both on GitHub and Stack Overflow. I feel like we should bump this to a P1. Thoughts?

Dpetters:
+1 on this, as I was about to file an issue for the same. When I initially used the component I thought I was doing something wrong because it wasn't taking up the whole div as it does in all of the samples. It took me a little while to realize that I was missing these styles. Having them as defaults and letting users override them would probably be the way to go.

tonyanziano:
Addressed in #1702

PR #1702, opened 2019-02-06
Added customizable height and width style options to root.

tonyanziano:
Fix for #1267 

===

This adds customizable `rootHeight` & `rootWidth` styleset options to the root style set.

**Images of working sample: (background element is pink)**

**50% height / width**

image

**100% height / 75% width**

image

image


tonyanziano:
I'll add a changelog entry and rebase upon approval.

PR #1704, opened 2019-02-06
Ponyfilling window.open

### LINKED ISSUE #368: Allow ponyfilling window.open

compulim:
> Fix #368.

## Background

To support Bot Framework Emulator and other hosting scenarios, all open URL actions should be customizable in JavaScript to allow different handling options.

## Design decisions

Some design decisions have been made while preparing this work.

### Function signature

`window.open` has [a signature](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) of `(url: string, windowName: string, windowFeatures: string) => void`.

Instead of adopting a similar signature of `(url: string) => void`, we prefer `({ cardAction: any }) => (url: string) => void`.

This allow us to send in `cardAction` and additional data, and not messing up with current signature of `window.open`.

### Intercepting `<a>`

Currently, we do not intercept `<a>` because it should be up to the hosting app on how to intercept them. And we believe the host should have ability to intercept them.

Intercepting `<a>` is a challenge because interception might break middle-click on them (normally it should open the link in a new window.)

### OAuth support

Today, OAuth flow do the following:

- Call `window.open()` with nothing, remember the result as `popup` variable
- Call ABS `getSessionId()` for `code_challenge`
- Update `popup.location.href` with a URL with `code_challenge` embedded in

The 1st and 3rd point is required to be separated to overcome popup blocker restrictions. We are advised not to pre-fetch `code_challenge` until the user click on it.

The function signature design allows us to do the job as of today, without intervention from popup blocker.

### Added
- `component`: Allow ponyfilling `window.open`, by [@compulim](https://github.com/compulim), in PR [#1704](https://github.com/Microsoft/BotFramework-WebChat/pull/1704)

### Samples
- `component`: [Customizing open URL behavior](https://microsoft.github.io/BotFramework-WebChat/18.customization-open-url), in PR [#1704](https://github.com/Microsoft/BotFramework-WebChat/pull/1704)


samples/18.customization-open-url/README.md
```diff
@@ -0,0 +1,111 @@
...
++       return url => popup.location.href = url;
++
++     default:
++       return url => {
++         if (confirm(`Do you want to open this URL?\n\n${ url }`)) {
++           window.open(url, '_blank');
++         }
++       };
++   }
+  },
```
a-b-r-o-w-n: This doesn't really feel like a ponyfill, its more of a custom url click handler

compulim:
If we name it "urlHandler" or similar, the user will have an impression that it could handle all `<a>`, which we don't. And in some cases, we can't control `<a>` too. Naming it "urlHandler" would give them a false impression that we will handle all URL in Web Chat, in fact, we don't and we can't.

But if we name it as `window.open` ponyfill, they know this is only for changing the behavior of programmatic access to a new window (a.k.a. popup). Not `<a>`. I want to be clear on this point.

Web Chat only use the first argument of `window.open`. And we don't use `windowName` and `windowFeatures` (2nd and 3rd argument). The user pass in the ponyfill, we only use certain features of the ponyfill, and we are using the ponyfill without changing its shape/signature. I think it didn't conflict with the idea of ponyfill.

compulim:
We can name it "cardActionOpenUrlHandler". Or we make the whole `onCardAction` replaceable using middleware pattern.

samples/18.customization-open-url/README.md
```diff
@@ -0,0 +1,111 @@
...
+…
+```
+
+To prevent getting blocked by popup blocker, the `window.open` call must be initiated from an user action.
```
corinagum:
```suggestion
To prevent getting blocked by a popup blocker, the `window.open` call must be initiated from a user action.
```

samples/18.customization-open-url/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+> Currently, when you click on "Sign in" of an OAuth card, it will fetch a `code_challenge` from our server before redirecting to the OAuth provider. To not getting blocked by popup blocker, we need to call `window.open` first, then fetch the `code_challenge`, and lastly redirect the user to OAuth provider.
```
corinagum:
```suggestion
> Currently, when you click on "Sign in" of an OAuth card, it will fetch a `code_challenge` from our server before redirecting to the OAuth provider. To not getting blocked by a popup blocker, we need to call `window.open` first, then fetch the `code_challenge`, and lastly redirect the user to the OAuth provider.
```

samples/18.customization-open-url/README.md
```diff
@@ -0,0 +1,111 @@
...
+  </body>
+</html>
+
+```
+
+# Further reading
+
+## Full list of Web Chat hosted samples
+
+View the list of available samples by clicking [here](https://github.com/Microsoft/BotFramework-WebChat/tree/master/samples)
```
corinagum:
```suggestion
View the list of available samples by clicking [here](https://github.com/Microsoft/BotFramework-WebChat/tree/master/samples)
```

samples/18.customization-open-url/index.html
```diff
@@ -0,0 +1,59 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <title>Web Chat: Customize open URL behavior</title>
+    <!--
+      For demonstration purposes, we are using the development branch of Web Chat at "/master/webchat.js".
+      When you are using Web Chat for production, you should use the latest stable release at "/latest/webchat.js",
+      or lock down on a specific version with the following format: "/4.1.0/webchat.js".
+    -->
+    <!-- <script src="https://cdn.botframework.com/botframework-webchat/master/webchat.js"></script> -->
```
corinagum:
```suggestion
    <script src="https://cdn.botframework.com/botframework-webchat/master/webchat.js"></script>
```

samples/18.customization-open-url/index.html
```diff
@@ -0,0 +1,59 @@
...
+    <script src="/webchat.js"></script>
```
corinagum: delete line

packages/component/src/Composer.js
```diff
@@ -96,20 +104,18 @@ function createCardActionLogic({ directLine, dispatch }) {
...
-          const popup = window.open();
+          const openUrl = openUrlPonyfillFactory({ cardAction });
 
           if (directLine.getSessionId)  {
             const subscription = directLine.getSessionId().subscribe(sessionId => {
-              popup.location.href = `${ value }${ encodeURIComponent(`&code_challenge=${ sessionId }`) }`;
+              openUrl(`${ value }${ encodeURIComponent(`&code_challenge=${ sessionId }`) }`);
 
               // HACK: Sometimes, the call complete asynchronously and we cannot unsubscribe
               //       Need to wait some short time here to make sure the subscription variable has setup
```
corinagum:
```suggestion
              //       Need to wait a short amount of time here to make sure the subscription variable has setup
```

samples/18.customization-open-url/README.md
corinagum:
```suggestion
View the list of [available Web Chat samples](https://github.com/Microsoft/BotFramework-WebChat/tree/master/samples)
```
```

compulim: Didn't see a change here.

samples/18.customization-open-url/index.html
compulim: Fixed.

samples/18.customization-open-url/index.html
compulim: Fixed.

cwhitten:
I agree that the way this is written, full control is yielded to the caller and we would be presumptive to assume that their implementation would always fall back to a closure that invokes `window.open`. we should give it a name that better describes that. how about `scopeCardActionHandler`?

compulim:
@cwhitten If this is named `scopeCardActionHandler`, it will need to understand how ["signin" card action lead to a popup](https://github.com/Microsoft/BotFramework-WebChat/blob/master/packages/component/src/Composer.js#L104). I.e. it must construct a URL by calling Direct Line service (`DLJS.getSessionId`), get back the result, and construct a URL, then open it. I think the implementer shouldn't need to know about these mechanics.

cwhitten:
@compulim nice work! let's rebase this and get it merged.

Bmastaz:
There seems to be no explanation why this was removed in https://github.com/microsoft/BotFramework-WebChat/pull/1888/

PR #1706, opened 2019-02-07
Hide suggested actions not destinated to the user

### LINKED ISSUE #1057: SuggestedActions that are accidentally sent to the bot are being rendered for the user
The v4 Emulator which uses [0.13.1-master.6da7b6c](https://github.com/Microsoft/BotFramework-Emulator/blob/v4/packages/app/client/package.json#L91) will show the user SuggestedActions that were sent to the bot. Testing this using the iFrame also renders the same result, SuggestedActions that should be only "seen" by the bot are showing for the user.

On other channels like Skype, if the suggestedAction `to` field is populated with the bot's id, the SuggestedActions do not render for the user. 

Example code from [**BB-Py #100**](https://github.com/Microsoft/botbuilder-python/issues/100):

```py
activity = Activity(
    type=ActivityTypes.message,
    channel_id=activity.channel_id,
    conversation=activity.conversation,
    from_property=activity.recipient,
    recipient=activity.from_property,
    service_url=activity.service_url
)

replies = [
    {
        "title": "Test1",
        "type": ActionTypes.im_back,
        "value": "Test1",
    },
    {
        "title": "Test2",
        "type": ActionTypes.im_back,
        "value": "Test2",
    }
]

activity.text = "My text"
# The following line is where the problem code in the bot exists: `to` uses
# `activity.from_property.id` instead of `activity.recipient.id`
activity.suggested_actions = SuggestedActions(actions=replies, to=[activity.from_property.id])
```

The activity's SuggestedActions show up for the user in WebChat and in the v4 Emulator, whereas this code sent to another channel would not work. Sending this activity to Skype results in the text, `"My text"` appearing, but not the SuggestedActions.

compulim:
> Fix #1057 

## Background

Suggested actions ([spec here](https://docs.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-api-reference?view=azure-bot-service-4.0#suggestedactions-object)) can specify `to` field to show the suggested actions only for a set of users.

Today, we did not filter suggested actions based on the `to` field.

This fix will not show suggested actions with a `to` field targeting another user. For backward compatibility reason, if the `to` field is empty or not defined, we will continue to show suggested actions.

### Fixed

- Fix [#1057](https://github.com/Microsoft/BotFramework-WebChat/issues/1057). Fixed suggested actions destinated for other recipients should not show up, by [@compulim](https://github.com/compulim) in PR [#1706](https://github.com/Microsoft/BotFramework-WebChat/pull/1706)


CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `bundle`: Fix [#1652](https://github.com/Microsoft/BotFramework-WebChat/issues/1652). Pass `pollingInterval` to DirectLineJS constructor, by [@neetu-das](https://github.com/neetu-das) in PR [#1655](https://github.com/Microsoft/BotFramework-WebChat/pull/1655)
 - `core`: Reworked logic on connect/disconnect for reliability on handling corner cases, by [@compulim](https://github.com/compulim) in PR [#1649](https://github.com/Microsoft/BotFramework-WebChat/pull/1649)
 - `core`: Fix [#1521](https://github.com/Microsoft/BotFramework-WebChat/issues/1521). Add connectivity status component and update localization, by [@corinagum](https://github.com/corinagum) in PR [#1679](https://github.com/Microsoft/BotFramework-WebChat/pull/1679)
+- Fix [#1057](https://github.com/Microsoft/BotFramework-WebChat/issues/1057). Fixed suggested actions destinated for other recipients should not show up, by [@compulim](https://github.com/compulim) in PR [#1706](https://github.com/Microsoft/BotFramework-WebChat/pull/1706)
```
a-b-r-o-w-n: should this be tagged with `core`?

corinagum:
```suggestion
- Fix [#1057](https://github.com/Microsoft/BotFramework-WebChat/issues/1057). Fixed suggested actions destined for other recipients should not show up, by [@compulim](https://github.com/compulim) in PR [#1706](https://github.com/Microsoft/BotFramework-WebChat/pull/1706)
```

__tests__/suggestedActions.js
```diff
@@ -148,4 +148,22 @@ describe('suggested-actions command', async () => {
 
     expect(base64PNG).toMatchImageSnapshot(imageSnapshotOptions);
   }, 60000);
+
+  test('should not show suggested actions not destinated for the user', async () => {
```
corinagum:
```suggestion
  test('should not show suggested actions not destined for the user', async () => {
```

tonyanziano: the test was a lot more simple than i thought! 👍

compulim: Thanks. Downside, if you want to try out corner cases, you will need to make the integration part (e.g. DLJS) fall at the angle you want. E.g. Corina's offline UI work, she actually need to make a faulty DLJS object to achieve that.

CHANGELOG.md
compulim: Fixed, tagged with `core`.

packages/core/src/sagas/incomingActivitySaga.js
```diff
@@ -26,13 +26,18 @@ function* observeActivity({ directLine, userID }) {
...
+    // TODO: [P3] We could put this logic inside reducer to minimize number of actions dispatched
     const messageActivities = yield select(activitiesOfType('message'));
     const lastMessageActivity = messageActivities[messageActivities.length - 1];
 
     if (activityFromBot(lastMessageActivity)) {
-      const { suggestedActions: { actions } = {} } = lastMessageActivity;
+      const { suggestedActions: { actions, to } = {} } = lastMessageActivity;
 
-      yield put(setSuggestedActions(actions));
+      if (!to || !to.length || to.includes(userID)) {
```
justinwilaby: This can optionally be simplified:
```js
const params = (to || []).includes(userId) ? actions : null;
yield put(setSuggestedActions(params));
```

compulim: I was thinking to simplify this too (in a single line). I thought people would hate me. Good to know you don't. 😄

The correct one should be `!to || !to.length || to.includes(userID) ? actions : null`. I will add a comment above it to explain.

I think it is better to write comment and have shorter-but-cryptic code, than no comment but longer code.

Will fix it.

PR #1709, opened 2019-02-08
Fix SendStatus and add tests

corinagum:
'>.< 

Here is the new Send Failed dialog working:
image

I broke send failed, so I fixed it and added more offline UI tests.


compulim:
Changelog, otherwise, looks good.

PR #1711, opened 2019-02-09
Improve test reliability

compulim:
- Added retry logic to test setup
   - Fail after 3 (consecutive) failures
- Moved "ping bot" test to test setup, so we can retry it
   - Set `pingBotOnLoad` to `false` if bot test should not be performed, for example, offline UI tests
- Added page objects
   - `dispatchAction` to dispatch an action
   - `pingBot` to send a ping to the bot and anticipate for pong
   - `sendMessageViaSendBox` will send message and wait until it is sent
- Added `timeouts.test` to `60000`

### DIFF
--- a/__tests__/basic.js
+++ b/__tests__/basic.js
@@ -9,7 +9,6 @@ import minNumActivitiesShown from './setup/conditions/minNumActivitiesShown';
 test('setup', async () => {
   const { driver, pageObjects } = await setupWebDriver();
 
-  await pageObjects.pingBot();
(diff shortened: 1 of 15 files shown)


cwhitten:
do we know what is causing the change in test behavior? why has it started to be so flakey?

compulim:
@cwhitten I believe more flakey is because we added more tests to it. My preliminary investigations suggest could be platform issue, will work with the platform team and see.

PR #1745, opened 2019-02-20
fixed locale pt-br not being used and added extra suport for pt-br

pedropacheco92:


joaopedsa:
Nice Work!!

packages/component/src/Localization/Localize.js
```diff
@@ -110,7 +110,7 @@ function normalizeLanguage(language) {
   } else if (language.startsWith('pl')) {
     return 'pl-PL';
   } else if (language.startsWith('pt')) {
-    if (language === 'pt-BR') {
+    if (language === 'pt-br') {
```
corinagum: could you explain why this is necessary?

packages/component/src/Localization/pt-BR.js
```diff
@@ -1,3 +1,44 @@
...
+      hour12: false,
+      minute: '2-digit',
+      month: '2-digit',
+      year: 'numeric',
+    });
+  }
+}
+
 export default {
   // FAILED_CONNECTION_NOTIFICATION: '',
```
corinagum: Are you able to fill the commented-out lines to finish up this localization file? :)

packages/component/src/Localization/Localize.js
pedropacheco92: in the first line of this function, the language is setted to lower case, and this condidion was never true.

packages/component/src/Localization/pt-BR.js
pedropacheco92: Sure!

packages/component/src/Localization/Localize.js
corinagum: Ah I understand now, thank you!

pedropacheco92:
Fixed conflict

corinagum:
Will merge in once tests pass. Thanks again!

corinagum:
These changes will be in `/latest/` of the 4.4 release. :)

PR #1832, opened 2019-03-21
Add word-wrap to messageActivity

### LINKED ISSUE #1774: Words or links that are longer than the text box is wide don't wrap in message boxes
Repro steps: go to the demo page https://microsoft.github.io/BotFramework-WebChat/01.a.getting-started-full-bundle/ and for the user message type "https://subdomain.domain.com/pathname0/pathname1/pathname2/pathname3/pathname4/". The last part of the URL doesn't show up in the user or bot message boxes.

This could be fixed by adding "word-break: break-all;" as a style to .row.message elements

corinagum:
Fix #1774 

Previously, long urls or long (single) words would be clipped. Now, Web Chat will by default break on word. This does mean **long URLs break arbitrarily**. 

image
Note that semi-long words will wrap normally. See 'monospace' below:
image

Web Chat users may use [defaultStyleSetOptions](https://github.com/Microsoft/BotFramework-WebChat/blob/master/packages/component/src/Styles/defaultStyleSetOptions.js) to change this default value. Available options are `normal`, `break-all`, `break-word`, and `keep-all`.


__tests__/basic.js
```diff
@@ -20,3 +20,16 @@ test('setup', async () => {
 
   expect(base64PNG).toMatchImageSnapshot(imageSnapshotOptions);
 });
+
+test('long urls', async () => {
```
compulim:
```suggestion
test('long URLs', async () => {
```

I believe the snapshot would work (Jest should lowercase the name).

PR #1880, opened 2019-04-08
Add reconnection logic

compulim:
## Background

Today, `DirectLineJS` does not support reconnection. But since we are working on "offline UI" feature, we need this feature. We assume in `DirectLineJS`, if `connectionStatus$` change from `2` (online) to `1` (connecting), it means it is reconnecting.

This work will unblock @corinagum work on offline UI.

This will also fixes some bugs:

- Any `DISCONNECT` actions is dispatched while we are dispatching `CONNECT_FULFILLING` and `CONNECT_FULFILLED`, we might miss them

## To-do

`Things still need to work on and needed for the completion of this PR:

- ~While reconnecting, we did not monitor `DISCONNECT` action;~
- ~Verify if the following bug is in the original code~
   - ~Any `DISCONNECT` actions is dispatched in `CONNECT_FULFILLED`, we might miss them.~

Looks like we did not handle `DISCONNECT` action before `CONNECT_FULFILLED`, so I am postponing the work to handle `DISCONNECT` action during `RECONNECT_*`, since it is "as-of" today. Crossing out to-do items.

(Crossed out lines are as-of today, 

### Changed
- Added handling of reconnection, by [@compulim](https://github.com/compulim), in PR [#1880](https://github.com/Microsoft/BotFramework-WebChat/pull/1880)


packages/core/src/sagas.js
```diff
@@ -22,6 +22,8 @@ import stopSpeakingActivityOnInputSaga from './sagas/stopSpeakingActivityOnInput
 import submitSendBoxSaga from './sagas/submitSendBoxSaga';
 
 export default function* () {
+  // TODO: [P2] Since fork() silently catch all exceptions, we need to find a way to console.error them out.
```
corinagum:
```suggestion
  // TODO: [P2] Since fork() silently catches all exceptions, we need to find a way to console.error them out.
```

packages/core/src/sagas/connectSaga.js
```diff
@@ -122,74 +133,129 @@ function* connectSaga( directLine ) {
...
       // If there are any errors from this point, we need to make sure we call endDirectLine() to release resources.
       try {
-        yield put( { type: CONNECT_FULFILLING, meta, payload: { directLine } } );
-        yield put( { type: CONNECT_FULFILLED, meta, payload: { directLine } } );
-
-        const terminateAction = yield take( [ DISCONNECT, negativeUpdateConnectionStatusAction ] );
+        let gotDisconnectAction;
+
+        for (let numConnect = 0;; numConnect++) {
+          // This code section use all() because we want to make sure CONNECT_FULFILLED is dispatched
```
corinagum:
```suggestion
          // This code section uses all() because we want to make sure CONNECT_FULFILLED is dispatched
```

packages/core/src/sagas/connectSaga.js
```diff
@@ -122,74 +133,129 @@ function* connectSaga( directLine ) {
...
+          // This code section uses all() because we want to ensure CONNECT_FULFILLED is dispatched
+          // before we dispatch any connection-related actions, including DISCONNECT_FULFILLING and RECONNECT_PENDING.
+          const {
+            disconnection: {
+              takeDisconnectAction,
+              takeUpdateConnectionStatusAction
+            }
+          } = yield all(
+            {
+              _: numConnect ?
```
cwhitten: it seems expected, but `numConnect` being 0 the first iteration is the only time this evaluates to false. I'm wondering if it get be written in a way to make it more clear. Maybe a comment denoting such and why would be helpful.

compulim: Updated the code (in a big way). Could you please review it again?

compulim:
Will add more unit tests before merging.

packages/core/src/sagas/connectSaga.js
```diff
@@ -97,99 +84,175 @@ function rectifyUserID( directLine, userIDFromAction ) {
...
+    baseAction,
+    function* () {
+      const { result } = race({
+        _: takeDisconnectAsError(),
+        result: callEffectFactory()
+      });
+
+      return result;
+    }
+}
```
cwhitten: this is missing an enclosing bracket

compulim: Fixed. Sorry for committing bad code.

Now got unit tests to make sure everything work as expected.

cwhitten:
Now a snapshot appears to be failing:

 FAIL  __tests__/offlineUI.js (53.956s)
  ● offline UI › should show "unable to connect" UI when credentials are incorrect
    Expected image to match or be a close match to snapshot but was 0.1896701388888889% different from snapshot (437 differing pixels).
    See diff for details: /home/travis/build/Microsoft/BotFramework-WebChat/__tests__/__image_snapshots__/chrome-docker/__diff_output__/offline-ui-js-offline-ui-should-show-unable-to-connect-ui-when-credentials-are-incorrect-1-diff.png

compulim:
> Now a snapshot appears to be failing:
> 
> FAIL **tests**/offlineUI.js (53.956s)
> ● offline UI › should show "unable to connect" UI when credentials are incorrect
> Expected image to match or be a close match to snapshot but was 0.1896701388888889% different from snapshot (437 differing pixels).
> See diff for details: /home/travis/build/Microsoft/BotFramework-WebChat/**tests**/**image_snapshots**/chrome-docker/**diff_output**/offline-ui-js-offline-ui-should-show-unable-to-connect-ui-when-credentials-are-incorrect-1-diff.png

This is a legit test fail. Fixed it in a commit, waiting for test to run. Good to have tests cover us.

Originally, we thought `DISCONNECT_FULFILLED` only means graceful disconnect, thus we hide the error UI if we see `DISCONNECT_FULFILLED`.

But with the reconnect work, I believe we should have `CONNECT` and `DISCONNECT` working in pair: if we ever dispatched `CONNECT_*`, we should also dispatch `DISCONNECT_*` eventually (peacefully or erroneous).

With paired actions, it will simplify UI work. And that's why I like coding logics in saga to interface with external systems (DLJS). We can turn not-strictly-deterministic eventing system, into an ordered and deterministic eventing system. 😄

PR #1884, opened 2019-04-09
Fix CSS collisions with Bootstrap and bump memoize-one

### LINKED ISSUE #1643: Upgrade memoize-one to v5.x
When installing `botframework-webchat@4.2.0`, there are a number of warnings saying that `memoize-one@4.1.0` should be upgraded:
```
warning botframework-webchat > memoize-one@4.1.0: New custom equality api does not play well with all equality helpers. Please use v5.x
```

### LINKED ISSUE #1837: Need style option to change margins around bubbles
I'm facing this issue:

image

On the left is without any styling changes, on the right is my fixed version using this CSS rule set:
```
#webchat .content .row {
    margin-left: 0;
    margin-right: 0;
}
```

It would be great if this could be added to the style options

corinagum:
Fixes #1837 
Fixes #1643 

- End collision of external styles like Bootstrap by changing the `.row` class in Web Chat to `.webchat__row`
- Bump `memoize-one`


compulim:
I am thinking to [convert it to BEM naming convention](http://getbem.com/introduction/), i.e. `webchat__row`.

But if @cwhitten is okay with the current one, `webchat-row`, then let's hit this first and then go to BEM later.

PR #1895, opened 2019-04-12
Add Reconnect UI to ConnectivityStatus when connection has been interrupted

### LINKED ISSUE #1523: Need offline UI: Connection interrupted
Original issue is #135, which has been split into multiple issues. 

We need UI specs and implementation for:
- When the bot connection has been interrupted.
  - Immediate interruption (including immediate reconnect?)
  - Non-immediate interruption (up to 2 mins)

### What we need
- [x] Design specs
- [x] Acceptance criteria
- [x] T-shirt sizing

-	Grey text
-	With spinner (spinner will be grey, small, size of icon in previous Offline UI items)
-	Text proposal: “Network interruption occurred. Reconnecting…”
-	This will show up as soon as an interruption has occurred, and either disappear (reconnect successful) or switch to “Web Chat is taking longer than usual to connect” after 15 seconds. (Will eventually turn to “Connection failed” if it fails completely)

### LINKED ISSUE #1831: Add Typing Indicator background URL to defaultStyleOptions
I'd like to know if it is possible to change the typing indicator? If so can someone provide a code snippet or reference?

corinagum:
Resolves #1523 
Resolves #1831

Annotation 2019-04-11 165420

- When Web Chat connects but then loses connection, the above component will show, indicating to user that interruption has occurred.
- After 15 seconds, spinner animation will revert to "Connection taking longer than usual" warning icon.
- Similar to connecting for the first time, the Network interruption animation has a short timeout to ensure it is legible for at least 0.4 of a second.
- Web Chat users may now customize the typing indicator and spinner animation through `defaultStyleSetOptions.js`


__tests__/offlineUI.js
```diff
@@ -214,4 +214,81 @@ describe('offline UI', async () => {
...
+        const scriptElement = document.createElement('script');
+        scriptElement.onload = resolve;
+        scriptElement.setAttribute('src', 'https://unpkg.com/core-js@2.6.3/client/core.min.js');
+
+        document.head.appendChild(scriptElement);
+      })
+    });
+
+    await driver.sleep(600);
+    const base64PNG = await driver.takeScreenshot();
```
tonyanziano: Is taking a screenshot necessary if they are never compared?

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -18,19 +18,21 @@ class connectConnectivityAlert extends React.Component {
     }
 
     if ( status === 'uninitialized' || ( status === 'connected' && timeoutCompleted === false && connectingSlow === false ) ) {
-      this.timerId = setTimeout( () => {
-        if ( status === 'connected' ) {
-          this.props.dispatch( { type: 'DIRECT_LINE/CONNECT_TIMEOUT_COMPLETE' } );
-        }
+      if ( status === 'connected' ) {
+        this.timerId = this.dispatchConnectivityTimeout();
```
tonyanziano: won't `this.timerId` be `undefined` in this case because `this.dispatchConnectivityTimeout()` does not return `setTimeout()` it just invokes it

tonyanziano: I also am having trouble understanding why `this.timerId` is set and then immediately after is set to `null`

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -18,19 +18,21 @@ class connectConnectivityAlert extends React.Component {
...
         <React.Fragment>
           <SpinnerAnimation />
           {localize( 'INITIAL_CONNECTION_NOTIFICATION', language )}
         </React.Fragment>
       );
-    } else if ( status === 'reconnecting' ) {
+    } else if ( status === 'reconnecting' || (status === 'reconnected' && timeoutCompleted === false && connectingSlow === false)) {
+      if ( status === 'reconnected' ) {
+        this.timerId = this.dispatchConnectivityTimeout();
+        this.timerId = null;
```
tonyanziano: same concerns here

__tests__/offlineUI.js
corinagum: The sleep is necessary so that if the screenshot is taken, it doesn't take the screenshot too early.

I left these lines here so that if we do need to compare screenshots, only the compare screenshots line needs to be added. But I see what you mean. What do you think of the following fix? 

```
// Snapshots are intentionally not compared because the spinner will cause the snapshot to fail regularly
// Uncomment the following lines to re-capture snapshot or do a manual comparison

// await driver.sleep(600);
// const base64PNG = await driver.takeScreenshot();
// expect(base64PNG).toMatchImageSnapshot(imageSnapshotOptions);
```

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -54,6 +56,12 @@ class connectConnectivityAlert extends React.Component {
     }
   }
 
+  dispatchConnectivityTimeout() {
+    setTimeout( () => {
```
tonyanziano: this should return the call to `setTimeout()` so that the caller of `dispatchConnectivityTimeout()` can receive the timer id.

Ex.

```
dispatchConnectivityTimeout() {
  return setTimeout(() => 
    ...
```

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -18,19 +18,21 @@ class connectConnectivityAlert extends React.Component {
     }
 
     if ( status === 'uninitialized' || ( status === 'connected' && timeoutCompleted === false && connectingSlow === false ) ) {
-      this.timerId = setTimeout( () => {
-        if ( status === 'connected' ) {
-          this.props.dispatch( { type: 'DIRECT_LINE/CONNECT_TIMEOUT_COMPLETE' } );
-        }
+      if ( status === 'connected' ) {
+        this.timerId = this.dispatchConnectivityTimeout();
         this.timerId = null;
```
tonyanziano: setTimeout is async so this line will immediately be executed and overwrite the timer id you just stored in `this.timerId`. Consider setting `this.timerId` to null inside of the `setTimeout()` callback

packages/core/src/reducers/connectivityStatus.js
```diff
@@ -26,6 +31,15 @@ export default function (state = DEFAULT_STATE, { type, meta }) {
 
       break;
 
+    case RECONNECT_PENDING:
+      state = {
+        ...state,
+        status: 'reconnecting',
+        timeoutCompleted: false
+      }
```
tonyanziano: semicolon

packages/core/src/reducers/connectivityStatus.js
```diff
@@ -35,6 +49,15 @@ export default function (state = DEFAULT_STATE, { type, meta }) {
 
       break;
 
+    case RECONNECT_FULFILLED:
+      state = {
+        ...state,
+        status: 'reconnected',
+        timeoutCompleted: false
+      }
```
tonyanziano: semicolon

compulim:
In ConnectivityStatus.js, few things need to be modified:
- Don't put JSX outside of render() function, it makes the output very hard to read
- Don't do side-effect during render()
   - In your `render()`, you call `connectivityStatusText()`
   - Then, in your `connectivityStatusText()`, you `setTimeout()`
   - Few reasons:
      - You *never* know when `render()` will happen
      - Render is for rendering HTML, not for business logic, need a super clean cut here
- Line 85, this if statement (`props !== state`) always truthy
   - This is shallow comparison
   - Only change what you need
   - Why put `styleSet` in `state`?
- Don't call `setState()` in `componentDidUpdate()`
   - `componentDidUpdate()` is being called *after* the screen is *rendered*
   - Then you call `setState()`, that means you are asking the screen to render *again*

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -1,110 +1,102 @@
...
+            || connectivityStatus === 'reconnecting'
+        })
+      }
+      role="status"
+    >
+      <ConnectivityDebounce
+        debounce={ (connectivityStatus === 'uninitialized' || connectivityStatus === 'error') ? 0 : 400 }>
+        { () =>
+          connectivityStatus === 'connectingslow' ?
+            <React.Fragment>
```
cwhitten: Are you able to render a single React.Fragment outside the chained ternary like this?
```javascript
() => {
  <React.Fragment>
    {
     connectivityStatus === 'connectingslow' ?
       <WarningNotificationIcon />
     : (connectivityStatus === 'error' || connectivityStatus === 'notconnected') ?
       <ErrorNotificationIcon />
     : connectivityStatus === 'uninitialized' ?
       <SpinnerAnimation />
       { localize('INITIAL_CONNECTION_NOTIFICATION', language) }
    }
 </React.Fragment>
}
```

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -1,110 +1,102 @@
...
+      renderConnectivity: props.children,
+      since: Date.now()
+    };
   }
 
-  componentDidUpdate( props ) {
-    const { connectivityStatus: { status } } = props;
-    const { connectingSlow } = this.state
+  componentWillReceiveProps(nextProps) {
+    const { children, debounce } = nextProps;
```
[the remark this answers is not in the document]
corinagum: `debounce` is the amount of time specified to wait between rerenders. If there is a change in `connectivityStatus`, but the time between the previous render and now is less than (ex.) 400 milliseconds, `setTimeout` will be called with the difference of how many milliseconds are left until 400 is reached. If the number is greater than 400, the timeout is set to 0 and state is set immediately. 

Please let me know if there's a better name for this and the component. :)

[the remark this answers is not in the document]
cwhitten: I got it. Conventionally `debounce` is known as a function, where the interval is a parameter. Maybe a more explicit name like `debounceInterval` would avoid confusion of it being a function or not.

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -1,110 +1,102 @@
...
-          <ErrorNotificationIcon />
-          {localize( 'FAILED_CONNECTION_NOTIFICATION', language )}
-        </React.Fragment>
-      );
-    }
-  }
 
-  constructor( props ) {
-    super( props );
+class ConnectivityDebounce extends React.Component {
```
cwhitten: Looking at this again, `DebouncedConnectivityStatus` seems like a better name

packages/component/src/SendBox/ConnectivityStatus.js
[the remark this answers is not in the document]
cwhitten: If you change the class name to what I suggested, having this value just be `interval` would be good enough, I think.

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -1,110 +1,102 @@
...
-          [ styleSet.connectivityNotification ]: status === 'connected' || status === 'uninitialized' || status === 'reconnecting'
-        } )}
-        role="status"
-      >
-        {this.connectivityStatusText()}
-      </div>
-    );
+    const { renderConnectivity } = this.state;
+
+    return typeof renderConnectivity === 'function' ? renderConnectivity() : false;
```
cwhitten: in a "Function as Children" pattern like what you have here, I typically just do something like below to self-document the intended control flow clearly without any naming sugar.

```javascript
  return typeof this.props.children === 'function' ? this.props.children() : false
```

__tests__/offlineUI.js
```diff
@@ -209,9 +209,93 @@ describe('offline UI', async () => {
 
       })
     });
+    // Snapshots are intentionally not compared because the spinner will cause the snapshot to fail regularly
+    // Uncomment the following lines to re-capture snapshot or do a manual comparison
 
-    const base64PNG = await driver.takeScreenshot();
+    // const base64PNG = await driver.takeScreenshot();
```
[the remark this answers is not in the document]
corinagum: Because of the gif in ConnectivityStatus, these screenshots will regularly fail. A previous suggestion was to take the screenshot and not do a comparison, and because there's been a lot of feedback I've been playing with what to comment out, if anything, so manual tests can still be done.

I got a new suggestion yesterday where I will temporarily replace the gif with a png during the tests so they won't fail regularly. This also allows customers to insert their own spinner if they want to via `defaultStyleSetOptions`. I'll make that a part of this PR (and the commented code will be gone).

__tests__/offlineUI.js
```diff
@@ -209,9 +209,93 @@ describe('offline UI', async () => {
...
+        const scriptElement = document.createElement('script');
+
+        scriptElement.onload = resolve;
+        scriptElement.setAttribute('src', 'https://unpkg.com/core-js@2.6.3/client/core.min.js');
+
+        document.head.appendChild(scriptElement);
+      })
+    });
+
+    await driver.sleep(17000);
```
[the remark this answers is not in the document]
corinagum: This test is waiting for the 'slow to connect' timeout, which is the 17 seconds. I'm not sure if there's any way to circumvent the timeout during tests, but if there is please let me know and I can fix it.

packages/component/src/SendBox/ConnectivityStatus.js
corinagum: I don't think this will work because each status also has different text to be localized, and you can't return multiple children, only one. 

this gives errors:

```() => {
  <React.Fragment>
    {
     connectivityStatus === 'connectingslow' ?
       <WarningNotificationIcon />
       { localize('SLOW_CONNECTION_NOTIFICATION', language) }
     : (connectivityStatus === 'error' || connectivityStatus === 'notconnected') ?
       <ErrorNotificationIcon />
       { localize('FAILED_CONNECTION_NOTIFICATION', language) }
     : connectivityStatus === 'uninitialized' ?
       <SpinnerAnimation />
       { localize('INITIAL_CONNECTION_NOTIFICATION', language) }
    }
 </React.Fragment>
}```

Playing around with parenthesis didn't get rid of the errors. Let me know if I should try something else. :)

corinagum:
PR modifications have been pushed up, with an added fix:
https://github.com/Microsoft/BotFramework-WebChat/issues/1831 - Typing Indicator and Spinner now have the ability to be specified by the user in `defaultStyleSetOptions.js`.

corinagum:
@cwhitten could you take another pass when you have a chance? Thank you!

PR #1909, opened 2019-04-17
Added SuggestedAction Image Functionality

### LINKED ISSUE #1739: Suggested actions with image/icon
Any way or plans to add images to suggested actions buttons?

Something similar to Facebook Quick Replies

image

Currently I'm trying to set the Image property of the CardActions given to the SuggestedActions but without success.

tdurnford:
Fixes #1739 

Added support for suggested action images.

suggested-actions-js-suggested-actions-command-should-show-suggested-actions-with-images-1-snap


packages/component/src/Styles/StyleSet/SuggestedAction.js
```diff
@@ -18,12 +19,14 @@ export default function createSuggestedActionStyle({
     paddingTop: paddingRegular,
 
     '& > button': {
+      alignItems: 'center',
+      borderRadius: suggestedActionBorderRadius,
+      display: 'flex',
       fontFamily: primaryFont,
       fontSize: 'inherit',
-      height: suggestedActionHeight,
+      minHeight: suggestedActionHeight,
```
corinagum: switch back to `height`?

corinagum: also should we allow the user to affect the image size directly? Not just the suggested action?

corinagum: nvm I see it below haha

corinagum:
Could you add more tests for adjusting the dimensions of the image? That way we can be explicit with design team about our assumptions and see what they think. :)

PR #1922, opened 2019-04-23
Added disabled property to send button

### LINKED ISSUE #1920: SendButton should pass disabled attribute to IconButton
I am attempting to use the "sendBoxButtonColorOnFocus" styleset property, but it doesn't appear to get used because even if I set `disabled=true` on the `ReactWebChat` component, the SendButton is still enabled (whereas the TextBox becomes disabled). Thus the following styling does not apply:
```
    '&:disabled svg': {
      fill: sendBoxButtonColorOnDisabled
    }
```
I believe in SendButton.js, the `disabled` attribute should be set on the IconButton element in the way that it is done in MicrophoneButton and UploadButton.

tdurnford:
Fixes #1920 

Added the disabled property to the send button, so the 'sendBoxButtonColorOnDisabled' default style applies to the button when Web Chat is disabled. 

**Example Code**
```js
  render() {

    const styleOptions = { 
      sendBoxButtonColorOnDisabled: 'blue'
    }
    
    return (
      this.state.directLine
        ? <ReactWebChat
          className="chat"
          disabled={true}
          directLine={ this.state.directLine }
          styleOptions={ styleOptions }
          { ...this.props }
        />
        :<div>Connecting to bot&hellip;</div>
    );
  }
```

**Screenshot**
disabled-ui-js-tests-the-ui-of-disabled-web-chat-change-default-disabled-button-color-1-snap


__tests__/disabledUI.js
```diff
@@ -0,0 +1,31 @@
...
+import { imageSnapshotOptions, timeouts } from './constants.json';
+
+// selenium-webdriver API doc:
+// https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_WebDriver.html
+
+jest.setTimeout(timeouts.test);
+
+describe('tests the UI of disabled Web Chat', () => {
+  test('generic disabled web chat', async () => {
+    const styleOptions = { sendBoxTextWrap: true };
```
corinagum: is this needed?

tdurnford: Not really. Just trying to thorough.

compulim:
How about focus and hover?

Issue #1525, opened 2018-12-20
Need offline UI: On JavaScript Error


corinagum:
Original issue is #135, which has been split into multiple issues. 

We need UI specs and implementation for:
- When a JavaScript error is detected and Web Chat has shut itself down. Currently, you will only know this has happened by checking the console.

### What we need
- [ ] Design specs
- [ ] Acceptance criteria
- [ ] T-shirt sizing

PR #1927, opened 2019-04-25
#1525 Add "Render error" connectivity status for JavaScript errors.

corinagum:
Fix: #1525 

- Updated README.md with `createStore` in the Web Chat API Reference
- Updated HowToQuestions.md with link to web-chat tag on Stack Overflow
- Add UI for render error when there's a JavaScript error in Web Chat

rendererror


packages/core/src/reducers/connectivityStatus.js
```diff
@@ -43,6 +45,10 @@ export default function (state = DEFAULT_STATE, { type, meta }) {
       state = meta.error ? 'error' : 'notconnected'
       break;
 
+    case SAGA_ERROR:
+      state = 'sagaerror'
```
tonyanziano: semicolon

tonyanziano: also missing one on line 45

packages/core/src/createStore.ts
```diff
@@ -8,7 +8,16 @@ export type State = {}
 export type ChatStore = Store<State>
 
 export default function (initialState, ...middlewares) {
-  const sagaMiddleware = createSagaMiddleware();
+  const sagaMiddleware = createSagaMiddleware({
+    onError: (...args) => {
+      const [err] = args;
+
+      console.error(err);
```
tonyanziano: is this console intentional?

packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -89,11 +90,16 @@ export default connectConnectivityStatus(
               <SpinnerAnimation />
               { localize('INITIAL_CONNECTION_NOTIFICATION', language) }
             </React.Fragment>
-          : connectivityStatus === 'reconnecting' &&
+          : connectivityStatus === 'reconnecting' ?
```
tonyanziano: chaining tertiary operators together seems like a less straight-forward pattern to me than a switch statement, but that's a personal preference

packages/core/src/createStore.ts
corinagum: yep, intentional. :)

packages/component/src/SendBox/ConnectivityStatus.js
corinagum: yeah, I had it as switch statements previously but was asked to change it lol. For simplicity I'm going to keep it as is.

corinagum:
New test is failing. I am investigating - sorry for the delay.

__tests__/basic.js
```diff
@@ -13,7 +13,7 @@ test('setup', async () => {
   const { driver, pageObjects } = await setupWebDriver();
 
   await driver.wait(uiConnected(), timeouts.directLine);
-  await pageObjects.sendMessageViaSendBox('layout carousel');
+  await pageObjects.sendMessageViaSendBox('layout carousel', { waitForSend: true });
```
compulim: `waitForSend` default to `true`, so you don't need to set it.

My general rationale around naming of option is, "put thing that is odd to the majority there (< 50%)". In this case, most of the time, you want to wait the message to send out.

I am okay to keep it for now, we will do test cleanup later.

__tests__/offlineUI.js
```diff
@@ -186,11 +193,10 @@ describe('offline UI', async () => {
   });
 
   test('should display the "Connecting..." connectivity status when connecting for the first time', async() => {
-    const WEB_CHAT_PROPS = { spinnerAnimationBackgroundImage: staticSpinner };
+    const WEB_CHAT_PROPS = { styleOptions: { spinnerAnimationBackgroundImage: staticSpinner } };
```
compulim: No wonder some of them are not working. 😂

__tests__/offlineUI.js
```diff
@@ -263,17 +270,31 @@ describe('offline UI', async () => {
...
+
+    // TODO: [P4] Understand why we need to fire tick() using two cross-VM calls
+    //       When we put everything in a single cross-VM call, the last tick has no effect
+    await driver.executeScript(() => {
+      window.WebChatTest.clock.tick(1); // Shortly after 15s, it will show "Network interruption occured."
     });
 
-    await driver.sleep(600);
     const base64PNG = await driver.takeScreenshot();
     expect(base64PNG).toMatchImageSnapshot(imageSnapshotOptions);
```
compulim: Can you add a new line between `const` and `expect(...)`?

At some points later, I am thinking to collapse these two lines into a single line. Will see.

packages/core/src/reducers/connectivityStatus.js
```diff
@@ -12,6 +12,8 @@ import {
 
 import { DISCONNECT_FULFILLED } from '../../lib/actions/disconnect';
```
compulim: Not your fault. But, I believe this should no go all the way up to `../../lib/actions/disconnect`. Should be `../actions/disconnect`. We can fix this later.

README.md
```diff
@@ -279,6 +279,7 @@ There are several properties that you might pass into your Web Chat React Compon
...
 | `attachmentMiddleware`     | A chain of middleware that allows the developer to add their own custom HTML Elements on attachments. The signature is the following: `options => next => card => next(card)`.                                                                                                                                                                                 |
 | `attachmentRenderer`       | The "flattened" version of `attachmentMiddleware`.                                                                                                            |
+| `createStore`              | A chain of middleware that allows the developer to modify the store actions. The middleware signature is the following: `createStore: ({}, ({ dispatch }) => next => action => next(cardAction)`                                                                                   |
```
compulim: Sort.

__tests__/basic.js
corinagum: When I didn't set it previously the tests were failing, which is why I added them. Can I leave as is?

__tests__/offlineUI.js
```diff
@@ -50,17 +49,25 @@ describe('offline UI', async () => {
         };
       },
       pingBotOnLoad: false,
-      setup: () => new Promise(resolve => {
-        const scriptElement = document.createElement('script');
+      setup: () =>
+      Promise.all([
+        window.WebChatTest.loadScript('https://unpkg.com/core-js@2.6.3/client/core.min.js'),
```
cwhitten: why are we installing these dependencies inline with the tests? is there another purpose than this?

compulim: We want to mock DirectLineJS in this test. And DirectLineJS use Observable. Thus, we need to temporarily include `core-js` in the browser to provide Observable for us to mock DLJS.

cwhitten: Got it!

PR #1938, opened 2019-04-29
Remove a layer of URLSearchParams

compulim:
Edge fails if `new URLSearchParams(new URLSearchParams())`, but Chrome and Firefox passed.

We should remove the extra `new URLSearchParams()` here.


compulim:
@corinagum I tested this PR and it fixed the problem.

PR #1949, opened 2019-05-01
Fix extra padding in the bottom of the page for IE11 and Firefox

### LINKED ISSUE #1943: Web Chat v4 shows two vertical scroll bars in IE11, Edge and Firefox
image

Using Web Chat `master` bits

Repro steps: Open Mock Bot in IE11
Type `help`
note that there are two scroll bars on the right

Edge: 
image

Firefox: 
image

compulim:
> Fix #1943.

The extra `<div>` above `<svg>` is causing the issue.

Updated all snapshots because the SVG is 0.5 pixels smaller. No other visible differences otherwise. Similar to the one below.

basic-js-setup-1-diff


cwhitten:
35.8% coverage increase? What?!

corinagum:
!!! 58% total coverage is where we've been sitting for a while... I don't know where that 35.8% is coming from.

PR #1988, opened 2019-05-14
Updated Markdown-it render options

### LINKED ISSUE #1429: Markdown render texts in single paragraph into two lines
Type `echo "Same line.\nSame line. \n2nd line.\n\r3rd line."` in MockBot. It shows:

image

Instead, both "Same line." should lies on the same line. When trying out with CommonMark and Markdown-It without Web Chat, the result is correct.

image

But in Web Chat, it renders like:

image

tdurnford:
Fixes #1429 

## Changelog Entry
Fix [#1429](https://github.com/Microsoft/BotFramework-WebChat/issues/1429). Changed Markdown-it options to render newline characters correctly, by [@tdurnford](https://github.com/tdurnford) in PR [#1988](https://github.com/Microsoft/BotFramework-WebChat/pull/1988)

## Description
When Markdown-it's `breaks` option is set to true, the renderer interprets all newline characters (`\n`) as line break elements (`<br />`), which is not the desired behavior for Web Chat. Setting the option to false changes the renderer's interpretation of a newline character to a new line within the paragraph tag unless it is preceded by two space characters. In that case, the renderer converts the newline character to a line break.

image

## Specific Changes
  - Changed the Markdown-it `breaks` option from true to false in renderMarkDown

- [x] Testing Added


corinagum:
lol just FYI you referenced the wrong PR ;)

packages/bundle/src/__tests__/renderMarkdown.spec.js
```diff
@@ -4,4 +4,8 @@ describe('renderMarkdown', () => {
   it('should render markdown', () => {
     expect(renderMarkdown('**Hello!**')).not.toBeFalsy();
   });
+
+  it('should properly render newline characters to markdown', () => {
+    expect(renderMarkdown('Same line.\nSame line.  \n2nd line.\n\r3rd line.')).toBe("<p>Same line.\nSame line.<br />\n2nd line.</p>\n<p>3rd line.</p>\n");
```
compulim: Single quote.

Thanks for adding this test, it's thoughtful.

PR #1996, opened 2019-05-16
Bump to 4.4.2-0

compulim:


corinagum:
What changes are being cherry-picked into 4.4.2?

compulim:
We aren't doing anything, but just marking the package as 4.4.2, so the dailies won't be 4.4.1-master-abcdef. They should be 4.4.2-master-abcdef.

PR #1999, opened 2019-05-17
Fix eslint

compulim:
Fixes eslint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 -->
 
 ### Added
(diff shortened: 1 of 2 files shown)


cwhitten:
the defaultProps are not necessary for these components?

compulim:
Yes, I was thinking the same at the first thought.

But looking closely, we are only passing `children` as prop to the `CarouselActivity`, not the `timestampClassName` because it's from the closure.

Ummm... I think I should refactor `CarouselActivity`/`StackActivity` in their own JS file for performance. My brain was quite working yesterday.

corinagum:
should we close this?

compulim:
No need to close, after twinkling it for a while (made new code, refactor them), I am back to ground zero.

I added a comment because apparently, the code is good. It is not a React functional component, but a middleware function. So no need prop types. 😄

PR #2003, opened 2019-05-20
Fix SAMPLES.md

nt-7:
## Details
- Correct link : https://github.com/Microsoft/BotFramework-WebChat/blob/master/packages/component/src/Styles/defaultStyleOptions.js
- Broken link : https://github.com/Microsoft/BotFramework-WebChat/tree/master/packages/component/src/Styles/defaultStyleSetOptions.js


corinagum:
My comment got eaten up, but thanks so much for the fix!

Issue #1829, opened 2019-03-20
 Text to Speech integrated with LUIS BOT does not voice out/playback in Safari browser on iOS/iPhone devices


corinagum:
Customer is facing problems using Speech API on a LUIS BOT where, Text to Speech/voice out/playback is not working in Safari browser on iOS devices, though it works on both Android and Windows clients. 
Here is a sample URL for testing which I have hosted.

Related to #995

compulim:
Need more investigations.

- Go to https://compulim.github.io/web-speech-cognitive-services/ and try if text-to-speech is working on iOS or not.
   - If it is not, find out the issue in that repository
   - If yes, it could be related to "playing audio without an user interaction"
      - Prime the audio playback engine when the user click microphone button on Web Chat

compulim:
The exception thrown at this line.

![image](https://user-images.githubusercontent.com/1622400/58120939-1c92e300-7bbb-11e9-9b06-5cd8291562b1.png)

![image](https://user-images.githubusercontent.com/1622400/58120911-0ab14000-7bbb-11e9-83c9-df4a952f36d1.png)

compulim:
The core bug of this issue is because we asked the speech synthesizer to cancel if it cannot start within a second. The kill-and-retry logic was used to workaround a Chrome bug and it no longer repro today.

Since we are using Cognitive Services, it could take more than a second to download and start playing the MP3. Thus, it is not very reliable today.

PR #2016, opened 2019-05-21
Should not stop after first spoken text

### LINKED ISSUE #1736: Speech: Queuing messages for text-to-speech
Reference: https://github.com/Microsoft/BotFramework-WebChat/issues/1714

Customer had an issue where text-to-speech has been cutting out on Web Chat. To quote the conclusion:

> So when the "additional questions" type message gets into the channel it interrupts the reading task to focus on the new task.
A workaround for now is to stop using "additional questions" type messages.
On the other hand **it would be great if the webchat.js could queue messages to read in queued fashion**.

(emphasis mine) 

Queuing messages would prevent the bot from 'cutting out' and interrupting whatever current utterance is being made. Is queuing messages to prevent interruption of speech something we want to support in the near future? @compulim in particular has done the most work with Speech recently and may have thoughts on the matter.

compulim:
Fixes #1736.

## Changelog Entry

- Fix [#1736](https://github.com/Microsoft/BotFramework-WebChat/issues/1736). Fixed only first activity in a batch is spoken, by [@compulim](https://github.com/compulim) in PR [#2016](https://github.com/Microsoft/BotFramework-WebChat/pull/2016)

## Description

In `startDictateAfterSpeakActivitySaga`, we are using a wrong selector.

Originally, we want to check if "all activities are spoken", but since we used a wrong selector, we was        checking "if the currently speaking activity is spoken". And of course it is, because we just set a flag on it.

## Specific Changes
- Instead of checking against "the currently spoken activity", we are checking against all activities
- ~Testing Added~
   - We don't have any test facility for speech right now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fix [#1980](https://github.com/Microsoft/BotFramework-WebChat/issues/1980). Changed sendBoxTextArea styles to break words longer than the textarea, by [@tdurnford](https://github.com/tdurnford) in PR [#1986](https://github.com/Microsoft/BotFramework-WebChat/pull/1986)
 - Fix [#1969](https://github.com/Microsoft/BotFramework-WebChat/issues/1969). Move `styleSet`s related to Adaptive Cards to full bundle, by [@corinagum](https://github.com/corinagum) in PR [#1987](https://github.com/Microsoft/BotFramework-WebChat/pull/1987)
 - Fix [#1429](https://github.com/Microsoft/BotFramework-WebChat/issues/1429). Changed Markdown-it options to render newline characters correctly, by [@tdurnford](https://github.com/tdurnford) in PR [#1988](https://github.com/Microsoft/BotFramework-WebChat/pull/1988)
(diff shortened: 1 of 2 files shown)


darrenj:
I believe this issue is the one we've hit with testing, here are some examples to help validate the change. Would be good to validate this with a Bot that @ryanlengel can help provide.

•	If one response is sent back WebChat appears to perform TTS reliably, eg:
image

•	If two responses are sent back WebChat fails to perform TTS on anything (despite Speak properties being set). Occasionally it will play the last message, sometimes truncated. Other times nothing is heard. I suspect the first is being played, the second comes in and does not wait until the first is completed. Due to the race condition it’s luck of the drawer if you hear anything.
image

•	This I believe is another example of “two responses” – “I found the following” and then a carousel of adaptive cards. Speak properties are set on both yet nothing is heard.

image

compulim:
@darrenj yes, this one is related.

The bug this PR will fix is, when we finish speaking *any* activities, we will start dictating and stop speaking.

Instead of "any", it should be "all". That's the bug.

The other bug about "cut off" or "not talking at all", is #1829. What I believe is:
- Chrome has a bug (in Feb) that it may not know it stuck with speech synthesis
   - Our workaround: if Chrome didn't respond in a second, we will restart the speech synthesis part
- If Chrome takes more than 1 second to decode the audio, we will restart the synthesis

So that's why if the text is too long, you may or may not able to hear it. Working on it now.

PR #2018, opened 2019-05-21
Fix 1876 a11y

### LINKED ISSUE #1876: Screen readers narrate all the messages from the beginning for every new message
Screen readers narrate all the messages from the beginning for every new message

corinagum:
Fixes #1876 

## Changelog Entry
- Fix [#1876](https://github.com/Microsoft/BotFramework-WebChat/issues/1876). A11y fixes on Web Chat transcript, by [@corinagum](https://github.com/corinagum) in PR [#2018](https://github.com/Microsoft/BotFramework-WebChat/pull/2018)

## Description
This was an extended dive into a11y so that the Narrator, NVDA, and VoiceOver will only read changes in the transcript once

## Specific Changes

- [x] Explicitly designate the aria-atomic and aria-relevant attributes
- [x] Remove nested live attributes
- [x] Remove timestamp from the "User sent __" and "Bot sent __" aria-label strings
- [x] Change aria-label of the <SendStatus> component to be an absolute timestamp for the "xMinutesAgo" feature, but remain as "Sending" before send has completed. This will presumably prevent the screen reader from repeatedly updating all <SendStatus>es in the transcript.
- [x] A chart confirming which browsers and what screen reader we expect Web Chat to work on.


packages/component/src/Activity/Bubble.js
```diff
@@ -23,14 +23,14 @@ const Bubble = ({
   </div>;
 
 Bubble.defaultProps = {
-  'aria-label': '',
+  'aria-hidden': true,
```
compulim: It seems weird to have it by default to `true`. People could unknowingly hit a bug.

```suggestion
  'aria-hidden': false,
```

packages/component/src/Activity/Timestamp.js
```diff
@@ -20,13 +22,15 @@ const Timestamp = ({
   </span>;
 
 Timestamp.defaultProps = {
+  'aria-hidden': true,
```
compulim: Also, please default this to `false`.

```suggestion
  'aria-hidden': false,
```

PR #2065, opened 2019-06-08
Added pt-pt localization.

tiagodenoronha:


corinagum:
@tiagodenoronha Thanks so much! Could you do me one favor? On your branch, could you run the command `npm run eslint` on the root folder, and commit any changes that pop up? After that, we are good to go! :)

tiagodenoronha:
@corinagum has to add a line to ignore eslint error due to the word "renderizar"... was it the right call?

corinagum:
@tiagodenoronha looks great! Let's wait for the build to complete and then I will merge this into the project. 

Thanks so much for your help and patience!

PR #2073, opened 2019-06-12
Fixed markdown renderer in Playground

### LINKED ISSUE #2069: Playground is broken due to recent renderMarkdown + CRLF change
## Screenshots

image

## Version

"master"

## Describe the bug

Playground throw exception after trying to connect to Mockbot.

## To Reproduce

1. Browse to https://webchat-playground.azurewebsites.net/
2. Click on "Connect to official Mockbot"

## Expected behavior

User can start talking to Mockbot.

## Additional context

Seems related to recent CRLF changes on renderMarkdown.

[Bug]

tdurnford:
Fixes #2069

 ## Changelog Entry	
Fix [#2069](https://github.com/microsoft/BotFramework-WebChat/issues/2069). Fixed markdown renderer issue in Playground, by [@tdurnford](https://github.com/tdurnford) in PR[#xxx](https://github.com/microsoft/BotFramework-WebChat/pull/xxx)	

 ## Specific Changes	
Updated the Playground to use Web Chat's default markdown renderer. Also made slight modification to the FullReactWebChat component to ensure the attachment middleware isn't falling back to the default markdown renderer when the developer chooses to use their own renderer. 

 ---

### DIFF
--- a/packages/bundle/src/FullReactWebChat.js
+++ b/packages/bundle/src/FullReactWebChat.js
@@ -27,22 +27,24 @@ class FullReactWebChat extends React.Component {
     );
 
     this.memoizeStyleSet = memoize((styleSet, styleOptions) => styleSet || createStyleSet(styleOptions));
-    this.memoizeRenderMarkdown = memoize((renderMarkdown, styleSet) => markdown => renderMarkdown(markdown, styleSet));
+    this.memoizeRenderMarkdown = memoize((renderMarkdown, { options }) => markdown =>
+      renderMarkdown(markdown, options)
+    );
   }
 
   render() {
     const {
       adaptiveCardHostConfig,
       attachmentMiddleware,
+      renderMarkdown,
       styleOptions,
       styleSet,
-      renderMarkdown,
       ...otherProps
     } = this.props;
 
     const memoizedStyleSet = this.memoizeStyleSet(styleSet, styleOptions);
     const memoizedRenderMarkdown =
-      renderMarkdown || this.memoizeRenderMarkdown(defaultRenderMarkdown, memoizedStyleSet.options.markdown);
+      renderMarkdown || this.memoizeRenderMarkdown(defaultRenderMarkdown, memoizedStyleSet);
 
     return (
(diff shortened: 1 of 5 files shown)


packages/bundle/src/FullReactWebChat.js
```diff
@@ -31,10 +31,18 @@ class FullReactWebChat extends React.Component {
   }
 
   render() {
-    const { adaptiveCardHostConfig, attachmentMiddleware, styleOptions, styleSet, ...otherProps } = this.props;
+    const {
+      adaptiveCardHostConfig,
+      attachmentMiddleware,
+      styleOptions,
+      styleSet,
+      renderMarkdown,
```
compulim: Can you check if Prettier is running correctly? This one has sorting problem.

packages/bundle/src/FullReactWebChat.js
```diff
@@ -31,10 +31,18 @@ class FullReactWebChat extends React.Component {
...
+      ...otherProps
+    } = this.props;
 
     const memoizedStyleSet = this.memoizeStyleSet(styleSet, styleOptions);
-    const memoizedRenderMarkdown = this.memoizeRenderMarkdown(renderMarkdown, memoizedStyleSet);
+    const memoizedRenderMarkdown =
+      renderMarkdown || this.memoizeRenderMarkdown(defaultRenderMarkdown, memoizedStyleSet.options.markdown);
```
compulim: I think `memoizedStyleSet.options.markdown` could be reduced to `styleOptions ? styleOptions.markdown : {}`. I am not 100% sure but sounds like it might work.

Less dot is better because member accessors in JavaScript is super expensive due to their take on prototypical OO model.

packages/bundle/src/FullReactWebChat.js
tdurnford: Prettier is running, but it's not alphabetizing deconstructed object's properties.

cwhitten:
Thanks @tdurnford

PR #2079, opened 2019-06-13
Timeouts increased for 4 tests.

BruceHaley:
These 4 tests were timing out at 5000 ms. I "hacked" it by setting the timeout to 30000 ms. I.e. changed from timeouts.directline to timeouts.postActivity. That fixed the timeout errors in the build.

I assume you would prefer a nicer way to set the timeouts than this. For example, define a new constant, timeouts.directLineLong = 30000.

Let me know what you would like and I will fix it.

Thanks.
--Bruce

### DIFF
--- a/__tests__/cardActionMiddleware.js
+++ b/__tests__/cardActionMiddleware.js
@@ -64,15 +64,15 @@ test('card action "signin"', async () => {
     }
   });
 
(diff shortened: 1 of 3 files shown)


BruceHaley:
Changing the fix to simply increase the timeout value for timeouts.directLine. Will issue a new PR.

Issue #1680, opened 2019-01-31
Tests occasionally fail when running on Travis CI


compulim:
Tests occasionally fail the `driver.wait(botConnected())` test, suspecting either slow network or CPU causing the issue.

Retry is not easy because `botConnected()` is in every of our tests. The reason why we prefer `botConnected()` in every test:

- Some tests can use bots other than MockBot
- We will partition MockBot into multiple bots:
   - Using BotAdapter to create a bot that run inside the browser
   - Mocking DirectLineJS to mimic respond from a non-bot

If retry is implemented, please make sure to consider these cases.

corinagum:
Bottom line is to make Mock Bot offline for testing instead of online.

PR #2081, opened 2019-06-13
Increase timeouts.directLine from 5000 to 15000

BruceHaley:
Fixes 4 tests timing out in BotFramework-WebChat-CI-PR.

### DIFF
--- a/__tests__/constants.json
+++ b/__tests__/constants.json
@@ -5,7 +5,7 @@
     }
   },
   "timeouts": {
-    "directLine": 5000,
+    "directLine": 15000,
     "fetch": 2500,
     "navigation": 10000,
     "postActivity": 30000,


compulim:
I believe this one will not work 90% of the time. It is because if Web Chat did not receive the message within 5 seconds, Web Chat probably won't receive the message within 15 seconds. It was because, most of the time, the message is lost in space.

Retry will make it more reliable, we set retry to 3 times now.

compulim:
After discussing offline, this may makes sense because the network infrastructure on VSTS hosted agent may be much slower than our previous agent. And 15s might fix a few issues.

Anyway, we will have #1680 to move Mockbot inside a browser, so we will skip the public Direct Line infrastructure. This could be ultimate fix for this.

Issue #1926, opened 2019-04-25
Scroll event is not triggering every time a message is received


keshavnagpal:
here is a video to explain the issue
https://www.loom.com/share/f22dd7020a0a4c5ea53c3a88be87ffee
(pass: webchat)

i'm using master/webchat-es5 (dated 25-Apr-2019) 

Update:
We implemented our own solution as a workaround for this, but it might not be as reliable as a fix in webchat library, pasting the implementation below:

```
var messageListSelector = '#chat-interface ul[role="list"][aria-live="polite"]';
var scrollControl = false;
When action.type === 'DIRECT_LINE/POST_ACTIVITY'
{
scrollControl = true;
}

When action.type === 'DIRECT_LINE/INCOMING_ACTIVITY'
{
if (scrollControl) {
        $(messageListSelector).parent().animate({ scrollTop: $(messageListSelector).height() }, 1000);
    }
    $('#chat-interface input[data-id="webchat-sendbox-input"]').focus();
}
//if user scroll up
$(document).ready(function(){
    $(messageListSelector).bind('mousewheel DOMMouseScroll', function(event){
        event.stopPropagation();
        scrollControl = false;
        var scrollPos = $(messageListSelector).parent().scrollTop();
        var elemHeight = $(messageListSelector).height();
        var srollSkew = elemHeight - scrollPos;
        if (srollSkew > 600 && srollSkew < 636) {
            scrollControl = true;
        }
    });
});
```

tdurnford:
@keshavnagpal the video link you shared is private. Can you please change the privacy settings on it so we can see the issue. 

This seems to be related to https://github.com/Microsoft/BotFramework-WebChat/issues/1031.

corinagum:
For confirmation, is this occurring in IE11?

keshavnagpal:
@tdurnford I had provided the pass in the post, just removed it for simplicity, video is public now.

@corinagum strangely this doesn't occur in ie11 but occurring in chrome (i have Version 73.0.3683.103 if that helps)

corinagum:
Repro steps:
1. receive an adaptive card from the bot
1. receive suggested actions from the bot
1. click on a button from the adaptive card

compulim:
Investigating if it is because the scrollable container resized while the content is changing. This could be tricky to fix.

Keshav-GravityILabs:
Hi, I tried master/webchat-es5.js
The issue seems to be still there
[...]

PR #2099, opened 2019-06-20
Fixed scroll event trigger problem

vefacaglar:
Fixes #

[#1926](https://github.com/microsoft/BotFramework-WebChat/issues/1926)

## Specific Changes

 

---

-  [ ] Testing Added


corinagum:
@vefacaglar thanks for the PR. It looks like there may have been a rebase issue that needs to be fixed - your other PR is showing up again in the Files changed section.

I know @compulim is working on this now, so I will defer to him for communication on coordinating the fix.

vefacaglar:
i can’t understand too why these changes showing up again.

thanks for your interest.

compulim:
Thanks for trying to fix the problem.

Although this would fix it, it sounds a bit hacky to fix it by forcing it to scroll to bottom always.

The main problem is when the browser (mainly Chrome variant) resize the element, they send a "scroll" event even the user did not scroll. This is for maintaining the scroll position for better UX in Chrome.

And our scroll to bottom think this scroll event is done by the user. Thus, we stopped being sticky to the bottom because "the user scrolled up, we shouldn't stick to the bottom". But in fact, we should continue to stick to bottom because it is not user's intention to "un-stick".

compulim:
The translation issue is already in. Since this fix is hacky and not fixing the root cause, I am closing this for now. Thanks for the effort.

PR #2115, opened 2019-06-23
Added richCardsWrapTitle to defaultStyleOptions

### LINKED ISSUE #2046: thumbnailCard title too long will be ellipsis
I used thumbnailCard with long title.
The Text will be ellipsis if the text in title is too long.
Is there any options to change  thumbnailCard title style for this case

[Edit by corinagum - Adding more detail from below]
---
The title of thumbnailCard  is showed with ellipsis. 
And in Devtools , I could see all text of the title .
Is there any options to change thumbnailCard title style In web chat .
So I could see the all of them , no by ellipsis.

image

image

```
let buttons = [];
            let button1 = {};
            button1.type = 'imBack';
            button1.title = "one";
            button1.value = "one";
            buttons.push(button1);

            let button2 = {};
            button2.type = 'imBack';
            button2.title = "two";
            button2.value = "two";
            buttons.push(button2);

            let button3 = {};
            button3.type = 'imBack';
            button3.title = "three";
            button3.value = "three";
            buttons.push(button3);

            await dialogContext.context.sendActivity({ attachments: [CardFactory.thumbnailCard("abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890abcdefghij1234567890", '', [], buttons)] });
```

tdurnford:
Fixes #2046 

## Changelog Entry
Added richCardsWrapTitle to defaultStyleOptions, by [@tdurnford](https://github.com/tdurnford), in PR [#2115](https://github.com/Microsoft/BotFramework-WebChat/pull/2115)

## Specific Changes
Added richCardsWrapTitle to defaultStyleOptions and added styleOptions as a parameter to the AdaptiveCardsBuilder and as a prop for the CommonCard component. Developers should be able to wrap the title attribute in Rich Cards by setting richCardsWrapTitle in styleOptions to true.

image

---

-  [ ] Testing Added


corinagum:
Let's get a test on this one :)

compulim:
Looking at this one.

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardBuilder.ts
```diff
@@ -104,7 +106,13 @@ export default class AdaptiveCardBuilder {
...
   addCommonHeaders(content: ICommonContent) {
-    this.addTextBlock(content.title, { color: TextColor.Dark, size: TextSize.Medium, weight: TextWeight.Bolder });
+    const { richCardsWrapTitle } = this.styleOptions;
+    this.addTextBlock(content.title, {
+      color: TextColor.Dark,
+      size: TextSize.Medium,
+      weight: TextWeight.Bolder,
+      wrap: richCardsWrapTitle
+    });
     this.addTextBlock(content.subtitle, { color: TextColor.Dark, isSubtle: true, wrap: true });
```
compulim: I think we should also use the same `richCardsWrapTitle` for `subtitle`, because subtitle is a type of title too.

packages/bundle/src/adaptiveCards/Attachment/CommonCard.js
```diff
@@ -39,5 +39,6 @@ CommonCard.propTypes = {
     content: PropTypes.shape({
       tap: PropTypes.any
     }).isRequired
-  }).isRequired
+  }).isRequired,
+  styleOptions: PropTypes.any.isRequired
```
compulim: I think `CommonCard` should use `connectToWebChat` to obtain `styleSet.options` directly.

packages/component/src/Styles/defaultStyleOptions.js
```diff
@@ -49,6 +49,9 @@ const DEFAULT_OPTIONS = {
   // Markdown
   markdownRespectCRLF: true,
 
+  // Rich Cards
+  richCardsWrapTitle: false,
```
compulim: I think we should name it `richCardWrapTitle`.

Could you put a comment here and say it will also applies to subtitle?

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardBuilder.ts
tdurnford: Would it make more sense to add `richCardWrapSubtitle` to the default style options? I think right now subtitles are designed to wrap and titles don't. Using `richCardWrapTitle` for both might change Web Chat's card behavior too much.

tdurnford:
The Support Team has seen an increase in issues regarding the title not wrapping, especially around QnA Prompting. I believe the SDK is mapping the prompts to Hero Cards and is adding the text response as the card's title. More often than not, the text is too long and gets truncated. We've been telling clients to either move the title to the card's text attribute or map the QnA prompt to an Adaptive Card themselves. 

I'm torn on wrapping the title by default. I think it might be better to wait until we move Rich Cards away from using the Adaptive Card builder.

compulim:
After discussing offline:
- One option to set both title and subtitle
- Default to either true or false, if false, then it's less breaking

Issue #1863, opened 2019-02-14
QnAMaker responses are said twice when using webchat


gabog:
**Project**
Virtual Assistant

**Description**
When you use speech, QnAMaker responses seem to be read twice when using webchat (this doesn't seem to happen in the emulator)

**To Reproduce**
1. Deploy VA to webchat with speech enabled.
2. Say "What is your name" using speech
3. You'll hear "My name is the Virtual Assistant! Nice to meet you :)" twice.

**Expected behavior**
It should only say the response once.

compulim:
## Background

All rich cards, including thumbnail card, is converted to Adaptive Cards on render.

## Investigation steps

1. Launch Narrator
1. Load Web Chat in Edge, thru microsoft.github.io/BotFramework-WebChat/01.a.getting-started-full-bundle/
1. Type "card bingsports"

If the problem persists, we should fix it. If not, we should try a ThumbnailCard and see if it repro.

### Expected

Narrator should only speak out "The Seattle Seahawks beat the Carolina Panthers 40-7", but not the content.

compulim:
@corinagum could you investigate with the repro steps above and see if it repro?

One thing Gabo pointed out "If the qna answer gets deserialized in the thumbnail card and the Speak property is populated, Speech reads the Speak property AND the text on the card."

corinagum:
@gabog, are you still having this issue? I am not able to reproduce this on Mock Bot with a Thumbnail card. **Could you clarify whether you are using Narrator, Cognitive Services Speech Services, or both?** If Narrator, I believe this is an Adaptive Cards problem (see below). If Cognitive Services, then I will need more information to reproduce.

Regarding Narrator: I tried the Narrator steps that @compulim applied above, but I think they show a different bug than what @gabog reported. This has to do with how Adaptive Cards handles `aria-label`, which provides a high level description of the card via the speak property, but doesn't allow the user to hide irrelevant containers or text, or a way to manually provide aria-labels that improve the understanding of the card. If this is the issue you're having, @gabog, please let me know and I will close this issue. I recommend filing this again with the Adaptive Cards team. 

If this is an ongoing Cognitive Services issue and Narrator is not involved, please copy-paste our 'New issue' template for reporting bugs and provide the information requested in that template. (You don't have to file a new issue, continuing the conversation in this thread is fine). Thanks!

gabog:
Hi, below is the repro code:

The minimalistic code to generate the card in C# is: 

[...]

PR #2118, opened 2019-06-24
Fix Speech speaking the title, subtitle, and text of cards

corinagum:
## Changelog Entry
-  Fix [#1863](https://github.com/Microsoft/BotFramework-WebChat/issues/1863). Remove title, subtitle, and text of cards from being spoken by [@corinagum](https://github.com/corinagum) in PR [#2118](https://github.com/Microsoft/BotFramework-WebChat/pull/2118)

## Description

In v3 of Web Chat, Speech did not read out title, subtitle, and text of cards. This fix brings parity back to that version. Previously, Speech in v4 would speak all three and cause 'doubling' of spoken text.

-  [ ] Testing Added
   
No testing added since this is a speech change.

### DIFF
--- a/packages/component/src/Activity/Speak.js
+++ b/packages/component/src/Activity/Speak.js
@@ -39,13 +39,8 @@ const Speak = ({ activity, markAsSpoken, selectVoice, styleSet }) => {
   const lines = [speak || text];
 
   attachments.forEach(({ content: { speak } = {}, contentType }) => {
-    switch (contentType) {
-      case 'application/vnd.microsoft.card.adaptive':
-        lines.push(speak);
-        break;
-
-      default:
-        break;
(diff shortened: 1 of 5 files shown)


packages/component/src/Activity/Speak.js
```diff
@@ -38,26 +38,12 @@ const Speak = ({ activity, markAsSpoken, selectVoice, styleSet }) => {
 
   const lines = [speak || text];
 
-  attachments.forEach(({ content: { speak, subtitle, text, title } = {}, contentType }) => {
+  attachments.forEach(({ content: { speak } = {}, contentType }) => {
     switch (contentType) {
```
tdurnford: Should this be changed to an if statement? Since we're only adding the speak attribute for AdaptiveCards, it seems a little excessive to have it as a switch.

PR #2140, opened 2019-07-02
Fix observeOnce to unsubscribe properly

### LINKED ISSUE #1872: `observeOnce` did not unsubscribe correctly
## Background

https://github.com/Microsoft/BotFramework-WebChat/blob/master/packages/core/src/sagas/effects/observeOnce.js#L11

In this line of code, we didn't assign the result of `observable.subscribe()` to `subscription`. Thus, `subscription` is always empty and thus, we never call `unsubscribe()`.

## Things to-do

1. Modify https://github.com/Microsoft/BotFramework-WebChat/blob/master/packages/core/src/sagas/effects/observeOnce.js#L11
   - Assign the result of `observable.subscribe()` to `subscription` variable
1. Verify `unsubscribe()` is being called correctly
1. Verify all callers of `observeOnce` will continues to work

compulim:
Fixes #1872.

## Changelog Entry

-  Fix [#1872](https://github.com/Microsoft/BotFramework-WebChat/issues/1872). Fixed `observeOnce` to unsubscribe properly, by [@compulim](https://github.com/compulim) in PR [#2140](https://github.com/Microsoft/BotFramework-WebChat/pull/2140)

## Description

Our `observeOnce` saga effect does not unsubscribe itself from the observable. New code added to unsubscribe after first `onNext`, `onError`, or `onCompleted`.

## Specific Changes

Modified `packages/core/src/sagas/effects/observeOnce.js` to unsubscribe regardless of conditions.

---

-  [x] Testing Added

### DIFF
--- a/packages/core/src/sagas/effects/observeOnce.js
+++ b/packages/core/src/sagas/effects/observeOnce.js
@@ -8,7 +8,7 @@ export default function observeOnceEffect(observable) {
       return yield call(
         () =>
           new Promise((resolve, reject) => {
-            subscription = observable.subscribe(resolve, reject);
+            subscription = observable.subscribe(resolve, reject, resolve);
           })
       );
     } finally {
--- a/packages/core/src/sagas/effects/observeOnce.js
+++ b/packages/core/src/sagas/effects/observeOnce.js
@@ -8,13 +8,11 @@ export default function observeOnceEffect(observable) {
       return yield call(
         () =>
           new Promise((resolve, reject) => {
-            observable.subscribe(resolve, reject);
+            subscription = observable.subscribe(resolve, reject);
           })
       );
     } finally {
-      if (subscription) {
-        subscription.unsubscribe();
-      }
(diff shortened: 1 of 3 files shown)


packages/core/src/sagas/effects/observeOnce.js
```diff
@@ -8,13 +8,11 @@ export default function observeOnceEffect(observable) {
       return yield call(
         () =>
           new Promise((resolve, reject) => {
-            observable.subscribe(resolve, reject);
+            subscription = observable.subscribe(resolve, reject, resolve);
```
corinagum: Can you explain why next and complete are the same? to me it seems like this would make resolve be called twice. :/

compulim: `observeOnce` is essentially converting a stream into an async binary operation (success or fail in async fashion, more-or-less Promise).

For "complete without any results", I think it should be considered as "positive with result of `undefined`".

PR #2218, opened 2019-07-23
Send Button remains enabled all time even when the text is not present in the textbox

cmevawala:
## Description
When the send button component is getting created, we need to check whether there is any value in textbox or not and based on that set the disabled property

## Specific Changes
Changes are done only in SendButton.js where we get the **sendBoxValue** and we used that value to set the disabled property


packages/component/src/SendBox/SendButton.js
```diff
@@ -14,16 +14,21 @@ const connectSendButton = (...selectors) =>
...
     ...selectors
   );
 
-const SendButton = ({ click, disabled, language }) => (
-  <IconButton alt={localize('Send', language)} disabled={disabled} onClick={click}>
-    <SendIcon />
-  </IconButton>
-);
+const SendButton = ({ click, disabled, language, sendBoxValue }) => {
+  disabled = sendBoxValue ? false : true;
```
corinagum: My concern is that this is overwriting the props value `disabled`. @compulim, thoughts?

corinagum:
Looking for design input from @DesignPolice

corinagum:
Hi @cmevawala, unfortunately I don't think this code is achieving what you intend. These changes will actually disable the send box and prevent a user from entering any message to the bot. Please take a look at the following link: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_input_disabled

I am closing this issue since this is not a change we want to implement.

packages/component/src/SendBox/SendButton.js
cmevawala: @corinagum : Here are snapshots of the issue which i have fixed

image

image

compulim:
I think the better fix is:

```jsx
<IconButton alt={localize('Send', language)} disabled={disabled || !sendBoxValue} onClick={click}>
```

PR #2253, opened 2019-07-31
Fix wrong shape on propTypes.tapAction

compulim:
Fixes #2250.

Also need to fix `card broken` command.

image

image

## Changelog Entry

-  Fix [#2250](https://github.com/microsoft/BotFramework-WebChat/issues/2250). Fix React warnings related prop types, by [@compulim](https://github.com/compulim) in PR [#2253](https://github.com/microsoft/BotFramework-WebChat/pull/2253)

## Description

Wrong `propTypes` on `AdaptiveCardRenderer.js`.

## Specific Changes

- Modify `AdaptiveCardsRenderer.propTypes.tapAction`

---

-  [x] ~Testing Added~

PR #2255, opened 2019-08-01
Fix wrong shape on propTypes.tapAction (#2253)

compulim:
* Fix wrong shape on propTypes.tapAction
* Update PR number
* Should pass language
* Pass language and rename to aria-label


compulim:
Self-approving cherry-picks.

PR #2384, opened 2019-09-06
Added botAvatarBackgroundColor and userAvatarBackgroundColor to the default style options

### LINKED ISSUE #2306: No option to change font size in defaultStyleOptions.js
I am trying to change the font size of my chat's messages (both from the user and the bot) but there does not seem to be any option to do so in the styleOptions.

I have seen a way to change the font size using the window.WebChat.createStyleSet method. 

However, my bot will be used on IE11 as well of modern browsers, and this way is not supported on IE11.

### LINKED ISSUE #2380: BOT Avatar has a blue circle background
The avatar image is not taking full background and showing blue circle line around. Not looking good. Though the image is having all white background.

## Screenshots
image

## Additional context

[Bug]

tdurnford:
Fixes #2380 
Fixes #2306

## Changelog Entry
-  Fix [#2380](https://github.com/microsoft/BotFramework-WebChat/issues/2380). Added `botAvatarBackgroundColor` and `userAvatarBackgroundColor` to the default style options, by [@tdurnford](https://github.com/tdurnford) in PR [#2384](https://github.com/microsoft/BotFramework-WebChat/pull/2384)

## Description
Added botAvatarBackgroundColor and userAvatarBackgroundColor to the default style options. When either is `null` respectively the avatar background color is the accent color.

```javascript
const styleOptions = {
  botAvatarBackgroundColor: 'red',
  botAvatarInitials: 'B',
  userAvatarBackgroundColor: 'blue',
  userAvatarInitials: 'TJ',

  suggestedActionBorderWidth: 4,
  suggestedActionBorderColor: 'silver',
  suggestedActionBorderStyle: 'double',

  suggestedActionDisabledBorderWidth: 1,
  suggestedActionDisabledBorderColor: 'blue',
  suggestedActionDisabledBorderStyle: 'dashed'
};
```

image

---

-  [x] Testing Added


packages/component/src/Styles/defaultStyleOptions.js
```diff
@@ -27,8 +27,10 @@ const DEFAULT_OPTIONS = {
 
   // Avatar
   avatarSize: 40,
+  botAvatarBackgroundColor: null, // defaults to accent color in styleSet
```
compulim: Prefer to put default here, just like `suggestedActionBorder` or `suggestedActionTextColor`. I can see your points about defaulting in the style set.

TBH, we should think thoroughly on how to tackle it better, but not now:

- The dev can easily "steal" the style options JSON without looking into any logic (like `||` in your case), it feel more "definitive" and less bug prone to them ("I changed it, why it didn't change the color?")
- Although there are more places need to be changed, these types of devs are used to "replace all" (CSS modders are like this)
- Think thoroughly because we always have some "arch options" like accent color, text color, or a few things that actually seed the style options, but try not to introduce another layer

tdurnford: My first thought was to set the background colors to `DEFAULT_ACCENT` in the default style options, but then the avatar background color wouldn't change when the developer changed the accent color. Would be a breaking change.

corinagum: Is this convo resolved and PR ready for merge?

tdurnford: Yea I added a comment in the PR description about our offline discussion.

samples/01.a.getting-started-full-bundle/index.html
```diff
@@ -7,7 +7,7 @@
       This CDN points to the latest official release of Web Chat. If you need to test against Web Chat's latest bits, please refer to pointing to Web Chat's MyGet feed:
       https://github.com/microsoft/BotFramework-WebChat#how-to-test-with-web-chats-latest-bits
     -->
-    <script src="https://cdn.botframework.com/botframework-webchat/latest/webchat.js"></script>
+    <script src="/packages/bundle/dist/webchat.js"></script>
```
corinagum: Undo this change

tdurnford: Thaaaaaaanks

corinagum:
@compulim ping, did all your change requests make it in?

PR #2392, opened 2019-09-10
Migrated tests to Offline Mockbot

### LINKED ISSUE #1680: Tests occasionally fail when running on Travis CI

tdurnford:
Fixes #1680
## Description
Migrated Web Chat tests from [MockBot](https://github.com/compulim/BotFramework-MockBot/tree/master/src/commands) to [Offline MockBot](https://github.com/tdurnford/BotFramework-Offline-MockBot)

## Specific Changes
Updated Web Chat tests to run with Offline Mock Bot by default. Developers can choose to have a test run with Online Mock Bot by setting the `offline` option to false when they setup the web driver.
```javascript
const { driver } = await setupWebDriver({ offline: false });
```
---

-  [ ] Testing Added
Tests would be redundant

### DIFF
--- a/__tests__/setup/web/index.html
+++ b/__tests__/setup/web/index.html
@@ -47,6 +47,7 @@
         width: 100%;
       }
     </style>
+    <script src="https://tdurnford.github.io/BotFramework-Offline-MockBot/index.js"></script>
     <script src="https://unpkg.com/event-target-shim@5.0.1/dist/event-target-shim.umd.js"></script>
     <script src="/createProduceConsumeBroker.js"></script>
     <script src="/mockWebSpeech.js"></script>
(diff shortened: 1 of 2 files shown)


__tests__/setup/web/index.html
```diff
@@ -92,27 +93,34 @@
...
-          } catch (err) {
-            console.error('Failed to fetch Direct Line token from Mockbot.');
-            console.error(err);
-
-            throw err;
-          }
-        }, 3);
-
+        let directLine;
+        if (offline !== false && !createDirectLine) {
```
tdurnford: I don't love this line... If `offline` is `undefined` or `false`, I want it to use Offline Mockbot, but `!undefined` is `true` and `!false` is `true so I cant use `if (!offline) {`... Thoughts?

tdurnford: @corinagum @compulim

corinagum: I would add a comment to the html outlining the same thing... to me it's going to confuse someone (me) no matter what, lol

tdurnford: William and I decided to change `offline` to `useProductionBot`. If `useProductionBot` is `false` or `undefined`, the tests will use Offline Mock Bot; otherwise, they will run with the Production version of Mock Bot.

PR #2410, opened 2019-09-19
Added attachment style options

### LINKED ISSUE #2232: Card backgrounds should be different than text backgrounds
## Feature Request

I am noticing that when I change the text background the card background changes the same.  These should 2 entities should be separate in their core styling.  This is very important.  If I change background color, border and radius that should not also affect my adaptive card background.  

would you be able to seperate the concerns?

[Enhancement]

stevkan:
Fixes #2232 

## Changelog Entry

Fix [#2232](https://github.com/microsoft/BotFramework-WebChat/issues/2232). Added attachment style options, by [@stevkan](https://github.com/stevkan) in PR [#2410](https://github.com/microsoft/BotFramework-WebChat/pull/2410)

## Description

Provides the ability to apply separate styling options for attachments (cards) and activities sent from the bot. Addresses attachment background and text color, attachment border styling, and nub border styling.

## Specific Changes

Updated defaultStyleOptions.js:
- Included style options for attachments.
- Altered 'bot' property names to include "FromBot". For example, changed `bubbleBackground` to `bubbleFromBotBackground`.

Updated adaptiveCardHostConfig.js:
- Included `backgroundColor` in `containerStyles` styles.
- Set some `containerStyles` values to `null`. Otherwise, values compile as inline CSS requiring `!important` to be appended to style options in order to override.
  - As default values are providedd via the defaultStyleOptions.js, including `null` values should not have an effect.

Propagated property name changes across project files.

---

-  [ ] Testing Added
Unable to test due to PowerShell/Docker issues. Will require another to perform testing.

### DIFF
--- a/__tests__/bubbleNub.js
+++ b/__tests__/bubbleNub.js
@@ -28,10 +28,12 @@ describe('bubble nub', () => {
     const { driver, pageObjects } = await setupWebDriver({
       props: {
         styleOptions: {
-          bubbleNubOffset: 0,
-          bubbleNubSize: 10,
+          bubbleFromBotNubOffset: 0,
+          bubbleFromBotNubSize: 10,
+          bubbleFromBotNubBorderColor: 'white',
           bubbleFromUserNubOffset: 0,
-          bubbleFromUserNubSize: 10
+          bubbleFromUserNubSize: 10,
+          bubbleFromUserNubBorderColor: 'white'
         }
       },
       zoom: 3
@@ -43,16 +45,18 @@ describe('bubble nub', () => {
   beforeEach(() => {
     props = {
       styleOptions: {
-        bubbleBorderColor: 'red',
-        bubbleBorderRadius: 10,
-        bubbleBorderWidth: 2,
+        bubbleFromBotBorderColor: 'red',
+        bubbleFromBotBorderRadius: 10,
+        bubbleFromBotBorderWidth: 2,
         bubbleFromUserBorderColor: 'green',
         bubbleFromUserBorderRadius: 10,
         bubbleFromUserNubOffset: 0,
         bubbleFromUserNubSize: 10,
+        bubbleFromUserNubBorderColor: 'white',
         bubbleFromUserBorderWidth: 2,
-        bubbleNubOffset: 0,
-        bubbleNubSize: 10
+        bubbleFromBotNubOffset: 0,
+        bubbleFromBotNubSize: 10,
+        bubbleFromBotNubBorderColor: 'white'
       }
     };
   });
@@ -140,7 +144,7 @@ describe('bubble nub', () => {
           styleOptions: {
             ...props.styleOptions,
             bubbleFromUserNubOffset: 5,
-            bubbleNubOffset: 5
+            bubbleFromBotNubOffset: 5
           }
         },
         zoom: 3
@@ -156,7 +160,7 @@ describe('bubble nub', () => {
           styleOptions: {
             ...props.styleOptions,
             bubbleFromUserNubOffset: 10,
-            bubbleNubOffset: 10
+            bubbleFromBotNubOffset: 10
           }
         },
         zoom: 3
@@ -172,7 +176,7 @@ describe('bubble nub', () => {
           styleOptions: {
             ...props.styleOptions,
             bubbleFromUserNubOffset: -5,
-            bubbleNubOffset: -5
+            bubbleFromBotNubOffset: -5
           }
         },
         zoom: 3
@@ -188,7 +192,7 @@ describe('bubble nub', () => {
           styleOptions: {
             ...props.styleOptions,
             bubbleFromUserNubOffset: -10,
-            bubbleNubOffset: -10
+            bubbleFromBotNubOffset: -10
           }
         },
         zoom: 3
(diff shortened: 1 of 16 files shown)


__tests__/bubbleNub.js
```diff
@@ -43,16 +45,18 @@ describe('bubble nub', () => {
   beforeEach(() => {
     props = {
       styleOptions: {
-        bubbleBorderColor: 'red',
```
tdurnford: We can't remove`bubbleBorder*` from the style options because customers might be using these options. Removing them would be a breaking change. Instead set the default value to `undefined`  and in the create style set function use `bubbleBorder*` if it's provided; otherwise, use `bubbleFromBotBorder*`

__tests__/bubbleNub.js
```diff
@@ -43,16 +45,18 @@ describe('bubble nub', () => {
...
+        bubbleFromBotBorderRadius: 10,
+        bubbleFromBotBorderWidth: 2,
         bubbleFromUserBorderColor: 'green',
         bubbleFromUserBorderRadius: 10,
         bubbleFromUserNubOffset: 0,
         bubbleFromUserNubSize: 10,
+        bubbleFromUserNubBorderColor: 'white',
         bubbleFromUserBorderWidth: 2,
-        bubbleNubOffset: 0,
-        bubbleNubSize: 10
```
tdurnford: Same thing as above

packages/bundle/src/adaptiveCards/Styles/StyleSet/AdaptiveCardRenderer.js
```diff
@@ -1,10 +1,32 @@
...
+    '& p': {
+      color: bubbleAttachmentTextColor
+    },
+
     '& .ac-pushButton': {
       backgroundColor: 'White',
       borderStyle: 'solid',
       borderWidth: 1,
-      color: accent,
+      color: 'accent',
```
tdurnford: Accent shouldn't be in quotes.

packages/bundle/src/adaptiveCards/Styles/adaptiveCardHostConfig.js
```diff
@@ -1,31 +1,32 @@
...
         },
         accent: {
-          default: accent,
+          default: null,
           subtle: accent
         }
       }
     },
     emphasis: {
-      backgroundColor: cardEmphasisBackgroundColor,
```
tdurnford: Why is this being removed. I think there are a lot of unnecessary changes to host config.

packages/component/src/Styles/defaultStyleOptions.js
```diff
@@ -5,14 +5,13 @@ function fontFamily(fonts) {
 }
 
 const DEFAULT_ACCENT = '#0063B1';
-const DEFAULT_SUBTLE = '#767676'; // With contrast 4.5:1 to white
 const PADDING_REGULAR = 10;
+const DEFAULT_SUBTLE = '#767676'; // With contrast 4.5:1 to white
```
tdurnford: Why did this like get moved? Should be in alphabetical order

packages/component/src/Styles/defaultStyleOptions.js
```diff
@@ -33,26 +32,34 @@ const DEFAULT_OPTIONS = {
   userAvatarInitials: '',
 
   // Bubble
-  bubbleBackground: 'White',
```
tdurnford: Removing style options is a breaking change. We want to support the legacy options as well as the new ones. Instead of removing, `bubble*`, the values should be set to `undefined`, and then in the style options we should use `bubble*` if it's not falsy; otherwise, use `bubbleFromBot*`.

__tests__/bubbleNub.js
```diff
@@ -28,10 +28,12 @@ describe('bubble nub', () => {
     const { driver, pageObjects } = await setupWebDriver({
       props: {
         styleOptions: {
-          bubbleNubOffset: 0,
```
tdurnford: You should be adding new tests instead of modifying the old ones.

packages/bundle/src/adaptiveCards/Styles/StyleSet/AdaptiveCardRenderer.js
stevkan: Was testing a feature, forgot to change back. Will correct.

packages/bundle/src/adaptiveCards/Styles/adaptiveCardHostConfig.js
stevkan: I believe you are referring to cardEmphasisBackgroundColor(?). As I mention in the PR description, setting actual values (vs null) in the HostConfig compiles to inline CSS on those elements. This then requires the developer to have to append '!important' to any styling changes for them to apply as inline has higher priority.
As these values are set in the defaultStyleOptions file, it's actually redundant and unnecessary to specify them here, as well. Plus (as mentioned), it amounts to extra work for the dev.

packages/component/src/Styles/defaultStyleOptions.js
stevkan: I'll move it back. Didn't look close enough. Thought it said 'DEFAULT_REGULAR'.

__tests__/bubbleNub.js
stevkan: Noted!

corinagum:
Ping on this PR, how's it going @stevkan? 

(@CoHealer for FYI)

corinagum:
Closing due to inactivity. Feel free to resume discussion on #2232

PR #2414, opened 2019-09-22
Language of timestamp should update on-the-fly

### LINKED ISSUE #2360: Timestamp locale does not change when updating language on-the-fly
## Screenshots

image

## Version

`master`

## Describe the bug

When `locale` prop is updated on-the-fly, the timestamp doesn't change accordingly.

## To Reproduce

1. Go to playground
2. Send some messages
3. Change the language to another one

## Expected behavior

Timestamp is immediately localized in another language.

Actual, it is kept the same until page refresh.

## Additional context

Playground is currently broke, change line 246 like below.

image

[Bug]

compulim:
> Fixes #2360.

### Fixed

-  Fix [#2360](https://github.com/microsoft/BotFramework-WebChat/issues/2360). Timestamp should update on language change, by [@compulim](https://github.com/compulim) in PR [#2414](https://github.com/microsoft/BotFramework-WebChat/pull/2414)

## Description

During our last React function component push, a bug was introduced to the `RelativeTime.js`. If language change, the display didn't adopt.

## Specific Changes

- Fix the bug in `RelativeText.js`
- Add tests
- Update test harness with a new `updateProps()` function, to introduce changes to the props, so we can modify language after the component is rendered

---

-  [x] Testing Added


__tests__/setup/web/index.html
```diff
@@ -137,15 +137,29 @@
...
           directLine,
           store,
           styleSet: createStyleSet && createStyleSet(props.styleOptions),
           username: 'Happy Web Chat user',
           ...props
-        }, document.getElementById('webchat'));
+        };
+
+        renderWebChat = props => {
+          console.log(props);
```
corinagum: remove console?

compulim: 😭

PR #2419, opened 2019-09-22
Fix waste render on connectivity status

### LINKED ISSUE #2418: Connectivity status should not waste-render every 400ms
## Version

`master`

## Describe the bug

When putting a `console.log()` inside the render function of `DebouncedConnectivityStatus`, it was called every 400 ms due to wasted render.

## To Reproduce

As described in the above section.

## Expected behavior

It should only render when changed.

## Additional context

This is caused by recent React functional component push.

[Bug]

compulim:
Fixes #2418.

### Fixed

-  Fix [#2418](https://github.com/microsoft/BotFramework-WebChat/issues/2418). Connectivity status should not waste-render every 400 ms, by [@compulim](https://github.com/compulim) in PR [#2419](https://github.com/microsoft/BotFramework-WebChat/pull/2419)

## Description

`DebouncedConnectivityStatus` is causing wasted render for every 400 ms.

## Specific Changes

This is due to misunderstanding of `useState` for assigning a function.

Instead of,

```js
const [predicate, setPredicate] = useState();

setValue(x => x === 'Hello, World!');
```

It should be,

```js
const [predicate, setPredicate] = useState();

setValue(() => x => x === 'Hello, World!');
```

This is because, if function is passed to `setXXX`, React will call the function with the existing value, and use the return value as the new state. If non-function is passed, React will directly set it as new state. This behavior is currently not documented.

---

-  [x] ~Testing Added~
   - We currently do not have a way to detect how many render done per seconds, etc. We should setup this tool.

### DIFF
--- a/packages/component/src/SendBox/ConnectivityStatus.js
+++ b/packages/component/src/SendBox/ConnectivityStatus.js
@@ -10,18 +10,30 @@ import WarningNotificationIcon from '../Attachment/Assets/WarningNotificationIco
 
 const CONNECTIVITY_STATUS_DEBOUNCE = 400;
 
+function setTimeoutSync(fn, interval) {
+  if (interval > 0) {
+    return setTimeout(fn, interval);
+  }
+
+  fn();
+}
+
 const DebouncedConnectivityStatus = ({ interval, children: propsChildren }) => {
-  const [children, setChildren] = useState(propsChildren);
+  const [children, setChildren] = useState(() => propsChildren);
   const [since, setSince] = useState(Date.now());
 
-  useEffect(() => {
-    const timeout = setTimeout(() => {
-      setChildren(propsChildren);
-      setSince(Date.now());
-    }, Math.max(0, interval - Date.now() + since));
+  const intervalBeforeSwitch = Math.max(0, interval - Date.now() + since);
 
-    return () => clearTimeout(timeout);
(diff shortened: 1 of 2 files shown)


packages/component/src/SendBox/ConnectivityStatus.js
```diff
@@ -10,18 +10,30 @@ import WarningNotificationIcon from '../Attachment/Assets/WarningNotificationIco
...
+  if (interval > 0) {
+    return setTimeout(fn, interval);
+  }
+
+  fn();
+}
+
 const DebouncedConnectivityStatus = ({ interval, children: propsChildren }) => {
-  const [children, setChildren] = useState(propsChildren);
+  const [children, setChildren] = useState(() => propsChildren);
```
tdurnford: Can you explain the difference between `useState(() => propsChildren)` and `useState(propsChildren)`? I don't understand the difference.

corinagum: (I believe) the return prevents it from rendering every time - just when there's a change in value.

corinagum: because I definitely HAVEN'T made this mistake 5 billion times

compulim: Explained offline. 😉

PR #2487, opened 2019-10-23
Use SVG for bubble nub

compulim:
### Added

-  Added bubble nub and style options, by [@compulim](https://github.com/compulim), in PR [#2137](https://github.com/microsoft/BotFramework-WebChat/pull/2137) and PR [#2487](https://github.com/microsoft/BotFramework-WebChat/pull/2487)

## Description

Emulator is using CSS variable for bubble background.

Today, bubble nub is rendered using SVG in Data URI as background image. This approach will not work with CSS variables.

Tomorrow, we are changing bubble nub to be rendered using pure SVG, thus, it can be styled using CSS properly, e.g. `fill`, `stroke`, and `stroke-width`.

### Technical details

- We render bubble nub using SVG, save it as data URI, and then save it under `background-image` of a CSS
- In this approach, we are using a package called `color` to normalize the color
   - This is because some colors are not supported by SVG
   - I don't exactly remember which, but probably `#F00` is not supported, so we need to normalize it to `#FF0000`
   - The `color` package cannot read CSS variables, that's why it's breaking
- The new approach, will use CSS for styling, but we can no longer use SVG serialized in `background-image`
   - We style `fill`, `stroke`, and `strokeWidth`
- The PR will render the SVG as a `<svg>`, instead of CSS style `background-image: data:image/svg+xml;utf8,<svg xmlns="" ...>`

## Specific Changes

- Modify `src/Activity/Bubble.js` to use SVG to render the bubble nub
- Remove data URI-based SVG from `src/Styles/StyleSet/Bubble.js`
- Remove `color` package because it is no longer being used

---

-  [x] Testing Added


packages/component/src/Activity/Bubble.js
```diff
@@ -18,22 +20,71 @@ const ROOT_CSS = css({
...
+      className={classNames(
+        ROOT_CSS + '',
+        bubbleCSS + '',
+        { 'from-user': fromUser, webchat__bubble_has_nub: nub },
+        className + '' || ''
+      )}
+    >
+      <div className="webchat__bubble__content">{children}</div>
+      {nub &&
+        !!(fromUser ? styleOptions.bubbleFromUserNubSize : styleOptions.bubbleNubSize) &&
```
corinagum: change variable names to include `bubbleFromBotNubSize`, etc.

corinagum: simplify code a little / leave comments?

packages/component/src/Activity/Bubble.js
```diff
@@ -18,22 +20,71 @@ const ROOT_CSS = css({
...
+
+  return (
+    <svg
+      className="webchat__bubble__nub"
+      version="1.1"
+      viewBox={`0 0 ${nubSize} ${nubSize}`}
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <g transform={`${horizontalTransform} ${verticalTransform}`}>
+        <path d={`M${p1} L${p2} L${p3}`} />
```
corinagum: this is the 'triangle' of the nub?

packages/component/src/Activity/Bubble.js
```diff
@@ -18,22 +20,71 @@ const ROOT_CSS = css({
...
-    {nub && !!(fromUser ? styleSet.options.bubbleFromUserNubSize : styleSet.options.bubbleNubSize) && (
-      <div className="webchat__bubble__nub" />
-    )}
-  </div>
-);
+function acuteNubSVG(nubSize, strokeWidth, side, upSideDown = false) {
+  const halfNubSize = nubSize / 2;
+  const halfStrokeWidth = strokeWidth / 2;
+  const horizontalTransform =
+    side === 'bot' ? '' : `translate(${halfNubSize} 0) scale(-1 1) translate(${-halfNubSize} 0)`;
```
corinagum: comments to explain this?

packages/component/src/Styles/StyleSet/Bubble.js
```diff
@@ -113,12 +63,17 @@ export default function createBubbleStyle({
...
-        backgroundImage: `url("${svgToDataURI(botNubSVG).replace(/"/gu, "'")}")`,
         bottom: isPositive(bubbleNubOffset) ? undefined : -bubbleNubOffset,
         height: bubbleNubSize,
         left: bubbleBorderWidth - bubbleNubSize + paddingRegular,
         top: isPositive(bubbleNubOffset) ? bubbleNubOffset : undefined,
-        width: bubbleNubSize
+        width: bubbleNubSize,
+
+        '& > g > path': {
+          fill: bubbleBackground,
```
corinagum: `...FromBot...`?

packages/component/src/Styles/createStyleSet.js
```diff
@@ -68,7 +68,14 @@ export default function createStyleSet(options) {
   // Keep this list flat (no nested style) and serializable (no functions)
 
   // TODO: [P4] Deprecate this code after bump to v5
```
corinagum: file an issue?

packages/component/src/Activity/Bubble.js
compulim: Simplified the code and added comments.

For `bubbleFromBot*`, created a new issue #2489. This PR is targeting a fix for Emulator and bubble nub is already in the code for quite some time.

packages/component/src/Activity/Bubble.js
compulim: Yes.

packages/component/src/Activity/Bubble.js
compulim: Added comments about the horizontal and vertical mirroring.

packages/component/src/Styles/StyleSet/Bubble.js
compulim: Filed issue #2489 to track this.

packages/component/src/Styles/createStyleSet.js
compulim: Let's do it as a separate pass. We have lots of "deprecate this after MM-DD-YY" in our codebase.

PR #2496, opened 2019-10-25
Update md files

corinagum:
Fixes N/A

## Changelog Entry

 
N/A
## Description

- Change language of `Add` section of CHANGELOG.md
- lint md files

## Specific Changes

 

---

-  [ ] Testing Added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 -->
 
 <!-- ### Added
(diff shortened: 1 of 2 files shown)


tdurnford:
Do we want to use 'resolve #xxx', 'resolves #xxx', or 'resolved #xxx'?

corinagum:
@tdurnford I intentionally chose 'resolve' since it matches with 'fix', what do you think?

tdurnford:
@corinagum I like resolves.

corinagum:
@tdurnford okay so that would mean I'm fixing all the 'fix's to 'fixes', sound g?

tdurnford:
@corinagum I think it should be 'fixes' anyway.

PR #2501, opened 2019-10-28
Added more translations to Finnish localization

sk91swd:
## Description

DirectLine does not show Finnish text in the chat window. I have tried to tackle this with custom code, but the simplest solution would be that DirectLine supports some basic Finnish text in the chat window (e.g. 'sending', 'new messages', 'x minutes ago'). I took a copy of the en-US.js file and translated it into Finnish. Some fields were already translated, good job.

## Specific Changes

 packages\component\src\Localization\fi-FI.js
-Added translations


sk91swd:
Editing missing information.

corinagum:
@sk91swd thanks for the contribution! More than happy to take a look at your PR once it's ready.

PR #2530, opened 2019-10-31
Add 4.6.0 to embed page

compulim:
Including @p-nagpal and @hristp.

~Will ping both to see if the "next minor" strategy is correct or not.~


corinagum:
Waiting for confirmation from Service reps :)

Issue #2539, opened 2019-11-01
Move to Redux Hooks


compulim:
## Feature Request

- [ ] Expose customization hooks
- [ ] Add deprecation notes for Context/HOC-based access
- [ ] Add test for reference grammar ID
   - Mock DirectLineJS to return `referenceGrammarID`
   - Expect: Web Chat should pass `referenceGrammarID` to `webSpeechPonyfillFactory`

### Expose customization hooks

- General hooks
   - [ ] `useActivities`
   - [ ] `useAdaptiveCardsHostConfig`
   - [ ] `useAdaptiveCardsPackage`
   - [ ] `useAvatarForBot`
   - [ ] `useAvatarForUser`
   - [ ] `useConnectivityStatus`
   - [ ] `useDictateInterims`
   - [ ] `useDictateState`
   - [ ] `useDisabled`
   - [ ] `useEmitTypingIndicator`
   - [ ] `useFocusSendBox`
   - [ ] `useGrammars`
   - [ ] `useGroupTimestamp`
   - [ ] `useLanguage`
   - [ ] `useLastTypingAt`
   - [ ] `useLocalize`
   - [ ] `useMarkActivityAsSpoken`
   - [ ] `usePerformCardAction`
   - [ ] `usePostActivity`
   - [ ] `useReferenceGrammarID`
   - [ ] `useRenderActivity`
   - [ ] `useRenderAttachment`
   - [ ] `useRenderMarkdownAsHTML`
   - [ ] `useScrollToEnd`
   - [ ] `useSendBoxValue`
   - [ ] `useSendEvent`
   - [ ] `useSendFiles`
   - [ ] `useSendMessage`
   - [ ] `useSendMessageBack`
   - [ ] `useSendPostBack`
   - [ ] `useSendTypingIndicator`
   - [ ] `useShouldSpeakIncomingActivity`
   - [ ] `useStartDictate`
   - [ ] `useStopDictate`
   - [ ] `useStyleOptions`
   - [ ] `useStyleSet`
   - [ ] `useSubmitSendBox`
   - [ ] `useSuggestedActions`
   - [ ] `useTimeoutForSend`
   - [ ] `useUserID`
   - [ ] `useUsername`
   - [ ] `useVoiceSelector`
   - [ ] `useWebSpeechPonyfill`
- Component-specific hooks
   - [ ] `useMicrophoneButtonClick`
   - [ ] `useMicrophoneButtonDisabled`
[...]

PR #2540, opened 2019-11-01
Add React hooks for customization (part 1)

compulim:
> Related to #2539.

## Changelog Entry

-  Resolves [#2539](https://github.com/Microsoft/BotFramework-WebChat/issues/2539), added React hooks for customization, by [@compulim](https://github.com/compulim) and [@corinagum](https://github.com/corinagum), in the following PRs:
   -  PR [#2540](https://github.com/microsoft/BotFramework-WebChat/pull/2540): `useActivities`, `useReferenceGrammarID`, `useSendBoxDictationStarted`

## Description

Add `useActivities`, `useReferenceGrammarID` and `useSendBoxDictationStarted`.

## Specific Changes

- Test harness
   - Added `runHook` function
- Exporting `hooks`
- Added the following hooks:
   - `useActivities`
   - `useReferenceGrammarID`
   - `useSendBoxDictationStarted`
- Updated components to use forementioned hooks
- Renamed `Context` to `WebChatUIContext`
- Removed unnecessary HOC for `<ScreenReaderText>`
- Updated `WebChatReduxContext` to export custom Redux hooks (this is not publicly exporting)
- Updated sample 21 to use aforementioned hooks
---

-  [x] Testing Added

### DIFF
--- a/__tests__/setup/web/index.html
+++ b/__tests__/setup/web/index.html
@@ -50,6 +50,8 @@
     <script src="https://tdurnford.github.io/BotFramework-Offline-MockBot/index.js"></script>
     <script src="https://unpkg.com/event-target-shim@5.0.1/dist/event-target-shim.umd.js"></script>
     <script src="https://unpkg.com/simple-update-in"></script>
+    <script src="https://unpkg.com/react@16.8.6/umd/react.development.js"></script>
+    <script src="https://unpkg.com/react-dom@16.8.6/umd/react-dom.development.js"></script>
     <script src="/createProduceConsumeBroker.js"></script>
     <script src="/mockWebSpeech.js"></script>
   </head>
@@ -62,6 +64,23 @@
         });
 
         return obj;
+      }
+
+      async function runHook(name, args) {
+        return new Promise((resolve, reject) => {
+          window.WebChatTest.store.dispatch({
+            type: 'DIRECT_LINE/INCOMING_ACTIVITY',
+            payload: {
+              activity: {
+                from: { id: 'bot' },
+                id: Math.random().toString(36).substr(2, 10),
+                name: 'hook',
+                type: 'event',
+                value: { args, name, reject, resolve }
+              }
+            }
+          });
+        });
       }
 
       function loadScript(src) {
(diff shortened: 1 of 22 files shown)


__tests__/setup/web/index.html
```diff
@@ -138,7 +157,24 @@
           };
         });
 
+        const activityMiddleware = () => next => card => {
```
tdurnford: Why are you using the activity middleware instead of the store middleware?

packages/component/src/BasicSendBox.js
```diff
@@ -28,28 +29,46 @@ const DICTATION_INTERIMS_CSS = css({ flex: 10000 });
...
-      )}
-      <div>
-        {(webSpeechPonyfill || {}).SpeechRecognition ? (
-          <MicrophoneButton className={MICROPHONE_BUTTON_CSS + ''} />
+// TODO: [P3] We should consider exposing core/src/definitions and use it instead
+function activityIsSpeakingOrQueuedToSpeak({ channelData: { speak } = {} }) {
+  return !!speak;
+}
+
+function useSendBoxDictationStarted(dictateState) {
```
tdurnford: Why are you returning the boolean value in an an array? Seems unnecessary if you're just deconstructing a single value.

packages/component/src/hooks/useActivities.js
```diff
@@ -0,0 +1,5 @@
+import { useSelector } from '../WebChatReduxContext';
+
+export default function useActivities() {
```
tdurnford: Are you supposed to return values in an array when making custom hooks?

__tests__/setup/web/index.html
compulim: Because we can only run the hook while we are doing rendering.

packages/component/src/BasicSendBox.js
compulim: Do you mean this line?

```
const [activities] = useActivities();
```

`activities` is a property. This is for having a unified pattern for getter/setter pattern. Since we do not support setting properties, it looks this way.

For example, some properties support setting, e.g. suggested actions, it will looks like:

```
const [suggestedActions, setSuggestedActions] = useSuggestedActions();
```

packages/component/src/hooks/useActivities.js
compulim: Yap. This pattern is adopted from `useState()`, which return `[value: string, setter: (nextValue: string) => void]`.

For example, for settable properties, e.g. send box value, `useSendBox()` will return `['current text in send box', (nextValue) => void]`.

But for read-only properties, to maintain consistency on the signature, `useLanguage()` will return `['en-US']`. This also help feature detection, i.e. the developer can know if the property can be set or not by looking at the 2nd item of the array.

packages/component/src/BasicSendBox.js
tdurnford: Oooooh... that makes a lot of sense.....

Issue #2505, opened 2019-10-28
Reminder for React versioning and incompatibilities


compulim:
## Feature Request

1. Warning when using older version of React that is incompatible with Web Chat
1. Steps to investigate React Hooks rules violations
1. Show breaking change in npm install
1. Add error boundary to scope errors within Web Chat

[Enhancement]

tonyanziano:
[Here](https://github.com/zloirock/core-js/blob/master/packages/core-js/postinstall.js#L14) is the code responsible for creating the post-install banner for `corejs`.

And it is invoked via [node in the `postinstall` npm hook](https://github.com/zloirock/core-js/blob/master/packages/core-js/package.json#L51) in `package.json`.

PR #2584, opened 2019-11-11
Replace webpack-visualizer-plugin with webpack-stats-plugin

### LINKED ISSUE #2583: Playground is not loading because of dependencies problem with React
## Version

`4.6.0` (`master`)

## Describe the bug

This bug only affect users who "npm install" Web Chat with `devDependencies`, e.g. playground or any packages that use `npm link` or `lerna` to include Web Chat in their build.

When `botframework-webchat` is installed with `devDependencies`, it installed React 0.14.9 due to transient dependencies from `webpack-visualizer` plugin.

`botframework-webchat` do not include `devDependencies` of React 16.8.6. This is because of the "Rules of Hooks" to maintain a single instance of React. We do include React 16.8.6 at the root package, so both `playground` and `botframework-webchat-*` will share the same instance of React.

## To Reproduce

1. Clone the repository
2. Build
3. Run playground
4. Type `help`

## Expected behavior

A list of commands are shown.

Instead, console log say `thumbnailCard` is having issues with `React.useMemo`.

## Additional context

[Bug]

compulim:
Fixes #2583.

### Changed

- `bundle`: Webpack will now use `webpack-stats-plugin` instead of `webpack-visualizer-plugin`, by [@compulim](https://github.com/compulim) in PR [#2584](https://github.com/microsoft/BotFramework-WebChat/pull/2584)
   - This will fix [#2583](https://github.com/microsoft/BotFramework-WebChat/issues/2583) by not bringing in transient dependency of React
   - To view the bundle stats, browse to https://chrisbateman.github.io/webpack-visualizer/ and drop the file `/packages/bundle/dist/stats.json`

## Description

(Copied from #2583.)

> This bug only affect users who "npm install" Web Chat with `devDependencies`, e.g. playground or any packages that use `npm link` or `lerna` to include Web Chat in their build.
> 
> When `botframework-webchat` is installed with `devDependencies`, it installed React 0.14.9 due to transient dependencies from `webpack-visualizer-plugin`
> 
> `botframework-webchat` do not include `devDependencies` of React 16.8.6. This is because of the "Rules of Hooks" to maintain a single instance of React. We do include React 16.8.6 at the root package, so both `playground` and `botframework-webchat-*` will share the same instance of React.

## Specific Changes

- `bundle`: Replace `webpack-visualizer-plugin` plugin with `webpack-stats-plugin`

---

-  [x] ~Testing Added~
   - No tests are added for playground

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- `bundle`: Webpack will now use `webpack-stats-plugin` instead of `webpack-visualizer`
+- `bundle`: Webpack will now use `webpack-stats-plugin` instead of `webpack-visualizer-plugin`, by [@compulim](https://github.com/compulim) in PR [#2584](https://github.com/microsoft/BotFramework-WebChat/pull/2584)
    - This will fix [#2583](https://github.com/microsoft/BotFramework-WebChat/issues/2583) by not bringing in transient dependency of React
    - To view the bundle stats, browse to https://chrisbateman.github.io/webpack-visualizer/ and drop the file `/packages/bundle/dist/stats.json`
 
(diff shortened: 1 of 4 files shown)


compulim:
FYI, this should be caught earlier if we have #2505. We are prophets (or things just bite us back.)

Issue #2582, opened 2019-11-11
Test in Web Chat / Embed does not work on Internet Explorer 11


corinagum:
Extended from #2449, originally customer-reported.
## Screenshots

![image](https://user-images.githubusercontent.com/14900841/68607483-9074c080-0465-11ea-93a4-e735847368e9.png)

## Version
gemini on Azure Portal, in Internet Explorer 11

## Describe the bug
When loading the 'Test in Web Chat' panel in Azure Portal, Web Chat does not load and the console shows the error 'Promise is not defined

## To Reproduce

1. Open Azure Portal in Internet Explorer
2. Navigate to your (a) bot
3. Click on 'Test in Web Chat'
4. Open console
1. Note that Web Chat does not load and shows error in the console

## Expected behavior

Test in Web Chat should load without error

## Additional context

Extended from #2449, originally customer-reported.

[Bug]

corinagum:
This issue has been addressed with the Bot Framework Services team and will be available as soon as possible - hopefully by Monday 11/18. We will keep this issue open and update with further details as available.

pdemro:
@corinagum @a-b-r-o-w-n @hristp I have created a very simple sample project which reproduces the issue with IE11.  The only reason I am posting this is because the error I see in console is different from the OP

https://github.com/pdemro/botframework-webchat-ie11-bug

![image](https://user-images.githubusercontent.com/5431128/68694747-b7a3be80-0547-11ea-8104-044b3cb0b7ec.png)

corinagum:
Hi @pdemro, please see our getting started es5 sample to see details on getting Web Chat to function in IE11. Thank you!

PR #2586, opened 2019-11-11
Debump embed to 4.3

compulim:
## Description

We have a fix for IE11 related to #2582. But we don't want to roll out an updated version of Web Chat via embed over the holiday season. So we are debumping it back to 4.3.

We will still keep other 4.4 and 4.5 entries.

## Specific Changes

- `embed`: Modify `servicingPlan.json` and default everyone to `4.3`

---

-  [x] Testing Added
   - Will test manually


hristp:
Looks good!

PR #2644, opened 2019-11-29
Add React Hooks for customization (part 15)

compulim:
> Related to #2539

## Changelog Entry

-  Resolves [#2539](https://github.com/Microsoft/BotFramework-WebChat/issues/2539), added React hooks for customization, by [@compulim](https://github.com/compulim), in the following PRs:
   -  PR [#2644](https://github.com/microsoft/BotFramework-WebChat/pull/2644): Added `internal/useWebChatUIContext` for cleaner code

## Description

Cleaning up some code by adding a new `useWebChatUIContext`. This hook is for internal use only.

## Specific Changes

- Added `component/src/hooks/useWebChatUIContext.js`
   - Updated other hooks that call `useContext(WebChatUIContext)` to call `useWebChatUIContext` instead

---

-  [x] Testing Added
   - No tests added because it should be covered by other tests

### DIFF
--- a/packages/component/src/hooks/internal/useWebChatUIContext.js
+++ b/packages/component/src/hooks/internal/useWebChatUIContext.js
@@ -6,7 +6,7 @@ export default function useWebChatUIContext() {
   const context = useContext(WebChatUIContext);
 
   if (!context) {
-    throw new Error('This hook can only be used on component that is decendants of <Composer>');
+    throw new Error('This hook can only be used on a component that is a descendant of <Composer>');
   }
 
   return context;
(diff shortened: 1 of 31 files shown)


packages/component/src/hooks/internal/useWebChatUIContext.js
```diff
@@ -0,0 +1,13 @@
+import { useContext } from 'react';
+
+import WebChatUIContext from '../../WebChatUIContext';
+
+export default function useWebChatUIContext() {
+  const context = useContext(WebChatUIContext);
+
+  if (!context) {
+    throw new Error('This hook can only be used on component that is decendants of <Composer>');
```
corinagum:
```suggestion
    throw new Error('This hook can only be used on a component that is descendant of <Composer>');
```

PR #2710, opened 2019-12-11
Reduce wasted render for activities

### LINKED ISSUE #2709: Reduce wasted render on every incoming/outgoing activities
### Before

MicrosoftTeams-image

### After

MicrosoftTeams-image (1)

## Version

`master`

## Describe the bug

On every incoming activities, `<BasicTranscript>` is re-rendering previously rendered activities.

This is because every item in `activityElements` is being re-created.

## Steps to reproduce

1. Add a `console.log` to `<StackedLayout>`, to trace how many times this component is getting rendered
1. Browse to Web Chat
1. Open developer tools
1. Type "a", followed by "b" and "c"

## Expected behavior

There should be a fixed amount of render of `<StackedLayout>` for every incoming/outgoing activity.

Instead, it is showing an exponential growth. For example, with 3 message pairs, it is showing 12 renders instead of 4.

## Additional context

[Bug]

compulim:
> Fixes #2709.

### Fixed

-  Fixes [#2709](https://github.com/microsoft/BotFramework-WebChat/issues/2709). Reduce wasted render of activities by memoizing partial result of `<BasicTranscript>`, in PR [#2710](https://github.com/microsoft/BotFramework-WebChat/pull/2710)

## Description

Add a new `useMemoArrayMap` hook for. It works like `Array.map` but also memoizing partial result from the mapper. The resultant array is a new one, but items in it will be memoized as needed.

If the mapper function changed, all memoized result will be invalidated.

## Specific Changes

- Added `useMemoArrayMap.js` hook for memoizing partial result for `Array.map`.
- Updated `BasicTranscript.js` to use `useMemoArrayMap` for the first-pass of render.

---

-  [ ] Testing Added

### DIFF
--- a/packages/component/src/BasicTranscript.js
+++ b/packages/component/src/BasicTranscript.js
@@ -12,7 +12,6 @@ import useRenderActivity from './hooks/useRenderActivity';
 import useRenderAttachment from './hooks/useRenderAttachment';
 import useStyleOptions from './hooks/useStyleOptions';
 import useStyleSet from './hooks/useStyleSet';
-import useTrackChanges from './hooks/internal/useTrackChanges';
 import useMemoArrayMap from './hooks/internal/useMemoArrayMap';
 
 const ROOT_CSS = css({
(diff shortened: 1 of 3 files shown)


packages/component/src/BasicTranscript.js
```diff
@@ -62,44 +64,48 @@ const BasicTranscript = ({ className }) => {
...
 
-        element && activityElements.push({ activity, element });
-
-        return activityElements;
-      }, []),
-    [activities, renderActivity]
-  );
+  const activityElements = useMemoArrayMap(activities, renderActivityElement);
 
+  // TODO: [P2] We can also use useMemoArrayMap() for this function.
```
corinagum: file issue

Issue #2722, opened 2019-12-12
The timestamp in the carousal layout should be moved out of the the film strip


tdurnford:
## Feature Request

**Is your feature request related to a problem? Please describe.**
Currently, the carousal layout's timestamp sits inside of the film strip, and, as a result, when the user scrolls through the attachments, the timestamp scrolls off the page. 

![scroll](https://user-images.githubusercontent.com/17039790/70744241-2a828f80-1cd6-11ea-95eb-690cf9f03087.gif)

**Describe the suggestion or request in detail**
We should consider moving the timestamp from the `CarouselFilmStrip` component to the `CarouselLayout` component to prevent the timestamp from moving.

![scroll2](https://user-images.githubusercontent.com/17039790/70744392-72a1b200-1cd6-11ea-9cd9-926421e568d9.gif)

**Additional Context**
Might resolve https://github.com/microsoft/BotFramework-WebChat/issues/2703

[Enhancement]

DesignPolice:
Got it! I will put this on the list, lots of things are wrapping up - so it will get done pretty soon.

PR #2735, opened 2019-12-14
Moved SendStatus into the Timestamp component

tdurnford:
### Change Log
-  Fixes [#2747](https://github.com/microsoft/BotFramework-WebChat/issues/2747). Moved `SendStatus` into the `Timestamp` component and removed the `sendStatus` style set, by [@tdurnford](https://github.com/tdurnford) in PR [#2735](https://github.com/microsoft/BotFramework-WebChat/pull/2735)

### Description
Moved the `SendStatus` component from the `StackedLayout` and `CarousalLayout` components to the `TimeStamp` component reducing redundant code across the layouts.

### Additional Context
Simplifies code changes for #2722

---

-  [ ] Testing Added


corinagum:
Hmm, changelog, just in case?

packages/component/src/Activity/SendStatus.js
```diff
@@ -31,10 +31,15 @@ const connectSendStatus = (...selectors) =>
...
 
 const SendStatus = ({ activity }) => {
-  const [{ sendStatus: sendStatusStyleSet }] = useStyleSet();
+  const [{ sendStatus: sendStatusStyleSet, timestamp: timestampStyleSet }] = useStyleSet();
   const focusSendBox = useFocusSendBox();
   const postActivity = usePostActivity();
 
+  sendStatusStyleSet &&
+    console.warn(
+      'Web Chat: styleSet.sendStatus is being deprecated. Please use styleSet.timestamp. This deprecation migration will be removed on or after December 17, 2021.'
```
corinagum: Hmm I'm not sure about the naming here. timestamp can be considered a send status, but send status is not a timestamp. Should we flip the variable names, or come up with a new container name?

packages/component/src/Activity/SendStatus.js
```diff
@@ -31,10 +31,15 @@ const connectSendStatus = (...selectors) =>
...
+    );
+
   // TODO: [P4] Currently, this is the only place which use a templated string
   //       We could refactor this into a general component if there are more templated strings
```
corinagum: What does this mean @compulim?

compulim: The templated string is referring to localization. "Send failed. {Retry}."

In order to get the hyperlink localized correctly, we localize it in 2-pass:

1. First, it localize `Retry` as `再送`
1. Then, it put the `再送` inside this `送信できませんでした。{Retry}。` and localize it again.

Template is referring to `{Retry}`.

We can easily write a better localizer that do nested localization, so it handle more strings than `{Retry}`. Or, we localize it using Markdown.

packages/component/src/Activity/SendStatus.js
compulim: I prefer flip too. Send status is broader than timestamp. 😉

packages/component/src/Activity/SendStatus.js
corinagum: Sounds like we may be contracting out localization help for Web Chat in the near future, so maybe we can take this discussion to next Applications standup. Thoughts?

packages/component/src/Activity/SendStatus.js
tdurnford: You guys are the worst

PR #2771, opened 2019-12-23
Added custom hooks to replace Timer component

### LINKED ISSUE #2313: Custom use hooks to replace <Timer>
## Feature request

Instead of using headless component, `<Timer at={Date.now() + 1000} onInterval={() => 0} />`. Create a new `useTimer` hook instead.

```js
import { useEffect } from 'react';

export default function useTimer(at, fn) {
  useEffect(() => {
    const timeout = setTimeout(fn, Math.max(0, at - Date.now()));

    return () => clearTimeout(timeout);
  }, [at, fn]);
}
```

Also, we could do the same for `RelativeTime`, create a `useIntervalSince` function for the `nextTimer()` replacement.

```js
import { useState } from 'react';

import useTimer from './useTimer';

function nextTimer(origin, interval) {
  const time = new Date(origin).getTime();
  const now = Date.now();

  return time > now ? time : now + interval - ((now - time) % interval);
}

export default function useIntervalSince(origin, interval, fn) {
  const [timer, setTimer] = useState(nextTimer(origin, interval));
  const handler = useCallback(() => {
    fn && fn();
    setTimer(nextTimer(origin, interval));
  }, [fn, origin, interval]);

  useTimer(timer, handler);
}
```

Also add tests, which use `lolex` to test against timestamp update.

## Additional context

Do this after our "convert to React functional component" PR is merged.

[Enhancement]

tdurnford:
Fixes #2313

## Changelog Entry
-  Added custom hooks - `useTimer` and `useIntervalSince` - to replace `Timer` component, by [@tdurnford](https://github.com/tdurnford), in PR [#2771](https://github.com/microsoft/BotFramework-WebChat/pull/2771)

## Description
- Added `useTimer` and `useIntervalSince` hooks
- Removed `Timer` component from the `RelativeTime` component

---

-  [X] Testing Added


packages/component/src/Utils/RelativeTime.js
```diff
@@ -1,40 +1,23 @@
 import PropTypes from 'prop-types';
-import React, { useCallback, useState } from 'react';
+import React from 'react';
 
+import useIntervalSince from '../hooks/useIntervalSince';
```
compulim: Sorting.

packages/component/src/Utils/RelativeTime.js
```diff
@@ -1,40 +1,23 @@
...
 const RelativeTime = ({ value }) => {
   const localizedAbsoluteTime = useLocalize('SentAt') + useLocalizeDate(value);
-
   const text = useLocalize('X minutes ago', value);
-  const [timer, setTimer] = useState(nextTimer(value));
 
-  const handleInterval = useCallback(() => {
-    setTimer(nextTimer(value));
-  }, [setTimer, value]);
+  useIntervalSince(value, TIMER_INTERVAL);
```
compulim: No callback function for `useIntervalSince`?

packages/component/src/hooks/useIntervalSince.js
```diff
@@ -0,0 +1,25 @@
...
+
+function nextTimer(origin, interval) {
+  const time = new Date(origin).getTime();
+  const now = Date.now();
+
+  return time > now ? time : now + interval - ((now - time) % interval);
+}
+
+export default function useIntervalSince(origin, interval, fn) {
+  const [timer, setTimer] = useState(nextTimer(origin, interval));
```
compulim: Can you use `useRef` here?

So if no `fn` is passed in, this will not causing any render.

packages/component/src/Utils/RelativeTime.js
tdurnford: We don't need one since `useIntervalSince` sets a new state forcing the component to rerender and relocalize the timestamp.

packages/component/src/hooks/useIntervalSince.js
tdurnford: If we use `useRef`, it won't trigger the `useEffect` hook in `useTimer` since the ref isn't updated until after the component is rerendered.

packages/component/src/Utils/RelativeTime.js
tdurnford: If we don't change the `useState` hook to `useRef` in `useIntervalSince`, then setting the `timer` would rerender the component; otherwise, we would have to add a `forceRender` state hook to create a callback and rerender the timestamp.

tdurnford:
@compulim ping

packages/component/src/Utils/RelativeTime.js
compulim: I am thinking about the hook design perspective, not about refactoring.

1. If we name it `useForceRenderAtInterval`, it works for me.
2. After starting the interval timer, how to stop it?

packages/component/src/hooks/useIntervalSince.js
compulim: I think this question is related to the one above. So I am resolving this and continue to discuss above.

PR #2830, opened 2020-01-18
Update daily build

compulim:
## Changelog Entry

(Build script and doc change, no changelog)

## Description

Daily build was not working because we recently  updated build scripts.

## Specific Changes

- All bot-required samples (under `sample/07.advanced-web-chat-apps`) should be linked to their corresponding hosting site
   - These samples has their own build pipeline, etc
- Exclude packages from bundled dependencies
   - Bundled dependencies is designed to help reducing `npm install` tarballs, but it doesn't work in our scenario
   - E.g. `botframework-webchat` bundle with `component` and `core` in the tarball. But it also bundle transient dependencies of `component` and `core`, that means the overall tarball size is > 100 MB and unnecessarily including transient dependencies
   - Personally, I think bundled dependencies are broken
- Fixed `01.g.hybrid-react-npm` was not hosted correctly
- Using lerna to build samples
- Updated `deploy_pages`
   - Reduce noise (we used to publish `.gitignore` and other unrelated files to GitHub Pages)
   - Fix `04.g.hybrid-react-npm`
- Updated Azure DevOps
   - Simplify build steps
   - Parallelizing daily build
   - Move "Push to GitHub Pages" from CI to CD
   - Publish daily build to GitHub Release (bundle and tarballs)

---

-  [x] ~Testing Added~
   -  No tests were added for doc and script change


corinagum:
Looks like this has merge conflicts :)

PR #2839, opened 2020-01-22
Fix call downstream middleware twice

### LINKED ISSUE #2838: Concat middleware failed to run next middleware twice
## Version

`master`

## Describe the bug

When writing a new (attachment) middleware that render the same downstream middleware twice, it failed.

For example, the following code using `concatMiddleware.js` will fail to return the correct value (`210`). It "skipped" the `next(value + 1)`, thus, returned `12` instead.

```js
const combine = () => next => value => {
  return next(value) + next(value + 1);
};

const oddMultiplyByTenAndEvenMultiplyByHundred = () => next => value => {
  if (value % 2) {
    return value * 10;
  }

  return value * 100;
};

const middleware = concatMiddleware(combine, oddMultiplyByTenAndEvenMultiplyByHundred);
const work = middleware()(value => value);

expect(work(1)).toEqual(210);
```

## Steps to reproduce

1. Write a new attachment middleware that call `next()` twice. So it will render attachment twice.
1. Open Web Chat
1. Type "echo Hello, World!"

## Expected behavior

It should render the attachment ("Hello, World!") twice.

Instead, the first time it render the attachment, it succeeded. The second render failed with "no renderer for this attachment" error.

## Additional context

IC3 team want to render a photo upload with the thumbnail and file name. So it could looks like the following screenshot.

image

Once the bug is fixed, the code below should produce the screenshot above.

```js
function createUploadMiddleware() {
  return () => next => ({
    activity = {},
    activity: { from: { role } } = {},
    attachment,
    attachment: { contentType, contentUrl, thumbnailUrl } = {}
  }) => {
    if (role === 'user' && /^image\//u.test(contentType) && thumbnailUrl) {
      const patchedAttachment = {
        ...attachment,
        contentType: 'application/octet-stream',
        thumbnailUrl: undefined
      };
      const patchedAttachments = activity.attachments.map(target =>
        target === attachment ? patchedAttachment : target
      );

      return (
        <React.Fragment>
          {next({ activity, attachment })}
          {next({ activity: { ...activity, attachments: patchedAttachments }, attachment: patchedAttachment })}
        </React.Fragment>
      );
    }

    return next({ activity, attachment });
  };
}
```

[Bug]

compulim:
> Fixes #2838.

### Fixed

-  Fixes [#2838](https://github.com/microsoft/BotFramework-WebChat/issues/2838). Fixed `concatMiddleware` should allow any middleware to call its downstream middleware twice, by [@compulim](https://github.com/compulim) in PR [#2839](https://github.com/microsoft/BotFramework-WebChat/pull/2839)

## Description

Allow a middleware to call downstream middleware twice. There are two call patterns:
- Handled by downstream middleware A first, return the result. Then on second call, middleware A skipped and fallback to middleware B, then return the result
- Handled by downstream middleware A twice, no fallback to other middleware

## Specific Changes

- Updated `concatMiddleware.js`
- Added unit tests for `concatMiddleware.js`
- Added integration test with previewing uploading
  - image

---

-  [x] Testing Added


packages/component/src/Middleware/concatMiddleware.spec.js
```diff
@@ -0,0 +1,71 @@
...
+  expect(work(1)).toEqual(10);
+  expect(work(2)).toEqual(200);
+});
+
+test('run 2 middleware for a single work', () => {
+  const combine = () => next => value => {
+    return next(value) + next(value + 1);
+  };
+
+  const oddMultiplyByTen = () => next => value => {
```
tdurnford: Can we reuse the `oddMultiplyByTen` and `evenMultiplyByHundred ` from the previous test? Seems redundant to define them twice.

compulim: Fixed!

PR #2846, opened 2020-01-22
Pass aria labels as props to StackedLayout

### LINKED ISSUE #2820: Pass in a prop to override screen reader text for chat messages i.e. "Bot Says..." -> "{name} Says..."
### The [Sample tag](https://github.com/microsoft/botframework-Webchat/issues?q=is%3Aissue+is%3Aopen+label%3ASample) also lists examples and implementations that have already been requested.

We would like to have the option to customize the screen reader text for incoming messages. We see that the screen reader is hard coded to use "Bot says {message}", but we would like to be able to pass in a value to change it to a different value i.e. "Alice says {message}"

**Is your feature request related to a problem? Please describe.**

Currently, screen readers will always begin with "Bot says {message}", but accessibility-wise, we don't always want to begin with that text.

**Describe the suggestion or request in detail**

We want to pass in a name as an optional override to replace "Bot" in the screen reader text, when we initialize the WebChat component.

**Describe alternatives you have considered**

**Additional context**

[Enhancement]

iwasn:
Fixes #2820 

## Changelog Entry

 
Allows props to change aria labels to be passed down to StackedLayout component. We can then pass values through using `activityMiddleware`

## Description

 
We needed a way to change the default aria-label announced by React WebChat on our app, as it begins each message with either "Bot says..." or "User says...", and "Bot says..." doesn't make sense to our screen reader users in the context of our application.

## Specific Changes

 
Allows props to change aria labels to be passed down to StackedLayout component

---

-  [ ] Testing Added

### DIFF
--- a/packages/component/src/Activity/StackedLayout.js
+++ b/packages/component/src/Activity/StackedLayout.js
@@ -91,7 +91,8 @@ const StackedLayout = ({ activity, children, timestampClassName }) => {
     from: { role } = {},
     text,
     textFormat,
-    timestamp
+    timestamp,
+    ariaLabels = {}
   } = activity;
 
   const activityDisplayText = messageBackDisplayText || text;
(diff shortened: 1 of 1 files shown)


tdurnford:
@iwasn Thank you for your willingness to contribute to Web Chat! 

In your PR, it looks like you are pulling the aria-labels for `botRoleLabel` and `userRoleLabel ` from the `ariaLabels` property you added to the activities sent from the bot. Unfortunately, the `ariaLabels` property is not included in the [ativities type definition](https://github.com/microsoft/botbuilder-js/blob/5b8b7efef63d31dc32fe99050aa15e42378e832d/libraries/botframework-schema/src/activityInterfaces.ts#L3) and is not support across all of the BotFramework SDKs. Since the property isn't supported by the SDK and is not recognized by other channels, we cannot use that property to set the activities' aria-labels. The `ariaLabels` property would also have to be added to every message sent by the bot for this solution to work which is not ideal from a development standpoint.

The Web Chat Development Team is still discussing the best way to implement customizing the activity's aria-label. Most likely the solution will include adding `botRoleLabel` and `userRoleLabel ` to Web Chat's [`styleOptions`](https://github.com/microsoft/BotFramework-WebChat/blob/master/packages/component/src/Styles/defaultStyleOptions.js).

PR #2866, opened 2020-01-28
Added default onError prop to <Dictation />

tonyanziano:
## Changelog Entry

- Added default `onError` prop to the `Dictation` component, by [@tonyanziano](https://github.com/tonyanziano), in PR [#2866](https://github.com/microsoft/BotFramework-WebChat/pull/2866)

## Description

 Added a default `onError` prop to the `<Dictation />` component so that errors caught by the `<Dictation />` component will be surfaced in the console.

## Specific Changes

 Added a default `onError` prop to the `<Dictation />` component so that errors caught by the `<Dictation />` component will be surfaced in the console.

===

Let me know if didn't follow a pattern correctly. I'm not very familiar with hooks or the Web Chat code base.


tonyanziano:
After some investigation, I think that my changes did not introduce a new error that is breaking the tests, but instead surfaced a preexisting error that is now breaking the tests.

It's hard to tell for sure without the ability to set breakpoints within the component package during test runs.

However, all my changes did was make it so that the `<Dictation />` component calls `console.error(<exception>)` whenever there is an error with speech / dictation. The test that is breaking clicks the microphone button, and that makes sure that no errors were logged via `console.error()`.

Before my changes, this test never failed because it **was impossible** for the component to call `console.error()` with an undefined error handler. However, now that the handler is defined, an error will cause the component to log a console error which then fails the test. The interesting part is that the error is just an empty string.

If I change my handler to do nothing, or to log to `console.warn() / .info() / .log()` then the test passes.

tonyanziano:
@corinagum no worries, I didn't fix it until yesterday 😃

PR #2993, opened 2020-03-07
Fix observeOnce

### LINKED ISSUE #2989: `observeOnce` should use ES Observable call pattern
## Version

`4.8.0`

## Describe the bug

In [`core/src/sagas/effects/observeOnce.js`](https://github.com/microsoft/BotFramework-WebChat/blob/master/packages/core/src/sagas/effects/observeOnce.js), the call pattern for `Observable.subscribe` is using older format (RxJS) instead of ES Observable.

The new signature is being used in `observeEach.js`.

This is causing test running in Jest not able to subscribe to the value correct.

The correct one is:

```diff
- subscription = observable.subscribe(resolve, reject, resolve);
+ subscription = observable.subscribe({
+   complete: resolve,
+   error: reject,
+   next: resolve
+ })
```

## Steps to reproduce

1. Write a test using Jest to call `DIRECT_LINE/POST_ACTIVITY`, and expect `DIRECT_LINE/POST_ACTIVITY_FULFILLED`

## Expected behavior

The `POST_ACTIVITY_FULFILLED` action should be dispatched back.

Instead, since `observeOnce` is not working under Jest + Node.js and failed silently, the `DIRECT_LINE/POST_ACTIVITY_FULFILLED` is never dispatched. Instead, it dispatched `POST_ACTIVITY_REJECTED`.

## Additional context

Tagging @youyu16.

[Bug]

compulim:
> Fixes #2989.

### Fixed

-  Fixes [#2989](https://github.com/microsoft/BotFramework-WebChat/issues/2989). Fix `observeOnce` to use ES Observable call pattern, by [@compulim](https://github.com/compulim) in PR [#2993](https://github.com/microsoft/BotFramework-WebChat/pull/2993)

## Description

`observeOnce` was using an older signature which is only used in RxJS but not ES Observable. This cause tests in DirectLineSpeech to fail.

## Specific Changes

- Update `observeOnce` to use the newer call signature

---

-  [x] Testing Added
   -  Updated tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 -  Resolves [#2897](https://github.com/microsoft/BotFramework-WebChat/issues/2897). Moved from JUnit to VSTest reporter with file attachments, by [@compulim](https://github.com/compulim) in PR [#2990](https://github.com/microsoft/BotFramework-WebChat/pull/2990)
 
+### Fixed
+
+-  Fixes [#2989](https://github.com/microsoft/BotFramework-WebChat/issues/2989). Fix `observeOnce` to use ES Observable call pattern, by [@compulim](https://github.com/compulim) in PR [#XXX](https://github.com/microsoft/BotFramework-WebChat/pull/XXX)
+
 ### Changed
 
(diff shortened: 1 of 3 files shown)


corinagum:
Looks like this one has some failing tests

PR #3031, opened 2020-03-18
Fix userID parameter

p-nagpal:
## Description
Fix typo in userID parameter


compulim:
@p-nagpal I am fixing the broken test.

PR #3068, opened 2020-04-02
Update README.md

v-kydela:
Fix broken links


v-kydela:
It turns out there are broken links in all sample readmes so I'll do another pull request

PR #3095, opened 2020-04-09
Keep props.locale when sending to bot

### LINKED ISSUE #3074: Specific Locale passed into webchat is not being sent to Bot service. (HealthBot)
## Version
https://cdn.botframework.com/botframework-webchat/latest/webchat-es5.gzip.js

>
>

## Describe the bug
Looks as though if I pass in a locale "es-US", it will send "es-ES" to the HealthBot service when posting activities. It seems the chatbot js only supports es-ES as a locale for it's strings and then therefore sending on this locale to the service. It'd be ideal that the locale passed in to the chatbot can translate to es-ES for the chatbot strings however respect the intention and forward on the locale that was passed in.

passing-in-locale

locale-sent-botsvc

## Steps to reproduce

1. Pass in locale "es-US" to store.dispatch({
                        type: 'DIRECT_LINE/POST_ACTIVITY',
                        meta: { method: 'keyboard' },
                        payload: {
                            activity: {
                                type: "invoke",
                                name: "InitConversation",
                                locale: "es-US" ...
2. Bot text is based on "es-ES" locale. 

## Expected behavior
Would like to see that the culture passed in is respected to forward on to botsvc.  

## Additional context

This was found using the HealthBot service as we created localization files for "es-US".

[Bug]

compulim:
> Fixes #3074.

### Fixed

-  Fixes [#3074](https://github.com/microsoft/BotFramework-WebChat/issues/3074). Keep `props.locale` when sending to the bot, by [@compulim](https://github.com/compulim) in PR [#3095](https://github.com/microsoft/BotFramework-WebChat/issue/3095)

## Description

If the dev set `props.locale` to `en-HK`, we should send this to the bot without modifications.

Today, since `en-HK` is not in the localization database, we normalize it to `en` and send it to the bot. We should only use the normalized form when we need it, e.g. date/unit formatting.

Normalizing means, we change it from an unknown string, into something closest in our database with our best effort.

## Specific Changes

- Modify `Composer` not to normalize the locale for `useLanguage` and Web Chat Core
- Modify `Composer` to do the normalization for:
   - Localized strings
   - Date/unit formatter (a.k.a. `globalize`)

---

-  [x] Testing Added


docs/HOOKS.md
```diff
@@ -394,7 +394,7 @@ useLanguage(options?: LanguageOptions): [string]
 
 This function will return the language of the UI. All UI components should honor this value.
 
-If no options are passed, the return value will be the written language.
+If no options are passed, the return value will be the written language. And this value should be same as `props.locale` passed to `<ReactWebChat>` or `<Composer>`.
```
corinagum, line 397:
```suggestion
If no options are passed, the return value will be the written language. This value should be the same as `props.locale` passed to `<ReactWebChat>` or `<Composer>`.
```

PR #3108, opened 2020-04-16
Jocupi/external link indication

### LINKED ISSUE #3087: Accessibility: The highlighted links below does not indicate whether they are internal or external links and whether they open in new window or not
image

**Actual**
In LHS screenshot.
- The first link is internal link and opens in a new window. There is no title attribute = “opens in a new window”. Also “opens in a new window” icon is not present.
- The second and third links are external links and opens in a new window. There is no title attribute = “opens in a new window, external”.  
 
**Expected**
Add title attribute as mentioned above and also use icons to represent “external links” and “links opening in new window” 
 
**WCAG Violation**
Predictable
https://www.w3.org/TR/WCAG20-TECHS/G201.html

JorgeCupiV:
Fixes #3087
## Description
- Add title for external and internal links
- Add icon for external links

## Specific Changes
- The **allowedAttributes** in the **SANITIZE_HTML_OPTIONS** at the **packages/bundle/src/renderMarkdown.js** file now also take the new *role* and *class* attributes
- A new function to verify if a link is external is added at the **packages/bundle/src/renderMarkdown.js** file
- A new style is added to add an icon for external links and is located at the **packages/component/src/Styles/StyleSet/TextContent.js** file

### DIFF
--- a/packages/bundle/src/renderMarkdown.js
+++ b/packages/bundle/src/renderMarkdown.js
@@ -7,7 +7,7 @@ import sanitizeHTML from 'sanitize-html';
 
 const SANITIZE_HTML_OPTIONS = {
   allowedAttributes: {
-    a: ['aria-label', 'href', 'name', 'target', 'title'],
+    a: ['aria-label', 'href', 'name', 'target', 'title', 'role', 'class'],
     img: ['alt', 'src']
   },
   allowedSchemes: ['data', 'http', 'https', 'ftp', 'mailto', 'sip', 'tel'],
@@ -48,6 +48,10 @@ const SANITIZE_HTML_OPTIONS = {
   ]
 };
 
+function isExternal(url) {
+  return url.host !== window.location.host;
+}
+
 const customMarkdownIt = new MarkdownIt({
   breaks: false,
   html: false,
@@ -65,6 +69,13 @@ const customMarkdownIt = new MarkdownIt({
       tokens[index].attrs[targetAttrIndex][1] = '_blank';
     } else {
       tokens[index].attrPush(['target', '_blank']);
+      tokens[index].attrPush(['class', 'externalLink']);
+    }
+
+    const urlAttrIndex = tokens[index].attrIndex('href');
+    if (isExternal(urlAttrIndex)) {
+      tokens[index].attrPush(['title', 'Opens in a new window, external']);
+    } else {
       tokens[index].attrPush(['title', 'Opens in a new window']);
     }
 
--- a/packages/bundle/src/renderMarkdown.js
+++ b/packages/bundle/src/renderMarkdown.js
@@ -4,6 +4,9 @@ import iterator from 'markdown-it-for-inline';
 import MarkdownIt from 'markdown-it';
 import markdownItAttrs from 'markdown-it-attrs-es5';
 import sanitizeHTML from 'sanitize-html';
+import { hooks } from 'botframework-webchat-component';
+
+const { useLocalizer } = hooks;
 
 const SANITIZE_HTML_OPTIONS = {
   allowedAttributes: {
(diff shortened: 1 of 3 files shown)


compulim:
Please add test case.

packages/bundle/src/renderMarkdown.js
```diff
@@ -7,7 +7,7 @@ import sanitizeHTML from 'sanitize-html';
 
 const SANITIZE_HTML_OPTIONS = {
   allowedAttributes: {
-    a: ['aria-label', 'href', 'name', 'target', 'title'],
+    a: ['aria-label', 'href', 'name', 'target', 'title', 'role', 'class'],
```
ericmaino, line 10: I don't think that 'class' belongs in this list. As my understanding is that these are supported markdown attributes and we don't want the markdown author to support overriding this behavior. 

It also looks like role was added as part of this change, but not necessarily used. It might be left over from another change you inherited.

packages/bundle/src/renderMarkdown.js
```diff
@@ -65,6 +69,14 @@ const customMarkdownIt = new MarkdownIt({
       tokens[index].attrs[targetAttrIndex][1] = '_blank';
     } else {
       tokens[index].attrPush(['target', '_blank']);
+      tokens[index].attrPush(['class', 'externalLink']);
+    }
+
+    const urlAttrIndex = tokens[index].attrIndex('href');
+    if (isExternal(urlAttrIndex)) {
```
ericmaino, line 76: These should be conditionally added since the markdown author could supply their own title

packages/component/src/Styles/StyleSet/TextContent.js
```diff
@@ -24,6 +24,15 @@ export default function createTextContentStyle({ bubbleMaxWidth, bubbleMinHeight
...
         overflow: 'hidden'
       }
+    },
+    '& .externalLink': {
+      margin: 0,
+      padding: 0,
+      backgroundPosition: 'center right',
+      backgroundRepeat: 'no-repeat',
+      backgroundImage:
+        'linear-gradient(transparent,transparent),url(https://en.wikipedia.org/w/skins/Vector/resources/skins.vector.styles/images/external-link-ltr-icon.svg?b4b84)',
```
ericmaino, line 34: There is likely a media process that must be followed. Legally the image must be able to be used in open source. Images should not be linked to other sources as this causes a many issues.

It look like the image should follow this [similar pattern](https://github.com/microsoft/BotFramework-WebChat/blob/master/packages/component/src/Styles/StyleSet/TypingAnimation.js) and should also be embedded using base64 encoding.

packages/bundle/src/renderMarkdown.js
JorgeCupiV: Can you elaborate on the condition part? In the commit 6039f009617bdc366c1085f9938dce433b977236 I'm adding localized strings as also requested by @compulim

packages/component/src/Styles/StyleSet/TextContent.js
corinagum: Please check and use the Fabric UI, which is the source of all of our current icons.

JorgeCupiV: Hi @corinagum, can you provide me with an example on how to properly use the Fabric UI? I can't find something similar on the webchat code, other than the example Eric is providing with the base64 images

JorgeCupiV: For example, when I see how the Microphone or Attachment icons are being used, I see that they are SVGs:
```html
<svg height="28" viewBox="0 0 25.75 46" width="28"><path clip-rule="evenodd" d="M20.75 11.75v21.37a7.69 7.69 0 0 1-.62 3.07 7.95 7.95 0 0 1-4.19 4.19 7.89 7.89 0 0 1-6.13 0 7.95 7.95 0 0 1-4.19-4.19 7.69 7.69 0 0 1-.62-3.07v-22.5a5.27 5.27 0 0 1 .45-2.17 5.69 5.69 0 0 1 3-3 5.48 5.48 0 0 1 4.35 0 5.69 5.69 0 0 1 3 3 5.27 5.27 0 0 1 .45 2.17v22.5a3.41 3.41 0 0 1-.26 1.32 3.31 3.31 0 0 1-1.8 1.8 3.46 3.46 0 0 1-2.63 0 3.31 3.31 0 0 1-1.8-1.8 3.41 3.41 0 0 1-.26-1.32V14h2.25v19.12a1.13 1.13 0 1 0 2.25 0v-22.5a3.4 3.4 0 0 0-.26-1.31 3.31 3.31 0 0 0-1.8-1.8 3.46 3.46 0 0 0-2.63 0 3.31 3.31 0 0 0-1.8 1.8 3.4 3.4 0 0 0-.26 1.31v22.5a5.32 5.32 0 0 0 .45 2.18 5.69 5.69 0 0 0 3 3 5.48 5.48 0 0 0 4.35 0 5.69 5.69 0 0 0 3-3 5.32 5.32 0 0 0 .45-2.18v-21.37z"></path></svg>
```

JorgeCupiV: Commit 4dd74342123906411fa3494df2ef6368ccbb3a8d resolves this as @compulim has provided me an approved Office UI Fabric icon in svg format.
Said svg icon is encoded in base64 and used

corinagum:
Is there an update for this PR? There are a number of tree changes to Web Chat incoming, so my recommendation is to close this PR and open a new one. Please be sure to address compulim's comments above before filing the new PR.

PR #3235, opened 2020-06-10
Add properties for stacked suggested actions height and overflow

### LINKED ISSUE #3182: Manually set Suggested Actions container height and manage overflow
Hi, 
 I'm using these style for suggested actions. When the number of suggested actions are more, i want the suggested actions to be in a container with fixed height and overflow hidden with scrollable. 
It looks like there are no style options to do the currently. Could you please check if this can done ?
const styleOptions = {
      suggestedActionBackground: 'White',
      suggestedActionBorder: 0,
      suggestedActionBorderRadius: 4,
      suggestedActionBorderStyle: 'solid',
      suggestedActionBorderWidth: 1,
      suggestedActionHeight: 32,
      suggestedActionLayout: 'stacked', // either "carousel" or "stacked"
    };

image

corinagum:
Fixes #3182

## Changelog Entry

-  Resolves [#3182](https://github.com/microsoft/BotFramework-WebChat/issues/3182). Added stacked suggested actions height properties, by [@corinagum](https://github.com/corinagum), in PR [#3235](https://github.com/microsoft/BotFramework-WebChat/pull/3235)

## Description

This gives the bot owner the ability to force the height of the suggested actions container when in stacked mode. Previously, height would always be calculated to 'auto'. Now, the height can be specified and overflow set (to hidden, scroll, etc)

## Specific Changes

`suggestedActionsStackedHeight`
`suggestedActionsStackedOverflow`

---

-  [x] Testing Added


packages/component/src/Styles/defaultStyleOptions.js
```diff
@@ -111,9 +111,13 @@ const DEFAULT_OPTIONS = {
   suggestedActionDisabledTextColor: undefined, // defaults to subtle
   suggestedActionHeight: 40,
   suggestedActionImageHeight: 20,
-  suggestedActionLayout: 'carousel', // either "carousel" or "stacked"
+  suggestedActionLayout: 'carousel', // either 'carousel' or 'stacked'
   suggestedActionTextColor: null,
 
+  // Suggeseted actions 'stacked' layout
+  suggestedActionsStackedHeight: undefined, // defaults to 'auto'
+  suggestedActionsStackedOverflow: undefined,
```
tdurnford, line 119: I think since `suggestedActionsStackedOverflow` doesn't default to a value we want to have it set to null. Not sure exactly why we're using that convention, but we are. @compulim Thoughts?

corinagum: Ah I had forgotten, thanks

PR #3286, opened 2020-07-01
Deploy to webchat-playground2.azurewebsites.net

compulim:


.github/workflows/playground.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - .github/workflows/playground.yaml
+      - packages/**
+
+# env:
```
tdurnford, line 11: Remove comments

compulim: Done.

PR #3287, opened 2020-07-01
Fix post-4.9.1 accessibility issues

### LINKED ISSUE #3236: [Accessibility] Put `aria-live` region outside of transcript
## Background

Instead of putting the `aria-live` region into the transcript, we can create another `<ul>` container with the `aria-live="polite"` and it contains **only** the screen reader text.

Since the screen reader text could be "scanned" by the user later, we can mitigate it by removing the narrating element after 2 seconds. So they can't really interact with them. 2 seconds should be enough to "queue" the element into screen reader.

```html
<ul aria-live="polite">
  <li>Bot said, this is the first message. Sent at 12:34 PM.</li>
  <!-- Read by screen-reader only during aria-live. We can remove it after 2 seconds. -->
</ul>
<ul>
  <li>
    <div>This is the first message.</div>
    <!-- This is visualized, and latter, accessible through scan mode of screen reader. This one may have interactive buttons. -->
  </li>
</ul>
```

## Action items

- Create a separate container with `aria-live="polite"` and it only contains what we want the screen reader to narrate
   - We will clean up this container after 2 seconds, to prevent the user from interacting with non-interactive content in this container
- Leave the existing visualized transcript, as long as it does not have `aria-live="polite"`, it is fine. The user will interact with buttons in this tree
- Remove `role="log"`
   - It is okay to not having `role="log"` or `role="feed"` on our transcript, `role="list"` and `role="listitem"` should be sufficient
   - We could also use `aria-label="transcript"` for "role renaming"

## 2020-06-26 Update

Making a new branch with multiple fixes, https://github.com/compulim/BotFramework-WebChat/tree/fix-a11y-post-4.9.1.

- Live region is now in a separated-but-invisible DOM tree
   - New message arrived, will show up in this DOM tree for 100 ms, screen reader will queue it
   - After 100 ms, the message will be removed from this DOM tree to workaround bugs in browsers
   - The message in live region will not announce the attachment content, it will only say
         - "Bot said: Hi I am Clara. 1 attachment. Sent at June 25 12:34 PM"
         - This is a technical barrier: as we have 2 DOM trees for the same transcript, we prefer not to render attachments twice
- Messages should be read in correct order
   - Due to distributed system nature in the backend, messages are arriving randomly, we sort the DOM tree by timestamp
      - For users who can see, messages are in correct order
      - For users who are using screen reader, messages appears to be out of order, because screen reader is first-come-first-serve
   - Since most bot message comes with a "reply to ID", I am using this as a hint to momentarily delay messages that their "reply to" was not arrived yet
- Cleaned up accessibility text
   - "User said" -> "You said"
   - "Bot/user sent" -> "Bot/user attached"
   - The narration for live region is "Bot said: Hi I am Clara. 1 attachment. Sent at June 25 12:34 PM."
   - The narration for transcript (e.g. scan/browse mode) is "Bot said: Hi I am Clara. Bot attached: Yes button, No button. Sent at June 25 12:34 PM."
- Change widget disable logic from "disable on blur" to "set `tabindex="-1"`"
- When disabling an Adaptive Card, don't set `tabindex="-1"` on its root `<div>`
- Do not strip Markdown from text marked as plain text
- Links in Markdown can be navigated by screen readers (fixing bug #3176)
- Added some `aria-roledescription` to make the screen reader narration more friendly

compulim:
> Fixes #3236.

## Changelog Entry

-  Fixes [#3236](https://github.com/microsoft/BotFramework-WebChat/issues/3236), by [@compulim](https://github.com/compulim) in PR [#3287](https://github.com/microsoft/BotFramework-WebChat/pull/3287)
   -  Isolated screen reader only live region for incoming activities and added a new `<ScreenReaderActivity>` component
   -  Removed screen reader text for activities outside of live region, including `<CarouselFilmstrip>`, `<StackedLayout>`, `<TextContent>`, and `<Timestamp>`
   -  Updated some accessibility texts
   -  Rectified activities render order by delaying activities with `replyToId` that reference an activity which is not ACK-ed, for up to 5 seconds
   -  Disabled widgets will have `tabindex="-1"` set, instead of `disabled` attribute
   -  Remove `tabindex="-1"` from Adaptive Cards container
   -  Hide activities of type `invoke`

## Description

Few areas of improvements:

- Transcript now use an isolated screen reader only live region
- Rectifying message order by using `replyToId` property as a hint
- Use `tabindex="-1"` for disabling button and prevent it from accessible through <kbd>TAB</kbd> key
- Remove unnecessary `tabindex="-1"` on the `<div>`, which is the Adaptive Card container
- Updated localization strings

For details, please refer to [specific changes](#specific-changes) section.

## Specific Changes

- Delaying activities with `replyToId` referencing an activity that is not received yet, the delay is up to 5 seconds (non-configurable)
   - This is to make sure activities with a reference, will be rendered only on or after the referencing activity is rendered
   - 5 seconds are chosen because both activities should be adjacent to each other in the message stream with only tiny delay (next packet), 5 seconds should be more than enough
   - Activities without `replyToId` reference, are not affected
- Disabled button will have `tabIndex` set to `-1` to prevent it from being focused on via the <kbd>TAB</kbd> key
   - Previously: when the button is blur (being out of focus), the `onblur` event handler will add `disabled` attribute to the element to disable the <kbd>TAB</kbd> key
- Do not add unnecessary `tabindex="-1"` to Adaptive Card container, which would cause them to be focusable on mouse click
- A new isolated screen reader only live region to workaround bugs in some browsers
   - The live region hold a copy of the activity, only seen by the screen reader
   - After the activity is added to this live region to be narrated, it will be removed within 100ms
   - Since the activity is already queued in the screen reader, removing it has no impact on the narration
   - Having the copy removed, will prevent end-user from navigating into the this region and get confused with two version of same activity in the DOM tree
- Updated localization strings
   - "Bot sent" -> "Bot attached"
   - "User said/sent" -> "You said/sent"
- Add a GitHub Actions workflow `preview-branch.yml`
   - When a branch named `preview-*` is pushed to `microsoft/BotFramework-WebChat` fork, it will create a new release (tarball + bundle)
   - This release can be used for customers to test upcoming features, before it goes to our main dev branch (`master`)

---

-  [x] Testing Added


corinagum:
> Do not add unnecessary tabindex="-1" to Adaptive Card container, which would cause them to be focusable on mouse click

Can you clarify? `-1` would make the card **non**-focusable

__tests__/html/accessibility.delayActivity.withReplyToId.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * @jest-environment ./__tests__/html/__jest__/WebChatEnvironment.js
+ */
+
+describe('accessibility requirement', () => {
+  test('should delay activity with "replyToId"', () => runHTMLTest('accessibility.delayActivity.withReplyToId.html'));
```
corinagum, line 6: can you clarify meaning? it delays the activity until `replyToId` is assigned?

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
 
 Lastly, we style the "New messages" separator like a normal button, styled it to float on top of the transcript, and added `click` and `keypress` event listener for <kbd>ENTER</kbd> and <kbd>SPACEBAR</kbd> key for [activation](https://www.w3.org/TR/wai-aria-practices-1.1/#button).
 
+## UX: Message ordering
+
+### User story
+
+Azure Bot Services is a distributed system and message order is not guaranteed. Web Chat use insertion sort based on the timestamp to order messages.
+
+Messages with a latter timestamp may arrive sooner than messages with a former timestamp. Thus, messages with latter timestamp could appear on the screen first. Then, messages with a former timestamp will get inserted before it.
```
tdurnford, line 138:
```suggestion
Messages with a newer timestamp may arrive before messages with an older timestamp. Thus, messages with a newer timestamp could appear on the screen first. Then, messages with an older timestamp will get inserted before it.
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+
+Because the time between the insertion is usually very short (adjacent packet in a Web Socket connection), users may not see the insertion visually. But screen reader always read messages in the order they appear on the screen, regardless of their positions in the DOM tree. Thus, message order could be confusing to users who relies on screen reader.
```
tdurnford, line 140:
```suggestion
Because the time between the insertion is usually very short (adjacent packet in a Web Socket connection), users may not see the insertion visually. But the screen reader always reads the messages in the order they appear on the screen, regardless of their positions in the DOM tree. Thus, the message order could be confusing to users who rely on the screen reader.
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+
+![Direct Line sequence diagram](https://raw.githubusercontent.com/microsoft/BotFramework-WebChat/master/docs/media/direct-line-sequence-diagram.png)
+
+### Implementations
+
+We will rectify message order using the `replyToId` property.
+
+`replyToId` is a property set by the Bot Framework SDK and it references the activity the bot is replying to. Web Chat use the `replyToId` property as a hint when rectifying message order.
```
tdurnford, line 148:
```suggestion
`replyToId` is a property set by the Bot Framework SDK and it references the activity the bot is replying to. Web Chat uses the `replyToId` property as a hint when rectifying the message order.
```

packages/component/src/Activity/StackedLayout.js
```diff
@@ -138,13 +128,16 @@ const StackedLayout = ({ activity, children, nextVisibleActivity }) => {
           'webchat__stackedLayout--hasAvatar': renderAvatar && !!(fromUser ? bubbleFromUserNubSize : bubbleNubSize)
         }
       )}
+      // role="article" // This will read "Landmark" in iOS Safari
```
corinagum, line 131: delete?

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+
+-  When a message with a `replyToId` property arrive, Web Chat will check if it received the activity with the specified ID:
```
tdurnford, line 150:
```suggestion
-  When a message with a `replyToId` property arrives, Web Chat will check if it received the activity with the specified ID:
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+   -  If no activities were received with the specified ID, Web Chat will wait up to 5 seconds for the referencing activity to arrive
+      -  If the activity arrive within 5 seconds, Web Chat will render the activity in the same render loop
+      -  If the activity did not arrive within 5 seconds, Web Chat will render the activity
+-  When a message without a `replyToId` property arrive, Web Chat will render the activity immediately
+
+## UX: Live region transcript
+
+### User story
+
+In a [live region](https://www.w3.org/TR/wai-aria-1.1/#live_region_roles), it is difficult to control which part to be read or excluded from the screen reader.
```
tdurnford, line 161:
```suggestion
In a [live region](https://www.w3.org/TR/wai-aria-1.1/#live_region_roles), it is difficult to control which part is read or excluded from the screen reader.
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+
+And sometimes, browser or screen reader could be buggy on how the live region is getting read. For example:
```
tdurnford, line 163:
```suggestion
Also, browsers and screen readers can be inconsistent on reading the live region. For example:
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+-  [Chromium bug #1067257](https://bugs.chromium.org/p/chromium/issues/detail?id=1067257)
+-  [Android TalkBack bug #157888765](https://issuetracker.google.com/issues/157888765)
+
+### Implementation
+
+To make the live region more consistent across browsers and easier to control, we separated the live region from the visible transcript:
+
+-  Two copies of transcript
+   -  Visible, rich, dynamic, and interactive transcript
+   -  Screen reader only transcript marked as live region
```
tdurnford, line 175: Think you're missing a work here. The screen reader only reads the?

```suggestion
   -  Screen reader only transcript marked as live region
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+      -  Attachment contents will not be narrated: attachments can be customized and the DOM tree could be very complex with interactive elements
```
tdurnford, line 176:
```suggestion
      -  Web Chat does not narrate attachment contents: attachments can be customized and the DOM tree could be very complex with interactive elements
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+-  The live region contains recently arrived activities
```
tdurnford, line 177:
```suggestion
-  The live region contains activities that were recently
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+   -  When the DOM element appear in the live region, screen reader will compute the alternative text and queue it for narration in a first-come-first-serve manner
```
tdurnford, line 178:
```suggestion
   -  When the DOM element appear in the live region, the screen reader will compute the alternative text and queue it for narration in a first-come-first-serve manner
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+   -  Screen reader will keep the alternative text in the queue even after the DOM element is removed from the live region
```
tdurnford, line 179:
```suggestion
   -  The screen reader will keep the alternative text in the queue even after the DOM element is removed from the live region
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+-  One second after the activity is rendered in the live region, Web Chat will remove it from the live region, this has a few benefits:
```
tdurnford, line 180:
```suggestion
-  One second after the activity is rendered in the live region, Web Chat will remove it from the live region. This has a few benefits:
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+   -  Workaround some browser and screen reader bugs that may keep repeating the entire transcript
+   -  Screen reader users will not be able to navigate into it and they will not notice there are 2 copies of the transcript
```
tdurnford, line 182: Replace 'it' with what it is referring to. The transcript? 

```suggestion
   -  The screen reader users will not be able to navigate into it and they will not notice there are 2 copies of the transcript
```

docs/ACCESSIBILITY.md
```diff
@@ -127,6 +129,61 @@ When the "New messages" separator is activated, it logically moves the separator
...
+   -  If the removal is too fast:
+      -  0-100 ms: Chrome and TalkBack on Android may miss some of the activities
+      -  One second is chosen after some experiments
```
tdurnford, line 185:
```suggestion
      -  The development team settled on using one second after some experimentation
```

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
```diff
@@ -133,7 +133,11 @@ function disableInputElementsWithUndo(element, observeSubtree = true) {
     disableElementWithUndo(element)
   );
 
-  undoStack.push(setAttributeWithUndo(element, 'tabindex', '-1'));
+  const tag = element.nodeName.toLowerCase();
+
+  // Only set tabindex="-1" on focusable element. Otherwise, we will made <div> focusable by mouse.
```
corinagum, line 138:
```suggestion
  // Only set tabindex="-1" on focusable element. Otherwise, we will make <div> focusable by mouse.
```

__tests__/clockSkew.js
```diff
@@ -101,7 +101,7 @@ describe('Clock skew', () => {
 
     await driver.wait(minNumActivitiesShown(3), timeouts.directLine);
 
-    const lastActivity = await driver.findElement(By.css('[role="list"] > li:last-child p'));
+    const lastActivity = await driver.findElement(By.css('.webchat__basic-transcript__activity:last-child p'));
```
tdurnford, line 104: can you use `getTracnscript` here instead?

packages/component/src/BasicTranscript.js
```diff
@@ -138,11 +147,21 @@ const BasicTranscriptContent = () => {
...
+              const key = (activity.channelData && activity.channelData.clientActivityID) || activity.id || index;
+
               result = [
                 {
                   activity,
                   element,
-                  key: (activity.channelData && activity.channelData.clientActivityID) || activity.id || index,
+                  key,
+
+                  // If this key change, the content of this attachment will be reannounced.
```
corinagum, line 163:
```suggestion
                  // If this key changes, the content of this attachment will be reannounced.
```

packages/component/src/Utils/Fade.js
```diff
@@ -0,0 +1,28 @@
+import { useLayoutEffect, useState } from 'react';
```
corinagum, line 1: Should add a short comment that this is for a11y/live region

compulim:
@corinagum Setting a `<div>` with `tabindex="-1"` will make it mouse-focusable, but not keyboard <kbd>TAB</kbd>-able.

Try this one, https://codepen.io/compulim/pen/xxZpXEX.

docs/ACCESSIBILITY.md
compulim: Thank you. This is always something I struggle when reading English: why sometimes people use former/latter? Is it only good for referencing things in the current article? E.g. "the former section told you ABC, the latter section told you XYZ."

__tests__/clockSkew.js
compulim: Quality of midnight coding... fixed!

__tests__/html/accessibility.delayActivity.withReplyToId.js
compulim: Fixed by renaming the test to "should delay activity with "replyToId" *referencing a missing activity*"

docs/ACCESSIBILITY.md
compulim: When I slack for a bit, my English turns really bad. 🤦🏻‍♂️

packages/component/src/Utils/Fade.js
compulim:
```js
// Initially, <Fade> will render all of its children.
// After a period of time (specified through the "fadeAfter" prop), <Fade> will stop rendering its children and removing it from DOM.
```

docs/ACCESSIBILITY.md
tdurnford: Yea, typically 'former' and 'latter' refer to something previously mentioned.

docs/ACCESSIBILITY.md
tdurnford: Hahaha it happens

Issue #3249, opened 2020-06-12
OC: Text-to-emoji (Jul 15)


compulim:
## Description

Convert `:)` into 😀 while typing (a.k.a. on-the-fly).

## Investigations

- Find out what common native/web apps on the market support this feature
   - Include 1-2 Microsoft products (no more than 40% of the scenarios we studied)
   - Include 3-5 products in total
- Document out how they behave
   - How to undo the automatic emoji conversion?
      - <kbd>CTRL</kbd> + <kbd>Z</kbd>
      - <kbd>BKSP</kbd>
   - How they works when copy-and-pasting `:)` into the box?
   - How they behave in gray areas (when emoji conversion should be disabled)?
      - Do they requires spaces around the emoji?
         - Type <kbd>A</kbd><kbd>B</kbd><kbd>C</kbd><kbd>:</kbd><kbd>)</kbd>, will they convert to "ABC😀", or does it requires a space between <kbd>C</kbd> and <kbd>:</kbd>?
   - What emoji text do they support?
      - E.g. do they support `:)` and `(:` at the same time?

## Goals

- Support <kbd>CTRL</kbd> + <kbd>Z</kbd> for undo
   - Try to use OS undo capability as much as possible, otherwise, we might conflict
   - Verify if other apps also support using <kbd>BKSP</kbd> key to undo
   - Test:
      1. Focus on the text box, type `:)`, it converted to 😀
      1. Blur the focus
      1. Focus on the text box again, press <kbd>CTRL</kbd> + <kbd>Z</kbd>, it should undo
- Need to make sure we don't convert unneeded `:)`
   - Verify behavior from other apps to see what scenario we should not convert
   - Type/paste `stream.forEach(System.out::println);`, it should not convert `::println` into `:😋rintln`
- Options
   - Disable this feature (either by a flag, or emptying out the conversion table)
   - Override the emoji conversion table
- We should only ship culture neutral emojis
- Make sure the work is componentized and can be carried forward when we implement #2211
- Make sure it works with both `<input>` and `<textarea>` (we have two controls today)
- Sample
   - E.g. type `spa`, it convert to ♨
      - The word is chosen instead of symbols, because we want to see how the feature behave in words, like `space`, `spatula` or `asparagus`

## Non-goals

- Emoji panel
   - OS already provided an emoji panel
- Any overlapping panel to select emojis
- Turning our text box into `contenteditable` box
- Only support OS-based Unicode emojis
[...]

PR #3303, opened 2020-07-13
Convert typed emoticons into Emoji in the TextBox

corinagum:
#3249 (Feature incomplete)

## Changelog Entry

-  Convert typed emoticons into emoji in the Textbox: [#3249](https://github.com/microsoft/BotFramework-WebChat/issues/3249), by [@corinagum](https://github.com/corinagum), in PR [#3303](https://github.com/microsoft/BotFramework-WebChat/pull/3303)

## Description

- [x] Adds the ability to convert various emoticon (e.g. `:)`) into emoji (😀) while typing on the fly.
- [x] Emoji autocorrect is disabled by default. Enable via  `styleOptions`
- [x] Developer may add custom emoji list. Enable via `styleOptions`
-  Accessibility:
    - [x] Emoji are announced by Narrator in Edge and Chrome after typing <kbd>Space</kbd>
    - [ ] ❌ Emoji are announced by NVDA and Chrome
   - This is an external bug. Issue filed with NVDA team https://github.com/nvaccess/nvda/issues/11372
- [ ] End-user may undo a typed emoji by using <kbd>Ctrl + Z</kbd>
    - This feature has not been completed

## Specific Changes

- By enabling `emojiAutocorrect` to `true` in `styleOptions`, emoticon typed ins SendBox will be converted to emoji on the fly
- `emojiList` in style options allows customers to assign custom map. 
  - The default list of Emoji uses guidance from [Microsoft Office behavior](https://support.microsoft.com/en-us/office/autocorrect-emojis-in-documents-and-email-601cad94-1a54-41e2-8cf8-4b09a6eadfe1)
  - Default map:
```js
  const emojiUnicodeMap = emojiList || {
    ':)': '😊',
    ':-)': '😊',
    '(:': '😊',
    '(-:': '😊',
    ':-|': '😐',
    ':|': '😐',
    ':-D': '😀',
    ':D': '😀',
    ':-p': '😛',
    ':p': '😛',
    ':-P': '😛',
    ':P': '😛',
    ':-o': '😲',
    ':o': '😲',
    ':O': '😲',
    ':-O': '😲',
    ':-0': '😲',
    ':0': '😲',
    ';-)': '😉',
    ';)': '😉',
    '<3': '❤️',
    '</3': '💔',
    '<\\3': '💔'
  };
```
- Default RegExp for detecting emoji: `/([:<()\\|\/3DPpoO0-]{2,3})/gim`
   - Custom RegExp will be determined via the custom emoji list in styles

### Example styleOptions to enable custom emoji: 
```js
const styleOptions = {
    emojiAutocorrect: true,
    emojiList: {
        ':sheep:': '🐑',
        '<3': '❤️
    }
 };

window.WebChat.renderWebChat(
  {
    directLine: window.WebChat.createDirectLine({ token }),
    styleOptions
  },
  document.getElementById('webchat')
 );
```

-  [x] Testing Added
   

### Emoji conversion behavior comparison by app:
|                       |🌟 Web Chat               | Teams            | Skype                     | Slack                                      | WhatsApp   | Messenger              |
|-----------------------|---------------------------|------------------|---------------------------|--------------------------------------------|------------|------------------------|
| Undo autocorrect      | N/A                       | Ctrl + Z (buggy) | No                        | Ctrl + Delete; Deletes emoji or shows text | Backspace  | No                     |
| Copy/Paste            | Yes                       | No               | Converts after send       | Yes                                        | No         | Converts after send    |
| Undo after blur       | N/A                       | Yes              | N/A (converts after send) | Deletes emoji                              | Yes        | No                     |
| `:)`                  | Yes                       | Yes              | Yes                       | converts to `:)` emoji                     | No (`:-)`) | Yes                    |
| `(:`                  | Yes                       | No               | No                        | Yes                                        | No         | converts to `:)` emoji |
| spaces required?      | No                        | No               | No                        | Yes                                        | No         | Yes                    |
| Disable feature       | Yes (disabled by default) | No               | No                        | No                                         | No         | No                     |
| Add list of own emoji | Yes                       | No               | No                        | Yes                                        | No         | No                     |

### Note on Assistive Technology behavior when Speak Typed Words is enabled:
| Announces | Chrome/NVDA | Edge/Narrator | Chrome/Narrator |                                         |
|-----------------|-------------|---------------|-----------------|-----------------------------------------|
| hi              | ❌           | ❌             | ❌               |                                         |
| hi[space]       | ✔           | ✔             | ✔               |                                         |
| :)              | ❌           | ❌             | ❌               |                                         |
| :)[space]       | ❌           | ✔             | ✔               |                                         |
| [C&P]hi         | ❌           | ❌             | ❌               |                                         |
| [C&P]hi[space]  | ❌           | ✔             | ✔               |                                         |
| [C&P]:)[space]  | ❌           | ✔             | ✔               |                                         |
| [Emoji panel]:) | ❌           | ❌             | ❌               | (all combos read on selection in panel) |

File issue: https://github.com/nvaccess/nvda/issues/11372


packages/component/src/SendBox/TextBox.js
```diff
@@ -124,7 +126,13 @@ const TextBox = ({ className }) => {
   const sendBoxString = localize('TEXT_INPUT_ALT');
   const typeYourMessageString = localize('TEXT_INPUT_PLACEHOLDER');
 
-  const handleChange = useCallback(({ target: { value } }) => setTextBoxValue(value), [setTextBoxValue]);
+  const handleChange = useCallback(
+    ({ target: { value } }) => {
+      value = setEmoji(value);
+      setTextBoxValue(value);
+    },
+    [setEmoji, setTextBoxValue]
```
beyackle, line 134: I don't understand using the setter functions here as dependencies of the callback. Are we expecting those functions to change at some point that would need us to recalculate the handleChange thing?

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,39 @@
...
+    ':-0': '😲',
+    ':0': '😲',
+    ';-)': '😉',
+    ';)': '😉',
+    '<3': '❤️',
+    '</3': '💔',
+    '<\\3': '💔'
+  };
+
+  const emojiRegExp = regex || new RegExp(/([:<()\\|\/3DPpoO0-]{2,3})/, 'gum');
```
beyackle, line 36: I'm not sure I like forcing a developer to write a regex to capture anything that's a key into an arbitrary object they're supplying. You might be able to get away with testing if a substring of the last (maximum length of a key) characters in the textbox is a key in the emojiUnicodeMap object, though I'm not sure how to write that efficiently offhand.

corinagum: I disagree. The regex is clean and precise, and the default regex provided by Web Chat is simple enough that adding or modifying the regex is accessible. It's always going to be up to the user to set up their customization (e.g. the object) correctly.

packages/component/src/SendBox/TextBox.js
corinagum: It's a 'good practice' that Web Chat uses throughout the repo

https://stackoverflow.com/questions/57180731/react-is-usestate-hooks-setter-able-change-during-a-component-life

packages/component/src/hooks/useEmojiFromStyles.js
compulim: My thought:
- There are two location for same piece of information, in two formats that can't be "naively-converted":
  1. Key of `emojiUnicodeMap`
  1. The "uber" `emojiRegExp`
- Dev can easily say A on (i), and then B on (ii)
   - Things suddenly doesn't work for them
   - To less prone to bugs, we could overcome this by, running all keys against `emojiRegExp` and tell them what need to be added
- Also depends on their `eomjiUnicodeMap`, it could be difficult for them to write a single `RegExp` to capture all of them
- Devs can type `::`. It will be captured in `emojiRegExp`, but the text is not converted to any emojis
   - Then, devs don't understand the value of `emojiRegExp`
   - Then, devs will end up writing `new RegExp(/.*/, 'gum')` and everything still seems to works
   - Then, devs ignore this option

corinagum: Okay, you convinced me. I'll revise this after 4.9.2 release.

corinagum:
@beyackle @compulim could you take another look? Thanks for the feedback.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+    ':-0': '😲',
+    ':0': '😲',
+    ';-)': '😉',
+    ';)': '😉',
+    '<3': '❤️',
+    '</3': '💔',
+    '<\\3': '💔'
+  };
+
+  const escapedString =
```
beyackle, line 40: I like this solution. Would it make more sense not to expose emojiRegExp at all and just do this logic inside useRenderEmoji instead? That way you wouldn't need to pass it between the two, it would just always be correct in the one place it needs to be.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "botframework-webchat-root",
-  "version": "4.9.2-0",
+  "version": "4.9.2",
```
compulim, line 3: Please wait until 4.9.2 is released on CDN. Then we do post-release 4.9.2.

This version number will become `4.9.3-0`.

packages/component/src/BasicTranscript.js
```diff
@@ -97,7 +97,7 @@ const BasicTranscript = ({ className }) => {
   const [{ activities: activitiesStyleSet, activity: activityStyleSet }] = useStyleSet();
   const [{ hideScrollToEndButton }] = useStyleOptions();
   const [activities] = useActivities();
-  const [animatingToEnd] = useAnimatingToEnd();
+  const [animatingToEnd] = useAnimating();
```
compulim, line 100: ?

packages/component/src/SendBox/TextBox.js
```diff
@@ -115,6 +116,7 @@ const PREVENT_DEFAULT_HANDLER = event => event.preventDefault();
 
 const TextBox = ({ className }) => {
   const [{ sendBoxTextWrap }] = useStyleOptions();
+  const [, setEmoji] = useRenderEmoji();
```
compulim, line 119: For a verb-based hook name, e.g. use*Render*Emoji. This should return a function, when called, to render emoji.

Can you change it to a noun-based hook name?

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+    customEmojiList &&
+    Object.keys(customEmojiList)
+      .map(escapeRegexp)
+      .join('|');
+
+  const emojiRegExp = customEmojiList
+    ? new RegExp(escapedString, 'gum')
+    : new RegExp(/([:<()\\|\/3DPpoO0-]{2,3})/, 'gum');
+
+  return [{ emojiAutocorrect, emojiUnicodeMap, emojiRegExp }];
```
compulim, line 50: This is returning an internal implementation of emoji. Please put this hook under `/hooks/internal`.

Internal implementation is because the developers doesn't need to know about `emojiRegExp`.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+    '<\\3': '💔'
+  };
+
+  const escapedString =
+    customEmojiList &&
+    Object.keys(customEmojiList)
+      .map(escapeRegexp)
+      .join('|');
+
+  const emojiRegExp = customEmojiList
```
compulim, line 46: `RegExp` is an object and is expensive. Please use `useMemo` to cache the result.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+import useStyleOptions from './useStyleOptions';
+
+function escapeRegexp(emoticon) {
+  return emoticon.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
+}
+
+export default function useEmojiFromStyles() {
+  const [{ emojiAutocorrect: autocorrect, emojiList: customEmojiList }] = useStyleOptions();
+
+  const emojiAutocorrect = autocorrect || false;
```
compulim, line 12: 1. Please do all the patching in `Composer`. By patching, I mean, in this line, you are patching `autocorrect` to become truthy/`false`
   - This is because other components may call `useStyleOptions` and then, get an unpatched value
   - Patching everything as close to the source as possible will reduce complexity. Source means, the props passed by the dev.
1. Instead of truthy/`false`, do you think it is more appropriate for `true`/`false`?

packages/component/src/Styles/defaultStyleOptions.js
```diff
@@ -187,7 +187,11 @@ const DEFAULT_OPTIONS = {
   toastSuccessBackgroundColor: '#DFF6DD',
   toastSuccessColor: '#107C10',
   toastWarnBackgroundColor: '#FFF4CE',
-  toastWarnColor: '#3B3A39'
+  toastWarnColor: '#3B3A39',
+
+  // Emoji
+  emojiAutocorrect: false,
```
compulim, line 193: Why do we need a boolean? The user could just make the `emojiList` falsy to turn off this feature, and this approach is more intuitive, because it reduce "2 ways to turn off emoji" into "a single way to turn off emoji."

packages/component/src/hooks/useRenderEmoji.js
```diff
@@ -0,0 +1,23 @@
+import useEmojiFromStyles from './useEmojiFromStyles';
```
compulim, line 1: Sort.

packages/component/src/hooks/useRenderEmoji.js
```diff
@@ -0,0 +1,23 @@
...
+import { useTextBoxValue } from '../SendBox/TextBox';
+import { useCallback } from 'react';
+
+export default function useRenderEmoji() {
+  const [{ emojiAutocorrect, emojiUnicodeMap, emojiRegExp }] = useEmojiFromStyles();
+
```
compulim, line 7: No need new line here.

New line is needed only for: the block after the new line is using variables/knowledges from the block before the new line.

packages/component/src/hooks/useRenderEmoji.js
```diff
@@ -0,0 +1,23 @@
...
+  const [value] = useTextBoxValue();
+
+  const setEmoji = useCallback(
+    value => {
+      const emoticon = value.match(emojiRegExp);
+
+      if (emojiAutocorrect && !!emoticon && !!emojiUnicodeMap[emoticon]) {
```
compulim, line 14: You don't need `!!` here. If-condition will handle truthy/falsy.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const emojiUnicodeMap = customEmojiList || {
```
compulim, line 14: Please put the default list in the style options.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+    ':o': '😲',
+    ':O': '😲',
+    ':-O': '😲',
+    ':-0': '😲',
+    ':0': '😲',
+    ';-)': '😉',
+    ';)': '😉',
+    '<3': '❤️',
+    '</3': '💔',
+    '<\\3': '💔'
```
compulim, line 37: Please add two tests.

```js
{
  ':o': '😲',
  ':o)': '🤡'
}
```

And

```js
{
  ':o)': '🤡',
  ':o': '😲'
}
```

FYI, key order is random and not guaranteed. It is up to the JS engine to decide which one come first, and the dev has no control over it.

If the dev want to have control, we need to use array.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+  return emoticon.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
+}
+
+export default function useEmojiFromStyles() {
+  const [{ emojiAutocorrect: autocorrect, emojiList: customEmojiList }] = useStyleOptions();
+
+  const emojiAutocorrect = autocorrect || false;
+
+  const emojiUnicodeMap = customEmojiList || {
+    ':)': '😊',
```
compulim, line 15: Is it expected not to have `:(` in the default list? The list from Office has it.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+    ';-)': '😉',
+    ';)': '😉',
+    '<3': '❤️',
+    '</3': '💔',
+    '<\\3': '💔'
+  };
+
+  const escapedString =
+    customEmojiList &&
+    Object.keys(customEmojiList)
```
compulim, line 42: I might be wrong, but looks like this line is missing indentation.

__tests__/html/emoji.enabledCustomMapDefaultRegExp.html
```diff
@@ -0,0 +1,40 @@
...
+                ':)': '😊'
+              }
+            }
+          },
+          document.getElementById('webchat')
+        );
+
+        await pageObjects.wait(conditions.uiConnected(), timeouts.directLine);
+
+        await pageObjects.typeInSendBox(':) (:  :-| :D :-p :-o <3 :(');
```
compulim, line 29: Two spaces between `(:` and `:-|`.

__tests__/html/emoji.enabled.html
```diff
@@ -0,0 +1,35 @@
...
+            directLine: window.WebChat.createDirectLine({ token: await token.fetchDirectLineToken() }),
+            store: createStore(),
+            styleOptions: { emojiAutocorrect: true }
+          },
+          document.getElementById('webchat')
+        );
+
+        await pageObjects.wait(conditions.uiConnected(), timeouts.directLine);
+
+        await pageObjects.typeInSendBox(':) (:  :-| :D :-p :-o <3 :(');
```
compulim, line 24: Two spaces between `(:` and `:-|`.

__tests__/html/emoji.disabled.html
```diff
@@ -0,0 +1,35 @@
...
+            directLine: window.WebChat.createDirectLine({ token: await token.fetchDirectLineToken() }),
+            store: createStore()
+            // styleOptions: { emojiAutocorrect: false } // (default)
+          },
+          document.getElementById('webchat')
+        );
+
+        await pageObjects.wait(conditions.uiConnected(), timeouts.directLine);
+
+        await pageObjects.typeInSendBox(':) (:  :-| :D :-p :-o <3 :(');
```
compulim, line 24: Two spaces between `(:` and `:-|`.

__tests__/html/emoji.disabled.html
```diff
@@ -0,0 +1,35 @@
...
+    <div id="webchat"></div>
+    <script type="text/babel" data-presets="env,stage-3,react">
+      const { conditions, createStore, expect, host, pageObjects, timeouts, token } = window.WebChatTest;
+
+      (async function() {
+        window.WebChat.renderWebChat(
+          {
+            directLine: window.WebChat.createDirectLine({ token: await token.fetchDirectLineToken() }),
+            store: createStore()
+            // styleOptions: { emojiAutocorrect: false } // (default)
```
compulim, line 17: Comment?

packages/component/src/hooks/useRenderEmoji.js
```diff
@@ -0,0 +1,23 @@
...
+        value = value.replace(emoticon, emojiUnicodeMap[emoticon]);
+      }
+      return value;
```
compulim, line 17: Always a new line before `return`.

packages/component/src/hooks/useRenderEmoji.js
```diff
@@ -0,0 +1,23 @@
...
+    },
+    [emojiAutocorrect, emojiUnicodeMap, emojiRegExp]
+  );
+
+  return [value, setEmoji];
+}
```
compulim, line 23: Please add a hook-specific test for this.

You can search `useXxx` for the set up for the hook-specific tests.

Also, update `HOOKS.md`.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
+/*eslint no-useless-escape: "off"*/
+/*eslint require-unicode-regexp: "off" */
```
compulim, line 2: Why these are off?

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const escapedString =
+    customEmojiList &&
+    Object.keys(customEmojiList)
+      .map(escapeRegexp)
+      .join('|');
+
+  const emojiRegExp = customEmojiList
+    ? new RegExp(escapedString, 'gum')
+    : new RegExp(/([:<()\\|\/3DPpoO0-]{2,3})/, 'gum');
```
compulim, line 48:
```suggestion
    : new RegExp(/([:<()\\|\/3DPpoO0-]{2,3})/gmu);
```

packages/component/src/SendBox/TextBox.js
```diff
@@ -12,6 +12,7 @@ import useFocus from '../hooks/useFocus';
 import useLocalizer from '../hooks/useLocalizer';
 import useScrollToEnd from '../hooks/useScrollToEnd';
 import useSendBoxValue from '../hooks/useSendBoxValue';
+import useRenderEmoji from '../hooks/useRenderEmoji';
```
compulim, line 15: Sort.

__tests__/html/emoji.enabled.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * @jest-environment ./__tests__/html/__jest__/WebChatEnvironment.js
+ */
+
+describe('Correct emoji should display if emojiAutocorrect has been set to true', () => {
+  test('', () => runHTMLTest('emoji.enabled.html'));
+  test('and a custom emojiUnicodeMap has been set', () => runHTMLTest('emoji.enabledCustomMapDefaultRegExp.html'));
```
compulim, line 7: Please put in a separate JS file.

packages/component/src/hooks/useRenderEmoji.js
```diff
@@ -0,0 +1,23 @@
+import useEmojiFromStyles from './useEmojiFromStyles';
+import { useTextBoxValue } from '../SendBox/TextBox';
+import { useCallback } from 'react';
+
+export default function useRenderEmoji() {
```
compulim, line 5: This hook, right now is:

- Read value from SendBox/TextBox
- Return the converted value

Few things:
1. It is not in a pair
   - Pair means, either
      1. Get from text box, set to text box
      1. Get from arg, return the value
1. If this is related to `SendBox/TextBox.js` only, put this hook under the `SendBox/TextBox.js`
   - Think about this: if the dev want to rewrite the text box, do we allow them to use this function?
   - Since you are grabbing the value via `useTextBoxValue` from `SendBox/TextBox.js`, this hook is "very specific to our implementation of TextBox"

I would recommend making this hook universal to any implementations of send box, including those written by developers.

packages/component/src/hooks/useEmojiFromStyles.js
```diff
@@ -0,0 +1,51 @@
+/*eslint no-useless-escape: "off"*/
```
compulim, line 1:
```suggestion
/*eslint no-useless-escape: "off" */
```

packages/component/src/hooks/useRenderEmoji.js
compulim, line 23: 1. Please add a sample on how to customize the emoji set.
1. Did you test pressing <kbd>CTRL</kbd> + <kbd>Z</kbd> after an emoticon is replaced with emoji? Can you add a test?
   - <kbd>CTRL</kbd> + <kbd>Z</kbd> may not work, but it should not break the UX.
1. Add a test for pasting text via <kbd>CTRL</kbd> + <kbd>V</kbd>. And set an expectation.
   - Expectations means: pasted text will have emoticons converted, or pasted text will not have emoticons converted
   - Please paste a text with 2+ emoticons

> FYI, some tests can only run under WebDriver but not local Chrome because <kbd>CTRL</kbd> + <kbd>V</kbd>/<kbd>Z</kbd> cannot be emulated using JS.

packages/component/src/BasicTranscript.js
corinagum: The method name was changed in react-scroll-to-bottom - I was getting errors until I updated them.

__tests__/html/emoji.disabled.html
corinagum: Intentional, to show what option is being tested.

packages/component/src/hooks/useEmojiFromStyles.js
corinagum: not sure how I missed this.

packages/component/src/hooks/useEmojiFromStyles.js
corinagum: Looks fine to me.

packages/component/src/hooks/useEmojiFromStyles.js
corinagum: `/` not needing an escape makes me nervous, I guess... removed the escape.

For the second one, I can't the linting error to go away, even when 'u' is added. So I added the comment. Am I missing something?
image
image

packages/component/src/hooks/useEmojiFromStyles.js
corinagum: do you mean `defaultStyleOptions`? Just making sure.

packages/component/src/Styles/defaultStyleOptions.js
corinagum: I wanted the boolean so that the user can enable or disable, and then they only need to provide an emoji list if they want to. This comment and:
image
seem to conflict. Could you clarify?

If `emojiList` is set to falsy, then user would always need to provide a list. Which isn't necessarily incorrect, but I thought it would be more important for the emoji feature to work 'out of the box' if enabled.

packages/component/src/BasicTranscript.js
compulim: There are two hooks in the `react-scroll-to-bottom`. One is `useAnimating` (which indicate if it is animating or not). The other one is `useAnimatingToEnd` (which indicate whether it is animating to the end or not).

image

packages/component/src/Styles/defaultStyleOptions.js
compulim: To disable emoji:
  
- Pass `false` to `emojiAutocorrect` or do nothing (default is `false`)
- Pass empty array to the emoji table

To enable emoji:
 
- Pass `true` to `emojiAutocorrect`
- And optionally, pass an array to the emoji table

Since there are multiple ways to do one thing, there can be conflicting options between them. What if:
 
- Pass an array to emoji table and did not set `emojiAutocorrect`?
- Pass an empty array to the emoji table and set `emojiAutocorrect` to `true`?

With two options and they are conflicting, there are no single source of truths.
 
Ultimately, we have 3 user stories to achieve:

- Disable emoji
- Enable emoji and use default mapping
- Enable emoji and use custom mapping
  - Customization can be:
    - Adding rules from default set
    - Removing rules from default set
    - Replacing whole set

(I am okay to leave the "add/remove/replace rules", since the default list is short and easily copyable.)
 
You can use a single options to reduce confusions.

- `emojiList: false` (disable emoji)
- `emojiList: true` (enable default emoji)
- `emojiList: { ':)': '' }`
  - This always replace the whole set.

packages/component/src/hooks/useEmojiFromStyles.js
compulim: Can you turn off the linting on a per-line basis, instead of per-file?

packages/component/src/hooks/useEmojiFromStyles.js
compulim: Retreating on this one. I think you might not able to put in `defaultStyleOptions.js`.

I think you might want to put it in `Composer.js`, i.e. patch the list as early as possible.

You can search `patchedXxx` in `Composer.js`.

packages/component/src/hooks/useEmojiFromStyles.js
corinagum: Tried this, and I still get the error. :(

PR #3367, opened 2020-07-30
Fix #3354 Adding access key to suggested actions

### LINKED ISSUE #3354: OC: Make suggested actions more accessible through TAB keys or accesskey attribute
## Feature Request

Today, when suggested actions pops up, the user would need to press <kbd>SHIFT</kbd> + <kbd>TAB</kbd> key to move to the (last) suggested actions and move their way up.

This is not intuitive and natural.

Tomorrow, we should make this more intuitive. But unfortunately, we cannot move the focus programmatically because it would interfere/interrupt the screen reader.

Few things we could consider

- Move suggested actions after the send box
   - This could change the UI
- Add `accesskey` attribute to the suggested actions container
   - Localization is not trivial as different browsers/platforms use different modifier keys

## Demo

recording (40)

### Access key modifiers

image

> Excerpt from https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey.

### Access key inconsistencies

- Depends on browsers, the element could be focused or activated (e.g. clicked)
- Depends on browsers, multiple elements with same `accesskey` values may have the first or last element triggered
   - As a component, we might conflict with the page host. We should only capture the access key when the user put focus on our UI
- Older browsers tends to deviate on modifier keys
   - Modern browsers generally converge to <kbd>ALT</kbd> + <kbd>SHIFT</kbd> on Windows and <kbd>CTRL</kbd> + <kbd>ALT</kbd> on Mac

> More read at https://webaim.org/techniques/keyboard/accesskey.

### `code` vs. `key` vs. `keyCode`

Naturally, `key` should be the better choice because:

1. It align with `accesskey` attribute
1. It is easy to understand what to fill in (probably `"a"` or `"A"`)

On Mac, when pressing <kbd>ALT</kbd> + <kbd>OPTION</kbd> + <kbd>A</kbd>, the `event.key` will return `"Å"` and `"å"` on `key` property. This is because, on Mac, <kbd>ALT</kbd> + <kbd>A</kbd> will be typed out as `"Å"` and `"å"`.

`event.code` do not take keyboard layout into account. For example, on a QWERTZ keyboard, press <kbd>Z</kbd> key will yield `KeyY` instead of `KeyZ`.

For `event.keyCode`, we cannot easily turn it into a localized string. And it is also deprecated by `event.code`.

So we end up need to use `key` and support multiple access keys, namely `"A a Å å"`.

[Enhancement]

compulim:
> Fixes #3354.

### Added

-  Resolves [#3354](https://github.com/microsoft/BotFramework-WebChat/issues/3354). Added access key (<kbd>ALT</kbd> + <kbd>SHIFT</kbd> + <kbd>A</kbd> for Windows and <kbd>CTRL</kbd> + <kbd>OPTION</kbd> + <kbd>A</kbd> for Mac) to suggested actions, by [@compulim](https://github.com/compulim), in PR [#3367](https://github.com/microsoft/BotFramework-WebChat/pull/3367)

## Description

Adding access key to suggested actions. Screen reader texts are updated to reflect this change.

Instead of using HTML's `accesskey` attribute, we are using our own implementation for a few reasons:

- Cross platform consistency
- Access key is only available when the focus is inside Web Chat component

## Specific Changes

- Added new sub-component `AccessKeySink` to handle the sinking of `keyup` events related to access keys
   - Include `useFocusAccessKeyEffect` hook for handling access key for focusing
- Added `suggestedActionsAccessKey` prop to `<Composer>`
- Added internal `useLocalizeAccessKey` hook to localize access key into string
- Added internal `useNavigatorPlatform` hook to identify platform
   - Different platform prefer different access key modifiers
- Updated `en-US.json` and `yue.json` for new loc strings

### `__tests__/suggestedActions.js`

One of the original snapshot for `__tests__/suggestedActions.js` is incorrect. It produce the following result when `styleOptions` is set to `{ suggestedActionLayout: 'stacked', suggestedActionsStackedHeight: 50 }`.

We should not render wrong layout with default style options. We should also put some efforts to make sure devs can't mess up the UI too far.

image

---

-  [x] Testing Added


packages/component/src/Composer.js
```diff
@@ -573,6 +576,7 @@ Composer.defaultProps = {
   sendTypingIndicator: false,
   styleOptions: {},
   styleSet: undefined,
+  suggestedActionsAccessKey: 'A a Å å',
```
tdurnford, line 579: Can you explain this for me?

packages/component/src/Utils/AccessKeySink/useFocusAccessKeyEffect.js
```diff
@@ -0,0 +1,29 @@
+import { useEffect } from 'react';
+
+import useContext from './internal/useContext';
```
tdurnford, line 3: Just for clarity 

```suggestion
import useAccessKeySinkContext from './internal/useContext';
```

packages/component/src/Utils/AccessKeySink/useFocusAccessKeyEffect.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default function useFocusAccessKeyEffect(key, ref) {
+  if (key && typeof key !== 'string') {
+    throw new Error('useFocusAccessKeyEffect: If defined, "key" must be of type "string".');
+  } else if (!ref || !('current' in ref)) {
+    throw new Error('useFocusAccessKeyEffect: "ref" must be defined and has "current" property.');
+  }
+
+  const context = useContext();
```
tdurnford, line 18:
```suggestion
  const context = useAccessKeySinkContext();
```

packages/component/src/Utils/AccessKeySink/internal/useContext.js
```diff
@@ -0,0 +1,7 @@
+import { useContext } from 'react';
+
+import Context from './Context';
```
tdurnford, line 3:
```suggestion
import AccessKeySinkContext from './Context';
```

packages/component/src/Utils/AccessKeySink/internal/useContext.js
```diff
@@ -0,0 +1,7 @@
...
+
+export default function useAccessKeySinkContext() {
+  return useContext(Context);
```
tdurnford, line 6: Clarity
```suggestion
  return useContext(AccessKeySinkContext);
```

packages/component/src/Utils/AccessKeySink/Surface.js
```diff
@@ -0,0 +1,56 @@
+import PropTypes from 'prop-types';
+import React, { useCallback, useRef } from 'react';
+
+import Context from './internal/Context';
```
tdurnford, line 4:
```suggestion
import AccessKeySinkContext from './internal/Context';
```

packages/component/src/Composer.js
compulim: Haha. I am also confused until I use Safari on Mac.

IIRC, when you press <kbd>ALT</kbd> + <kbd>SHIFT</kbd> + <kbd>A</kbd>, you will receive few things in the `HTMLEvent` object:

- `keyCode = 65` (deprecated in favor of `code`, I believe)
- `code = "KeyA"` (this don't honor keyboard layout. On a QWERTY keyboard, press <kbd>Y</kbd> -> `"KeyY"`. On a QWERTZ keyboard, press <kbd>Z</kbd> -> `"KeyY"` too)
- `key = "a"` (this honor keyboard layout)

So, `code` is useless. `keyCode`, hmmm, first, deprecated, second, I can't easily convert `65` -> `A`. I need `A` because I need to build the accessibility text: "Press ALT + SHIFT + A for suggested actions".

So, we down to `key`, which is a smaller nightmare.

Then, there are few de facto `accesskey` modifiers.

- Windows
   - <kbd>ALT</kbd> + <kbd>A</kbd>, `key = "a"`. We don't want to capture this one because some browsers ignore it.
   - <kbd>ALT</kbd> + <kbd>SHIFT</kbd> + <kbd>A</kbd>, `key = "a"` (or `"A"`, I don't remember)
- Mac
   - <kbd>CTRL</kbd> + <kbd>OPTION</kbd> + <kbd>A</kbd>, `key = "å"`
   - <kbd>CTRL</kbd> + <kbd>OPTION</kbd> + <kbd>SHIFT</kbd> + <kbd>A</kbd>, `key = "Å"`

On Mac, it's `å` but not an `a`, because <kbd>OPTION</kbd> + <kbd>A</kbd> will yield `å`. So, <kbd>CTRL</kbd> + <kbd>OPTION</kbd> + <kbd>A</kbd> will yield `å`.

HTML `accesskey` officially support 1+ keys, separated by spaces (because <kbd>SPACEBAR</kbd> is not allowed as an access key).

So, we are allowing 4 variants of A. And taking the first one for screen reader text.

I might be able to take out the first one. But since it doesn't hurt, and it is nice for screen reader text, I just leave it there.

(I learn new thing every day.)

PR #3405, opened 2020-08-12
Emojiundo

corinagum:
TBD


compulim:
> Fixes #3249.

## Changelog Entry

### Added

-  Resolves [#3249](https://github.com/microsoft/BotFramework-WebChat/issues/3249). Convert typed emoticons into Emoji, by [@corinagum](https://github.com/corinagum) and [@compulim](https://github.com/compulim), in PR [#3405](https://github.com/microsoft/BotFramework-WebChat/pull/3405)

### Samples

-  Added custom Emoji set sample, by [@corinagum](https://github.com/corinagum), in PR [#3405](https://github.com/microsoft/BotFramework-WebChat/pull/3405)

## Description

Added emoticon to Emoji in the send box.

## Design

### Style options

To enable set of Emojis (default), `{ styleOptions: { emojiSet: true } }`. To disable Emojis, `{ styleOptions: { emojiSet: false } }` (or `{}` instead of `false`).

For custom Emoji set:

```js
{
  styleOptions: {
    emojiSet: {
      '100': '💯'
    }
  }
}
```

### Hooks

It is not trivial for end-developer to re-build their own Emoji-enabled send box. This is because they must re-build their own undo stack. Otherwise, the "chemical reaction" between `string.replace` and React would mess up browser-based undo stack.

Thus, we are not exposing any hooks. However, they can access the `emojiSet` style options through the existing `useStyleOptions` hook, i.e. `useStyleOptions().emojiSet`.

We are not exposing the `useReplaceEmoticon.js` because the API signature requires 3 things: `value`, `selectionStart`, and `selectionEnd`.

The `selectionStart`/`selectionEnd` seems not trivial for the user to understand their need. Thus, if they are rebuilding the send box with Emoji, they might not understand how to pass these variables in and why they are needed.

FYI, they are needed because we need to know which character was changed. We are only replacing emoticons when the user finish the emoticon with its last character. I.e. user press <kbd>D</kbd> after the `:` (colon) letter.

We are not replacing emoticons for the whole string as that would make the UI very difficult to use (every key press will convert to Emoji).

Thus, the `useReplaceEmoticon.js` is internal for now.

## Specific Changes

- Modified `SendBox/TextBox.js`
   - Re-built undo stack
   - Add emoji conversion
- Added `internal/useReplaceEmoticon.js`


## Review Checklist

   - It is up to screen reader to read the Emoji
   - We are not bringing culture-specific emoticons (e.g. `:wink:`)

packages/component/src/SendBox/TextBox.js
```diff
@@ -97,34 +98,158 @@ function useTextBoxSubmit() {
...
+    rememberInputState();
+
+    placeCheckpointOnChangeRef.current = true;
+  }, [placeCheckpointOnChangeRef, rememberInputState]);
+
+  const handleKeyDown = useCallback(
+    event => {
+      const { ctrlKey, key, metaKey } = event;
+
+      if ((ctrlKey || metaKey) && (key === 'Z' || key === 'z')) {
```
a-b-r-o-w-n, line 237: Is `ctrl+Z` typical for undo?

compulim: <kbd>CTRL</kbd> + <kbd>Z</kbd> is de facto on Windows. And I guess <kbd>COMMAND</kbd> + <kbd>Z</kbd> is de facto on Mac.

You spotted the corner I cut here. 😁

Please let me know if I should only allow <kbd>COMMAND</kbd> + <kbd>Z</kbd> on Mac but not <kbd>CTRL</kbd> + <kbd>Z</kbd>.

compulim:
Thanks @a-b-r-o-w-n.

Agree, rather complex.

OS will push undo checkpoint on:

- Just before first change after focus
- Just before first change after cursor move
- Just before first change after cut/paste

But since HTML doesn't have cross-browser `beforechange`/`changing` event. So it looks complicated here. I need to keep remembering the current value just before the `change` event.

Better, I should "subclass" `<input>`/`<textarea>` and add a `beforechange`/`changing` event in the custom component. That could significantly simplify the code.

PR #3551, opened 2020-10-20
Add explicit role to stacked layout message

### LINKED ISSUE #3537: Accessibility issue : Ensure aria-roledescription is only used on elements with an implicit or explicit role
[edit by @corinagum 11/9/2020]

See latest [comment](https://github.com/microsoft/BotFramework-WebChat/issues/3537#issuecomment-724381338): I have reopened this bug due to 'layout double' and 'layout carousel' still showing as breaking accessibility in Accessibility Insights.

/end edit

---

Running a fast pass on WebChat bot, getting this error

Path | #webchat__stacked-layout__id--i5ta9

<div aria-roledescription="message" class="webchat__stacked-layout__message-row" id="webchat__stacked-layout__id--i5ta9">
**Fix the following:Give the element a role that supports aria-roledescription**

Similarly on all these:

Path | #webchat__stacked-layout__id--wpsrv
#webchat__stacked-layout__id--y6f93
#webchat__stacked-layout__id--vsynk

image

corinagum:
> Fixes #3537

## Changelog Entry
-  Fixes [#3537](https://github.com/microsoft/BotFramework-WebChat/issues/3537). [Accessibility]: Ensure `aria-roledescription` is only used on elements with implicit/explicit role based off of [WAI role attributes](https://www.w3.org/WAI/PF/HTML/wiki/RoleAttribute), by [@corinagum](https://github.com/corinagum) in PR [#3551](https://github.com/microsoft/BotFramework-WebChat/pull/3551)

## Description

- Accessibility Insights for Web showed messages as having no implicit/explicit role. This change adds explicit `role="group"` to messages.
- Although WAI-ARIA indicates that `role="group"` can have a child `role="listitem"`, Accessibility Insights files the pattern as an a11y bug. 

## Review Checklist


packages/component/src/Activity/StackedLayout.js
```diff
@@ -162,6 +162,7 @@ const StackedLayout = ({
               // Disable "Prop `id` is forbidden on DOM Nodes" rule because we are using the ID prop for accessibility.
               /* eslint-disable-next-line react/forbid-dom-props */
               id={ariaLabelId}
+              role="log"
```
compulim, line 165: First, `role="log"` implies `aria-live="polite"`. We shouldn't use this role.

Second, this is the message bubble, not a transcript.

compulim (changes requested):

packages/component/src/Activity/StackedLayout.js
corinagum: Then what do you suggest?

compulim: https://www.w3.org/WAI/PF/aria/roles

Let's pick one of the role here.

compulim (commented):
Please add test too.

packages/component/src/Activity/StackedLayout.js
```diff
@@ -152,7 +152,7 @@ const StackedLayout = ({
       })}
       role="group"
     >
-      <div className="webchat__stacked-layout__main">
+      <div className="webchat__stacked-layout__main" role="list">
```
compulim, line 155: "list" implies it's a container with multiple items to be narrated.

While it is true that activity could have multiple attachments, the text message itself is not part of attachments. Activity = text message + N x attachment.

So I think "listitem" is not a good role for text message as it is always single.

corinagum: Where does attachment come into this change? An activity has an ordered list of children, one of which is the message. There are various parts to the activity, but the message is the main child of that list.

cwhitten (approved):

PR #3608, opened 2020-11-17
Make 4.9.2 the default version for Hosted/Embed Webchat

p-nagpal:
## Description

Make 4.9.2 the default version for Hosted/Embed Webchat. This will allow us to reset the upgrade and restart rolling our 4.10.*


corinagum (approved):
LGTM

PR #3630, opened 2020-12-08
Bump adaptivecards@2.5.0

### LINKED ISSUE #3392: Bump Adaptive Cards to 2.*
## Feature Request

https://www.npmjs.com/package/adaptivecards

There are a few breaking changes we will need to modify code on our side to support them.

Adaptive Cards renderer 2.1 is using schema of **1.3**, not **2.x**.

### Ping service team

Service team may need to update their service code or dependencies to make sure their parser/stringifier support AC schema 1.3.

Today, when a bot send AC JSON to Direct Line channel, the channel parse, then restringify it. Then, they send it to Web Chat.

When we bump to AC 2.*, if the service is not ready for AC with schema 1.3, it may not be very meaningful (other than accessibility fixes).

[Enhancement]

corinagum:
> Fixes #3392

## Description
- Bump Adaptive Cards to use latest SDK version. Breaking changes list can be found at https://github.com/microsoft/AdaptiveCards/blob/f18d76e516dcb43b90f6587028f18747dd1bfbe2/source/nodejs/adaptivecards/README.md#breaking-changes

- With this bump, the following will be displayed when an Adaptive Card schema of 1.3 (or other) is sent: 
image
This is due to the service not yet supporting the 1.3 schema.
This message is from the Adaptive Cards sdk
- MockBot has two new cards: `agenda` and `wizard`
- All previously working Adaptive Cards have been verified with this bump
- `tests__/html/cardAction.adaptiveCard.cardAgendaTranscript.html` is an example of a 1.3 schema card working in Web Chat (not using DirectLine service)

## Tests image diffs updated

There are a number of tests that had to be updated with this bump. I have verified the behavior as either a **fix** of Adaptive Cards behavior,  **regression** (with corresponding issue), or trivial changes due to MockBot updates.

*'adaptive cards broken card of invalid version'*

image
This error message is from the Adaptive Cards sdk
This is a change in behavior to what should be expected. Broken card should show an error, not show an empty div.

*'useScrollToEnd calling scroll to end should scroll to end'* 

image

There are a number of scrollbar fixes (due to length of transcript)that are caused by adding new commands to MockBot. This is not a change in behavior.

'*clock skew - faking clock skew should affect activity order*'

image

Previously, images were not center-aligned even though MockBot's BingSports card indicates it. Fixed by Adaptive Cards bump.

*'adaptive cards disable card inputs'*

image
This is a behavior bug from Adaptive Cards, seen at https://github.com/microsoft/AdaptiveCards/pull/5134 This PR shows the change as fixed, and therefore should be fixed in the next AC release/bump. This will eventually mean updating the snapshots again. 

*'transcript activity grouping'*
These screenshots are additions, not changes. I'm assuming they are new tests that didn't have snapshots pushed up in a previous commit. *I am assuming the behavior in these screenshots is as expected*

-  ~[ ] I have updated documentation~

## Review Checklist


amal-khalaf (approved):

packages/bundle/src/adaptiveCards/Attachment/ReceiptCardContent.js
```diff
@@ -34,7 +34,7 @@ const ReceiptCardContent = ({ actionPerformedClassName, content, disabled }) =>
       builder.addTextBlock(title, { size: TextSize.Medium, weight: TextWeight.Bolder, wrap: richCardWrapTitle });
 
       if (facts) {
-        const [firstFactColumn, lastFactColumn] = builder.addColumnSet([75, 25]);
+        const [firstFactColumn, lastFactColumn] = builder.addColumnSet(['75', '25']);
```
cwhitten, line 37: this looks weird to me. the elements need to be strings that look like numbers?

corinagum: Yeah you're right. Looked at the code again and there's a way to provide the number instead of parsing a string. I'll push the fix up shortly.

PR #3641, opened 2020-12-15
Add suggested actions flow layout

### LINKED ISSUE #2745: Show suggested actions horizontally with line-break
Stacking the suggested actions takes up to much real estate. Most efficient way to show multiple suggested actions is horizontally with line-breaks with the option (nice to have) to align right, middle or center.
Good example sceenshot is in https://github.com/microsoft/BotFramework-WebChat/pull/2596

compulim:
> Fixes #2745.

### Added

-  Resolves [#2745](https://github.com/microsoft/BotFramework-WebChat/issues/2745). Added new `inline` layout to suggested actions, by [@compulim](https://github.com/compulim) in PR [#XXX](https://github.com/microsoft/BotFramework-WebChat/pull/XXX)

## Description

Added a new layout to suggested action: `inline`. And a new option to curb the height of the container and make it scrollable: `suggestedActionsInlineMaxHeight`.

image

## Specific Changes

- Added new layout to `SuggestedActions.js`, and refactored
- Modified `SuggestedAction.js` to use non-flexbox for laying out images inside buttons
- Updated `styleOptions`:
   - Added new `inline` option to `suggestedActionLayout`
   - Added new `suggestedActionsInlineMaxHeight`

## Review Checklist


packages/component/src/SendBox/SuggestedAction.js
```diff
@@ -81,7 +78,13 @@ const SuggestedAction = ({
   return (
     <div
       aria-hidden={ariaHidden}
-      className={classNames('webchat__suggested-action', rootClassName, suggestedActionStyleSet + '')}
+      className={classNames(
+        'webchat__suggested-action',
+        { 'webchat__suggested-action--rtl': direction === 'rtl' },
```
corinagum, line 83: Will this cause behavior change? I seem to recall that we didn't need to flip for RTL but I could be wrong.

corinagum (approved):
One question, + don't forget to rename to flow instead of inline. LGTM!

packages/component/src/SendBox/SuggestedAction.js
compulim: The current layout in RTL was not exactly right, the padding was too thin because we put the padding on the wrong side under RTL.

In the screenshot below, the green indicates the edge of the UI. And the padding on the right hand side was too narrow because we accidentally put it on the left (ignoring RTL).

So the change should be good.

image

(From https://github.com/microsoft/BotFramework-WebChat/blob/master/__tests__/__image_snapshots__/chrome-docker/rtl-js-rtl-ui-should-show-suggested-actions-with-images-1-snap.png)

eedwards36:
When will this be released? Also what is the exact usage? 

e.g. `suggestedActionLayout: 'inline',`

PR #3687, opened 2021-01-25
Fix anchors being disabled in obsolete Adaptive Cards

### LINKED ISSUE #3582: [Webchat modification required] : https://github.com/microsoft/BotFramework-WebChat/tree/master/samples/05.custom-components/l.disable-adaptive-cards
I was working on one repro and encountered that the : https://github.com/microsoft/BotFramework-WebChat/tree/master/samples/05.custom-components/l.disable-adaptive-cards has an extra line of code added on line 205 :               disabled={disabled}

## Version

>
>

## Steps to reproduce

1. Go to https://github.com/microsoft/BotFramework-WebChat/blob/master/samples/05.custom-components/l.disable-adaptive-cards/index.html
1. try to run the sample
1. we will see error 
1. post removing code from line 205 : disabled={disabled}
It will work 

It do not work with the code where we have  disabled={disabled} added under the ReactWebChat

## Expected behavior

provide a response 

## Additional context

[Bug]

### LINKED ISSUE #3618: Adaptive Card links should not be disabled even if the card is disabled
## Version
4.11.0

## Describe the bug

Currently, published behavior is that Web Chat's anchors will not work on first click, but will redirect to the url on second click. The original intention was to disable all element clicks, not just the first one. This behavior needs to be fixed.

When Adaptive Cards are disabled after use (when they are no longer the most recent activity), buttons and inputs should be disabled, but anchors should not. 

Right now the AC renderer will disable the following elements: `<a>`, `<button>`, `<input />`, `<select>`, `<textarea>`. All elements except anchors are reasonable to disable, but anchors should still be accessible even if the rest of the card is not. Since all anchors in Adaptive Cards are treated as links, as opposed to buttons, it's superfluous to disable them.

This bug is two-fold:
1.  Disabling the anchor should be removed
1. The event listener undoing the action for other elements should listen on ALL clicks, not just the first one.

## Steps to reproduce
1. Show a card that contains an anchor
1. Send another activity
1. Click on anchor
1. First click does nothing
1. Second click navigates to url

`attachmentMiddleware`:
```
      var attachmentMiddleware = function () {
        return function (next) {
          return function (card) {
            var attachment = card.attachment;
 
            switch (attachment.contentType) {
              case 'application/vnd.microsoft.card.adaptive':
                return window.React.createElement(window.WebChat.Components.AdaptiveCardContent, {
                  actionPerformedClassName: 'card__action--performed',
                  content: attachment.content,
                  disabled: true
                  // disabled: !recentBotMessage
                });
 
              default:
                return next(card);
            }
          };
        };
      };
```

## Expected behavior
Anchors should not be disabled, and disabled elements's event listener with undo should happen on every subsequent click, not just on the first click.

1. Remove anchors from disabled elements list
1. Add `addEventListenerWithUndo` that listens to all clicks, not just the first (as opposed to `addEventListenerOnceWithUndo`

### Addtional context

Testing scenarios:
1. With AC disabled, make sure links is still working and opening new window
1. Add a new test
   1. Add a new attachment middleware that disable AC, but not hero card
   1. Put a document.onclick handler
   1. Type "card breakfast"
   1. With AC disabled, click on one of those disabled button
   1. Expect, the document.onclick handler is not fired
   1. Type "herocard qna"
   1. Click on one of the button on hero card
   1. Expect, the document.onclick handler is fired once

[Bug]

corinagum:
> Fixes #3618 
> Fixes #3582

## Description

Previously, Adaptive Cards that have been marked as obsolete in the DOM were disabling anchors, which is invalid behavior. This change makes it so that while other inputs are disabled, anchors remain enabled and can be clicked/navigated to when in an older message in the transcript.

This also fixes the Disable Adaptive Cards sample, which had an error in it.

-  ~[ ] I have updated documentation~

## Review Checklist

### DIFF
--- a/packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
+++ b/packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
@@ -131,7 +131,7 @@ function disableElementWithUndo(element) {
 }
 
 function disableInputElementsWithUndo(element, observeSubtree = true) {
-  const undoStack = [].map.call(element.querySelectorAll('a, button, input, select, textarea'), element =>
+  const undoStack = [].map.call(element.querySelectorAll('button, input, select, textarea'), element =>
     disableElementWithUndo(element)
   );
 
--- a/packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
+++ b/packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
@@ -84,6 +84,12 @@ function addEventListenerOnceWithUndo(element, name, handler) {
   return detach;
 }
 
+function addEventListenerWithUndo(element, name, handler) {
+  element.addEventListener(name, handler);
+
+  return () => element.removeEventListener(name, handler);
+}
+
 function disableElementWithUndo(element) {
   const undoStack = [];
   const isActive = element === document.activeElement;
@@ -91,12 +97,6 @@ function disableElementWithUndo(element) {
 
   /* eslint-disable-next-line default-case */
   switch (tag) {
-    // Should we not disable <a>? Will some of the <a> are styled as button?
-    case 'a':
-      undoStack.push(addEventListenerOnceWithUndo(element, 'click', disabledHandler));
-
-      break;
-
     case 'button':
     case 'input':
     case 'select':
@@ -114,7 +114,7 @@ function disableElementWithUndo(element) {
       }
 
       if (tag === 'input' || tag === 'textarea') {
-        undoStack.push(addEventListenerOnceWithUndo(element, 'click', disabledHandler));
+        undoStack.push(addEventListenerWithUndo(element, 'click', disabledHandler));
         undoStack.push(setAttributeWithUndo(element, 'readonly', 'readonly'));
       } else if (tag === 'select') {
         undoStack.push(
--- a/packages/playground/src/App.js
+++ b/packages/playground/src/App.js
@@ -69,7 +69,7 @@ function App() {
   const [messageActivityWordBreak, setMessageActivityWordBreak] = useState('break-word');
 
   // 'middlewareDisableObsoleteAC': Middleware that renders inputs disabled on an Adaptive Card that is no longer the latest activity.
-  const [middlewareDisableObsoleteAC, setMiddlewareDisableObsoleteAC] = useState(
+  const [disableObsoleteAC, setDisableObsoleteAC] = useState(
     () => window.sessionStorage.getItem('PLAYGROUND_MIDDLEWARE_DISABLE_OBSOLETE_AC') === 'true'
   );
 
@@ -123,7 +123,7 @@ function App() {
     const { activity, attachment } = card;
     const { activities } = store.getState();
 
-    if (middlewareDisableObsoleteAC) {
+    if (disableObsoleteAC) {
       const messages = activities.filter(activity => activity.type === 'message');
       const mostRecent = messages.pop() === activity;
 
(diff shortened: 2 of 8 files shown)


packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
```diff
@@ -84,19 +84,19 @@ function addEventListenerOnceWithUndo(element, name, handler) {
...
+
 function disableElementWithUndo(element) {
   const undoStack = [];
   const isActive = element === document.activeElement;
   const tag = element.nodeName.toLowerCase();
 
   /* eslint-disable-next-line default-case */
   switch (tag) {
-    // Should we not disable <a>? Will some of the <a> are styled as button?
-    case 'a':
```
amal-khalaf, line 95: should we also  remove 'a' on line 134 ?

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
```diff
@@ -84,19 +84,19 @@ function addEventListenerOnceWithUndo(element, name, handler) {
...
 }
 
+function addEventListenerWithUndo(element, name, handler) {
+  element.addEventListener(name, handler);
+
+  return () => element.removeEventListener(name, handler);
+}
+
 function disableElementWithUndo(element) {
   const undoStack = [];
```
amal-khalaf, line 94: just wondering why do we call it **undo**stack?

packages/playground/src/App.js
```diff
@@ -57,6 +68,11 @@ function App() {
 
   const [messageActivityWordBreak, setMessageActivityWordBreak] = useState('break-word');
 
+  // 'middlewareDisableObsoleteAC': Middleware that renders inputs disabled on an Adaptive Card that is no longer the latest activity.
+  const [middlewareDisableObsoleteAC, setMiddlewareDisableObsoleteAC] = useState(
```
amal-khalaf, line 72: nit: rename  to DisableObsoleteAC

packages/playground/src/App.js
```diff
@@ -185,14 +233,14 @@ function App() {
...
     (e, checked) => {
-      setSendTypingIndicator(!!checked);
+      setSendTypingIndicator(checked);
     },
     [setSendTypingIndicator]
   );
 
   const handleUIDisabledChange = useCallback(
     (e, checked) => {
-      setDisabledUI(!!checked);
```
amal-khalaf, line 195: why did we remove `!!`

packages/playground/src/App.js
```diff
@@ -21,8 +21,19 @@ function App() {
   const REDUX_STORE_KEY = 'REDUX_STORE';
 
   // fluentui settings
-  const stackTokens = { childrenGap: 10 };
-
+  const stackTokens = { childrenGap: 0 };
+
+  // TODO: Add action buttons in tooltip later
+  const wordBreakTooltipProps = {
```
amal-khalaf, line 27: it is not clear what is this WordBreakToolTip do could we add  2 liners  what is this doing/testing?

amal-khalaf (changes requested):
pending comments

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
corinagum: yeah, thanks for the catch

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardRenderer.js
corinagum: not sure - would have to ask compulim

packages/playground/src/App.js
corinagum: this is the renderer for that tooltip, since it needs jsx. Not sure what you mean by 2 liners?

amal-khalaf (approved):

packages/playground/src/App.js
corinagum: I already have falsy checks in state or session storage, plus I think it was confusing me during testing. Do you prefer keeping it? I can revert and I'll test to verify behavior if yes.

PR #3711, opened 2021-02-08
#3453 Fixes plain text file attachments to show download link when uploaded

### LINKED ISSUE #3453: Sending .txt or .csv file results in empty chat bubble in Web Chat, Emulator
## Screenshots
image

image

## Version
v4.10.0 CDN

## Describe the bug
A customer reports that sending a .txt or .csv file via the send attachment button results in an empty chat bubble. Looking at the generated HTML, the element(s) that should populate the `<div class="webchat__bubble__content">` div is missing. The activity with attached file is still sent to the bot. 

## Steps to reproduce

1. Open Web Chat and/or Emulator
1. Click the send attachment button to send a file
1. Select either a .txt or .csv file and click 'Open'
1. Observe the empty chat bubble

## Expected behavior
The chat bubble should show the name and size of the attached .txt or .csv file.

## Additional context
Reported via an internal IcM. Provided customer with a temporary workaround for Web Chat, however they are wondering when a fix will be implemented.

-------------------------
**Edit** - Here is the requested workaround provided to the customer:

> Essentially, all that is happening here is we are updating the HTML that represents the file being sent.
> 
> The "store" first monitors for any dispatched "SEND_FILES" actions from Web Chat and captures the file name and size when they pass thru.
> It, next, waits for the message activity generated by Web Chat where an attachment is present and the attachment is either of type "text" or "csv".
> When the "store" identifies an appropriate activity, it filters thru the HTML looking for the last chat bubble from the user. This should be the bubble that is missing the file name and size.
> It checks if the bubble is actually missing the content and, if so, adds in the missing HTML code.
>  
> 
> Here are a couple things to note:
> 
> The "setTimeout()" is necessary as, without it, the filtering process occurs before Web Chat actually renders the elements. Some adjustment to the delay period may be necessary, but going much less than 50ms and the rendering could be missed. Going above ~200ms and the users will start to see the HTML being appended. 
> Appending the HTML has a slight inherent risk because certain Web Chat classes are included. As Web Chat is updated those classes may be subject to changing. It may be better for the customer to add just a little bit more code to capture the actual CSS and then apply it the activity or create some CSS rules to override the classes. 
> 
> ```
> let fileName = {};
> let fileSize = 0;
> const store = window.WebChat.createStore( {}, ({dispatch}) => next => action => {
>   if (action.type === 'WEB_CHAT/SEND_FILES') {
>     fileName = action.payload.files[0].name;
>     fileSize = Math.round(action.payload.files[0].size / 1000);
>   }
>   if (action.type === 'DIRECT_LINE/INCOMING_ACTIVITY') {
>     const { activity } = action.payload;
>     if (activity.type === 'message') {
>       if (activity?.attachments ) {
>         if ( activity.attachments[ 0 ]?.contentType === 'text/plain' || activity.attachments[ 0 ]?.contentType === 'text/csv' ) {
>           setTimeout( () => {
>             let fromUser = document.querySelectorAll( '.webchat__bubble--from-user' )
>             fromUser = Object.values(fromUser);
>             let indexOfLastBubble = fromUser.length - 1;
>             let bubble = fromUser[indexOfLastBubble];
>             let bubbleContent = bubble.querySelector('.webchat__bubble__content');
>             if (bubbleContent.childElementCount < 1) {
>               bubbleContent.innerHTML = `<div aria-hidden="true" class="webchat__fileContent css-19hx76s css-is9rv4"><div class="css-9ohtah">${fileName} of size ${fileSize} kB</div><div aria-hidden="true" class="webchat__fileContent__badge"><div class="webchat__fileContent__fileName">${fileName}</div><div class="webchat__fileContent__size">${fileSize} kB</div></div></div>`;
>             }
>           }, 50 );
>         }
>       }
>     }
>   }
>   next(action);
> } );
> ```

[Bug]

corinagum:
> Fixes #3453

## Changelog Entry

-  Fixes [#3453](https://github.com/microsoft/BotFramework-WebChat/issues/3453). Fixes plain text file attachments to show download link when uploaded, by [@corinagum](https://github.com/corinagum) in PR [#3711](https://github.com/microsoft/BotFramework-WebChat/pull/3711)

## Description

Previously, attachments of type plain text were uploaded and shown as an empty bubble. This fix now shows the attachment as a download link with size listed as well.

-  [NA] I have updated documentation

## Review Checklist


packages/component/src/Middleware/AttachmentForScreenReader/createCoreMiddleware.js
```diff
@@ -17,7 +17,7 @@ export default function createCoreMiddleware() {
         }
       ] = args;
 
-      return role === 'user' && !/^text\//u.test(contentType) && !thumbnailUrl
+      return role === 'user' && !/^audio|image|video\//u.test(contentType) && !attachment.content && !thumbnailUrl
```
amal-khalaf, line 20: why are checking if the  attachment.content does not exist? is this the expected behavior with text  file Content?

amal-khalaf (commented):
LGTM - left one question

packages/component/src/Middleware/AttachmentForScreenReader/createCoreMiddleware.js
corinagum: If the attachment doesn't exist, the text renderer is used to render the text of an activity. This is the original cause of the error.

packages/component/src/Middleware/Attachment/createCoreMiddleware.js
```diff
@@ -20,8 +20,8 @@ export default function createCoreMiddleware() {
           attachment: { contentType, contentUrl, thumbnailUrl } = {}
         }
       ] = args;
-
-      return role === 'user' && !/^text\//u.test(contentType) && !thumbnailUrl ? (
+      // && !attachment.content
```
compulim, line 23: Remove?

packages/component/src/Middleware/AttachmentForScreenReader/TextAttachment.js
```diff
@@ -3,16 +3,21 @@ import { hooks } from 'botframework-webchat-api';
 import PropTypes from 'prop-types';
 import React from 'react';
 
+import FileAttachment from './FileAttachment';
 import useUniqueId from '../../hooks/internal/useUniqueId';
 
 const { useLocalizer } = hooks;
 
-const TextAttachment = ({ attachment: { content = '' } = {} }) => {
+const TextAttachment = (activity, attachment, { attachment: { content = '' } = {} }) => {
```
compulim, line 11:
```suggestion
const TextAttachment = ({ activity, attachment, attachment: { content = '' } = {} }) => {
```

In expanded form:

```js
const TextAttachment = props => {
  const { activity, attachment } = props;
  const { content = '' } = attachment || {}; // If no attachment, or attachment.content is undefined, return empty string.

  // ...
};
```

packages/component/src/Middleware/Attachment/createCoreMiddleware.js
```diff
@@ -20,8 +20,8 @@ export default function createCoreMiddleware() {
...
+      return role === 'user' && !/text\//u.test(contentType) && !thumbnailUrl ? (
```
compulim, line 24: Why removing the `^` (starts with operator)?

corinagum: this was a typo. fixing.

packages/component/src/Middleware/AttachmentForScreenReader/TextAttachment.js
```diff
@@ -3,16 +3,21 @@ import { hooks } from 'botframework-webchat-api';
...
   const labelId = useUniqueId('webchat__id');
   const localize = useLocalizer();
 
   const label = localize('ATTACHMENT_TEXT', content);
 
+  if (!attachment.content) {
```
compulim, line 17: Prefer to do this in the middleware, other than in `<TextAttachment>` component.

packages/component/src/Middleware/AttachmentForScreenReader/createCoreMiddleware.js
```diff
@@ -17,7 +17,7 @@ export default function createCoreMiddleware() {
         }
       ] = args;
 
-      return role === 'user' && !/^text\//u.test(contentType) && !thumbnailUrl
+      return role === 'user' && !/^text\//u.test(contentType) && !attachment.content && !thumbnailUrl
```
compulim, line 20: I think this one is, "show as file if":

- If this is from user, or
- If this is a `text/*`
  - Show as file if it does not have `attachment.content`
- Otherwise, if this is not a `text/*`
  - Show as file if it does not have `attachment.thumbnailUrl`


```suggestion
      return role === 'user' && !/^text\//u.test(contentType) && !attachment.content && !thumbnailUrl
```

packages/component/src/Middleware/AttachmentForScreenReader/TextAttachment.js
corinagum: I spent hours trying to get it to work. it stays here.

packages/component/src/Middleware/AttachmentForScreenReader/createCoreMiddleware.js
corinagum: I'm not seeing a diff.

__tests__/upload.js
```diff
@@ -282,3 +285,20 @@ test('upload a ZIP file', async () => {
 
   expect(base64PNG).toMatchImageSnapshot(imageSnapshotOptions);
 });
+
+test('upload a .txt (plain) file', async () => {
```
compulim, line 289: Do we have a test that, the bot is sending an activity with `attachment.content`?

This is because, the attachment that the user send/upload, will also be a file. Only the bot sending the text attachment may handle as preview or download link.

```json
{
  "attachments": [
    {
      "content": "Hello, World!",
      "contentType": "text/plain"
    }
  ],
  "type": "message"
}

compulim (approved):

PR #3959, opened 2021-06-22
Disable SRI for webchat.js hosted with DirectLine (vs CDN)

p-nagpal:
Disable SRI for webchat.js hosted with DirectLine (vs CDN)

## Description

Disable SRI for webchat.js hosted with DirectLine (vs **CDN)**. The hosted webchat gets signed differently during the service builds.


compulim (approved):

compulim:
@p-nagpal Let me know when you have finished testing it.

PR #3974, opened 2021-07-06
Add Web Audio priming/blessing

### LINKED ISSUE #2316: Consider priming text-to-speech thru keyboard typing and any possible gesture
## Feature Request

Today, we prime the speech synthesis engine by sending an empty utterance. When the user click/tap on the send button, we prime the engine.

If the user use <kbd>ENTER</kbd> to send the message, we might not prime the engine. This need to be verified.

If this is the case, we should prime when the user type. Or consider other entry points for priming the engine.

[Enhancement]

### LINKED ISSUE #3823: Speech to text issue in Safari
**Screenshots**

**Version**
Using the latest version of webchat.js (referred directly from CDN)
iOS: latest version

**Describe the bug**
The bug is only present in iOS Safari browser.
When a user taps the microphone and the last bot message contains a choice button, the microphone is not re-enabled and is stuck in "Starting...". From that moment on, there is no way for the microphone to work.

**Steps to reproduce**
- Enable Speech To Text (basic recognition)
- Tap the microphone
- Allow to access the microphone
- Say something
- Wait for the bot response (with a choice button)
- The microphone is not anymore re-enabled

**Expected behavior**
The microphone should be re-enabled or the user should be able to re-enabled it by pressing the microphone button.

**Additional context**
There is no error in the console browser.

[Bug]

### LINKED ISSUE #3899: [Investigations] Speech on iPhone stuck at "Starting" state.
> Pratyush reported this issue. Need a repro.

## Screenshots

image

## Version

>
>

## Describe the bug

Shortly after speech permission is granted to the website on iPhone + Safari, it get stuck at "Starting" state.

## Steps to reproduce

> Need to fill this in.

1. Click on '....'
1. Scroll down to '....'
1. See error

## Expected behavior

It should start recording on microphone.

Instead, it stuck at "Starting".

## Additional context

[Bug]

compulim:
> Fixes #2316. Fixes #3823. Fixes #3899.

### Added

-  Resolves [#2316](https://github.com/microsoft/BotFramework-WebChat/issues/2316). Added blessing/priming of `AudioContext` when clicking on microphone button, by [@compulim](https://github.com/compulim), in PR [#3974](https://github.com/microsoft/BotFramework-WebChat/pull/3974)

### Fixed

-  Fixes [#3823](https://github.com/microsoft/BotFramework-WebChat/issues/3823) and [#3899](https://github.com/microsoft/BotFramework-WebChat/issues/3899). Fix speech recognition and synthesis on Safari, in PR [#3974](https://github.com/microsoft/BotFramework-WebChat/pull/3974)

## Description

Instead of using microphone input source from Speech SDK (a.k.a. `AudioConfig.fromMicrophone`), we are implementing our own audio input source. For a few reasons:
- In Safari, `AudioContext` need to be primed/blessed
   - Blessing means we call `AudioContext.resume()` from code that is initiated from user gestures, such as clicking on the microphone button
- `AudioContext` should be reused to keep its blessing status

However, the adapter provided by Speech SDK:
- Will not keep the instance of `AudioContext`, always close after use
- May not call `AudioContext.resume()` from a user gesture event

This means, in Safari, it may occasionally lose access to microphone.

We also added new internal hook `useResumeAudioContext`. So we can call the hook to from time to time to bless the `AudioContext` continuously.

## Design

We are based from Speech SDK `AudioConfig.fromStreamInput` and `Push/PullAudioInputStream`.

There are few caveats with `Push/PullAudioInputStream`:

- `PushAudioInputStream`
   - We can call `write()` occasionally to push buffer to it
   - Does not support more than one recognition, will throw an error
   - Does not signal us when the input stream is tearing down by the speech recognition engine (such as, when the recognition is completed)
- `PullAudioInputStream`
   - We provide a callback function, which the `PullAudioInputStream` will call continuously to pull buffer from us
   - It will continuously call the callback function to pull in more data
   - The callback function must be synchronous
   - That means, before the `PullAudioInputStream` is set up, *all data* must be ready synchronously
   - Microphone input are not synchronous and data are not ready at `t=0`

Instead of basing off `Push/PullAudioInputStream`, we are basing off from their base class `AudioInputStream` instead. But there are lot of quirks in their original implementation.

- `attach` is called, but `detach` is never called
- `turnOff` is called, but `turnOn` is never called
- `close` is marked as abstract (must implement), but never called

We wrapped Speech SDK `AudioInputStream` implementation, cleaning up quirks, and internally expose as `createAudioConfig`, which only requires 1 callback:

- `attach`, which returns a `Promise` of:
   - Streams of `ArrayBuffer` chunks
   - Audio format (e.g. 16-bit mono 16 kHz)
   - Device info (manufacturer, model, connectivity type, device type)
- `turnOff` (optional), which will be called before the device is tearing down
   - Does not call when used in Direct Line Speech

We use our `createAudioConfig` in 2 different areas to proof its efficiency: real microphone input, and mocked input in test harness.

In the future, when we bump Speech SDK, we will need to make sure our `createAudioConfig` will continue to work with their latest version.

## Specific Changes

- Added new `createAudioConfig` for implementing custom `AudioConfig` for Speech SDK without understanding its complexity
- When using Cognitive Services or Direct Line Speech and `audioConfig` is not passed, we will use our custom `AudioConfig` for microphone input
   - When using `botframework-directlinespeech-sdk` alone without Web Chat, it will NOT use our custom `AudioConfig`
- Added new mock microphone input in test harness to use the new `createAudioConfig`
- Added internal `useResumeAudioContext` hook to continuously bless `AudioContext` object
- Bless `AudioContext` object when `pointerdown` event is received from `window` object

## Review Checklist


packages/bundle/src/speech/createAudioConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+    deviceInfo: DeviceInfo;
+    format: Format;
+  }> {
+    return this[SYMBOL_ATTACH](audioNodeId);
+  }
+
+  protected performTurnOff(): Promise<void> {
+    const { [SYMBOL_TURN_OFF]: turnOff } = this;
+
+    return turnOff && turnOff();
```
cwhitten, line 60: does this need a better typecheck than simple existence? what if `turnOff` is a string

cwhitten (approved):

packages/bundle/src/speech/createAudioConfig.ts
compulim: Added type checking to constructor.

PR #4329, opened 2022-06-25
Fix Adaptive Cards `TextBlock` element missing `aria-level`

### LINKED ISSUE #4327: Title in Hero Card does not have `aria-level` specified
## Screenshots

image

## Version

`main`

## Describe the bug

In Adaptive Cards library, by default, all `TextBlock` with `style="heading"` will have `role="heading"` set. However, the required attribute, `aria-level`, was not set. To set the `aria-level`, we need to *explicitly* specify `hostConfig.textBlock.headingLevel`.

Adaptive Cards library added this new host config option in their https://github.com/microsoft/AdaptiveCards/issues/5635 and https://github.com/microsoft/AdaptiveCards/pull/5698.

Without specifying this new host config option, the rendered DOM tree will be invalid by WCAG 4.1.2 (aria-required-attr). This is because all elements with `role="heading"` MUST have `aria-level` specified.

## Steps to reproduce

1. Talk to MockBot
2. Say "herocard"
3. Inspect the element "Details about image 1"

## Expected behavior

The element has `role="heading"`, however, it does not have `aria-level` attribute specified.

## Action items

Make sure to update our `validateDOMTree()` function to verify this issue on a broader scale.

## Additional context

New bug https://github.com/microsoft/AdaptiveCards/issues/7566 filed to `microsoft/AdaptiveCards` repo.

- https://github.com/microsoft/AdaptiveCards/issues/5635
- https://github.com/microsoft/AdaptiveCards/pull/5698

[Bug]

compulim:
> Fixes #4327.

### Fixed

-  Fixes [#4327](https://github.com/microsoft/BotFramework-WebChat/issues/4327). In Adaptive Cards, `TextBlock` with `style="heading"` should have `aria-level` set, by [@compulim](https://github.com/compulim), in PR [#4329](https://github.com/microsoft/BotFramework-WebChat/issues/4329)

## Description

By default, Adaptive Cards generate DOM with `role="heading"` without `aria-level` set. This violated [WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) ([ARIA5](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA5)).

In order to set `aria-level`, we need to explicitly set `ACHostConfig.textBlock.headingLevel` to `1`. This is an [undocumented option](https://docs.microsoft.com/en-us/adaptive-cards/rendering-cards/host-config#schema-textblockconfig) introduced by https://github.com/microsoft/AdaptiveCards/issues/5635 and https://github.com/microsoft/AdaptiveCards/pull/5698.

## Design

This is code in [Adaptive Cards repo `card-elements.ts`](https://github.com/microsoft/AdaptiveCards/blob/ff28a1fd75d5dbe62d3fc3940f47652190f5a506/source/nodejs/adaptivecards/src/card-elements.ts#L920).

image

If the Adaptive Card `TextBlock` element has `style="heading"`, the `role="heading"` will be applied. However, `aria-level` is not set unless `ACHostConfig.textBlock.headingLevel` is set. This is an undocumented option and *required* for accessibility. Otherwise, we would fail with [WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) ([ARIA5](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA5)).

We need to explicitly set `ACHostConfig.textBlock.headingLevel` to make sure `aria-level` is set on `role="heading"`. As we do not use heading, we default it to level 1.

## Specific Changes

- Update `adaptiveCardHostConfig.ts` to add `textBlock.headingLevel` to `1`
- Update `verifyDOMIntegrity.js` for verifying our DOM tree is good
- Add `verifyDOMIntegrity()` call to all card-related attachments

## Review Checklist

### DIFF
--- a/packages/test/page-object/src/globals/pageObjects/verifyDOMIntegrity.js
+++ b/packages/test/page-object/src/globals/pageObjects/verifyDOMIntegrity.js
@@ -3,16 +3,22 @@
 export default function verifyDOMIntegrity() {
   // If an element has "aria-labelledby", the label element must be present on the screen.
   [].forEach.call(document.querySelectorAll('[aria-labelledby]'), element => {
-    const labelId = element.getAttribute('aria-labelledby');
-    const labelElement = document.getElementById(labelId);
+    element
+      .getAttribute('aria-labelledby')
+      .split(' ')
+      .map(labelId => labelId.trim())
+      .filter(Boolean)
+      .forEach(labelId => {
+        const labelElement = document.getElementById(labelId);
 
-    if (!labelElement) {
-      const message = `verifyDOMIntegrity: Cannot find element referenced by aria-labelledby attribute with ID "${labelId}".`;
+        if (!labelElement) {
+          const message = `verifyDOMIntegrity: Cannot find element referenced by aria-labelledby attribute with ID "${labelId}".`;
 
-      console.warn(message, element);
+          console.warn(message, element);
 
-      throw new Error(message);
-    }
+          throw new Error(message);
+        }
+      });
   });
 
   // No two elements can have the same ID.
@@ -35,6 +41,19 @@ export default function verifyDOMIntegrity() {
 
     if (~className.indexOf('undefined')) {
       const message = `No elements should have the keyword "undefined" in it, we saw "${className}".`;
+
+      console.warn(message, element);
+
+      throw new Error(message);
+    }
+  });
+
+  // Elements of `role="heading"` must have `aria-level` set
+  [].forEach.call(document.querySelectorAll('[role="heading"]'), element => {
+    const ariaLevel = +element.getAttribute('aria-level');
+
+    if (!(ariaLevel >= 1 && ariaLevel === ~~ariaLevel)) {
+      const message = 'Elements of role="heading" must have aria-level set to an integer equal to or greater than 1.';
 
       console.warn(message, element);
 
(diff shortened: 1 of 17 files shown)


packages/test/page-object/src/globals/pageObjects/verifyDOMIntegrity.js
```diff
@@ -3,16 +3,22 @@
...
-
-      console.warn(message, element);
-
-      throw new Error(message);
-    }
+    element
+      .getAttribute('aria-labelledby')
+      .split(' ')
+      .map(labelId => labelId.trim())
+      .filter(Boolean)
```
tdurnford, line 10: :)

packages/test/page-object/src/globals/pageObjects/verifyDOMIntegrity.js
```diff
@@ -41,4 +47,17 @@ export default function verifyDOMIntegrity() {
       throw new Error(message);
     }
   });
+
+  // Elements of `role="heading"` must have `aria-level` set
+  [].forEach.call(document.querySelectorAll('[role="heading"]'), element => {
+    const ariaLevel = +element.getAttribute('aria-level');
+
+    if (!(ariaLevel >= 1 && ariaLevel === ~~ariaLevel)) {
```
tdurnford, line 55: Interesting... when would `ariaLevel === ~~ariaLevel` be `false`? ... `ariaLevel = true` is the only scenario I can think of for `!(ariaLevel >= 1 && ariaLevel === ~~ariaLevel)` to be `true`.

`!(true >= 1 && true === ~~true) => true`

tdurnford (approved):

packages/test/page-object/src/globals/pageObjects/verifyDOMIntegrity.js
compulim: Hehe, `0.1`.

```
> value = 0.1
0.1
> value === ~~value
false
```

compulim: A very large integer number could also return `false` for `bigInteger === ~~bigInteger`... such as an Unix epoch.

Issue #4263, opened 2022-04-21
Webchat Widget Accessibility issue


jason-walgreens:
When hidden or new content is revealed or available on the same screen because of any user interaction then focus need to move to the new content either immediately or with one additional keystroke, so that a screen reader user can relate and access the new content without any interruption. In current implementation when a user types something and because of this input a new set of options are available on the chat modal for user to select options from, focus remains in the input field (Type your message, field). Screen reader reads the appearance of the new content but does not read the location of the content and how to access them.
In order to access the options a user need to press “shift+tab” key multiple times. Which is not logical. Since user does not know the location of the newly available content, so he is also not sure on how to access them.

Expectation is, when a new content is available on the screen focus should also move to the new content so that a user can easily access them by pressing “tab” key.

WCAG references: 2.4.3

jason-walgreens:
[proposed labels] Bot-framework, Webchat Widget, A11Y , ADA Accessibility

tracyboehrer:
@jason-walgreens Thanks for the input.  I'll add this as a feature request for future consideration.

compulim:
If the message is interactive (card or contains buttons), the screen reader will narrate one of the followings and it should guide the end-user how to navigate to the message content.

- "Message is interactive. Press shift tab key 2 to 3 times to switch to the chat history. Then click on the message to interact."
- "One or more links in the message. Press shift tab key 2 to 3 times to switch to the chat history. Then click on the message to interact."

However, if the message is not interactive, we will not narrate these instructions as non-interactive elements (such as `<h1>`, `<p>`, or `<li>`) should never be focusable.

There are 2 possible failures for [WCAG 2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html):

- [F44: Failure of Success Criterion 2.4.3 due to using tabindex to create a tab order that does not preserve meaning and operability](https://www.w3.org/WAI/WCAG21/Techniques/failures/F44)
   - The tab order which Web Chat propose, is commonly used chat UI, where the send box is *below* the chat history
   - As the chat history (and potential interactive content) appears *above* the send box, we chose SHIFT-TAB to jump to the content, as this preserves the visual sequence and common understanding of chat UI
- [F85: Failure of Success Criterion 2.4.3 due to using dialogs or menus that are not adjacent to their trigger control in the sequential navigation order](https://www.w3.org/WAI/WCAG21/Techniques/failures/F85)
[...]

PR #4335, opened 2022-07-03
Fix accessibility issues in Adaptive Cards library

### LINKED ISSUE #3949: Adaptive Cards: Action set with role="menubar" must have 1 or more descendants with role="menuitem".
> This is not same as #3947. This one related to initial render by Adaptive Card. #3947 related to our hacks after rendering.

## Version

`main` which is using `adaptivecards@2.9.0`.

Also repro on [Adaptive Cards Designer](https://adaptivecards.io/designer/) independently.

Not repro on `adaptivecards@2.5.0` which was used by `4.13.0`.

## Describe the bug

Our customer reported that all elements with `role="menubar"` must have 1 or more descendants of element with `role="menuitem"`.

According to WAI-ARIA:

- `menubar` has "Required Owned Elements" of `menuitem`, `menuitemcheckbox`, `menuitemradio` (under or not under `group`)
   - See here, https://www.w3.org/TR/wai-aria/#menubar
- "Required Owned Elements" means, the role must contains 1 or more elements of 1 of the types
   - See here, https://www.w3.org/TR/wai-aria/#mustContain
   - Excerpt: "When multiple roles are specified as required owned elements for a role, at least one instance of one required owned element is expected."

`Action.Submit` buttons generated by `card.actions` is contained in a `menubar`. That `menubar` does not fulfill the "Required Owned Elements". This is because the action generated is a `button`, instead of `menuitem`.

## Steps to reproduce

Put the card in [Adaptive Card Designer](https://adaptivecards.io/designer/). Enable "Preview Mode". Then, inspect the DOM tree.

<details>
<summary>Expand to see the Adaptive Cards JSON</summary>

```json
{
  "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
  "type": "AdaptiveCard",
  "version": "1.3",
  "body": [
    {
      "type": "TextBlock",
      "text": "Do you want to take a survey?",
      "size": "Large",
      "weight": "Bolder",
      "wrap": true
    },
    {
      "type": "ActionSet",
      "actions": [
        {
          "type": "Action.Submit",
          "title": "Yes"
        },
        {
          "type": "Action.Submit",
          "title": "No"
        }
      ]
    },
    {
      "type": "TextBlock",
      "text": "Card action below",
      "size": "Large",
      "weight": "Bolder",
      "wrap": true
    }
  ],
  "actions": [
    {
      "type": "Action.Submit",
      "title": "Submit card"
    }
  ]
}
```
</details>

## Expected behavior

All elements with `role="menubar"` must have 1 or more descendants of element with `role="menuitem"`.

image

(This screenshot is taken from [Adaptive Cards Designer](https://adaptivecards.io/designer/) in Preview Mode)

## Additional context

[Bug]

compulim:
> Fixes #3949.

### Fixed

-  Fixes [#3949](https://github.com/microsoft/BotFramework-WebChat/issues/3949). For accessibility reasons, buttons in Adaptive Cards should be `role="button"` instead of `role="menubar"`/`role="menuitem"`, by [@compulim](https://github.com/compulim), in PR [#4263](https://github.com/microsoft/BotFramework-WebChat/issues/4263)

## Description

`ActionSet`/`Action` in AC is currently `menubar`/`menuitem` respectively. For accessibility reasons, actions should be `button`. This will help end-users who use screen reader to reach the action button via <kbd>B</kbd>/<kbd>F</kbd> keys.

We should also consider updating the instruction text for indicating how to navigate. Currently, it is "Message is interactive. Press SHIFT TAB two to three times to go to the chat history. Then select the message".

As `ActionSet`/`Action` are now buttons (form fields), we might be able to simplify it to "Message contains form fields."

This pull request will also refactor existing hacks into a more consumable/pluggable pattern.

## Design

We have isolated our DOM hacks to Adaptive Cards in a separate folder:

- Re-rendering should keep focus
- Re-rendering should keep values
- Adaptive Cards can be disabled
- `ActionSet` should not be `role="menubar"`
- `Action` should set `aria-pushed` attribute

We added a new internal `useAdaptiveCardModEffect()` hook to apply/undo DOM hacks. These hacks, need to undo right before we re-render, and right after we mount it on the DOM:

- Undo before re-render
   - Re-render could potentially unmount part of the DOM, say, hiding part of the card using `Action.ShowCard`
   - Some hacks, such as "persist user input values", are using the "undo mod" stage to save the values, then reapply the values during the "apply mod" stage
   - Thus, the "undo mod" stage should always run before re-render
- Reapply right after mounted
   - Some hacks, such as "persist user focus", need the Adaptive Cards to be mounted on the DOM tree. Otherwise, it will be no effect
   - Say, setting focus on an element which is not visible/mounted, is no-op
   - Thus, the "apply mod" stage should always run right after the DOM node is mounted

### Other designs

After Adaptive Card is rendered as an unmounted DOM tree, it could be walked using standard DOM traversal functions. When walking the DOM tree, we could programmatically turn it into React.

This approach leverage DOM reconciliation in React, which helps re-render.

However, for other hacks, we will need to modify the walking mechanism and apply the hack while walking the tree. This could be very difficult to add mods that requires traversing multiple nodes. The mod will be more-or-less similar to translating a DOM tree based purely on SAX interface, which could be very difficult to get done.

Thus, we are keeping our hacks as DOM-based, instead of React-based.

## Specific Changes

- Added infrastructure to apply/undo DOM hacks (a.k.a. mod) we done to Adaptive Cards
- Upgraded a legacy test `adaptiveCards.js/disable card inputs`

## Review Checklist


packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardHacks/private/useAdaptiveCardModEffect.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { AdaptiveCard } from 'adaptivecards';
+
+type ModFunction<TArgs extends unknown[] = []> = (
+  adaptiveCard: AdaptiveCard,
+  cardElement: HTMLElement,
+  ...args: TArgs
+) => () => void;
+
+class Mod<TArgs extends unknown[]> {
+  constructor(mod: ModFunction<TArgs>) {
```
tdurnford, line 15: I think you can do.. Not sure
```suggestion
  constructor(private #mod: ModFunction<TArgs>) {}
```

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardHacks/private/useValueRef.ts
```diff
@@ -0,0 +1,11 @@
+import { useRef } from 'react';
+
+import type { RefObject } from 'react';
+
+export default function useValueRef<T>(value: T): RefObject<T> {
+  const ref = useRef<T>(value);
+
+  ref.current = value;
```
tdurnford, line 8: should we wrap this is a `useEffect`?

```
useEffect(() => { ref.current = value;});
```

tdurnford (approved):

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardHacks/private/useAdaptiveCardModEffect.ts
compulim: image

Seems only working for TS's fake private modifier, but not JS's real `#` modifier.

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardHacks/private/closest.ts
```diff
@@ -0,0 +1,17 @@
+// Ponyfill `HTMLElement.closest`.
+// https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
+export default function closest(element: HTMLElement, selector: string): HTMLElement | undefined {
```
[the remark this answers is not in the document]
compulim: `core-js` is mainly targeting neutral environment and has very limited polyfills for DOM. Currently, `closest` is not polyfilled.

AFAIK, `core-js` will conditionally load polyfills based on capability detection.

https://github.com/zloirock/core-js/issues/317

packages/bundle/src/adaptiveCards/Attachment/AdaptiveCardHacks/private/useValueRef.ts
compulim: Not really. Unlike `usePrevious` hook which change the ref value after render (remember previous value), this hook change the value immediately. So the hook caller can immediately read the new value.

Before using this hook:

```js
const SomeComponent = ({ value }) => {
  const handleClick = useCallback(() => {
    // This function will be recreated every time `value` change. However, this is unnecessary because the `value` is only used post-render (during `onClick` event).
    alert(value);
  }, [value]);

  return <button onClick={handleClick}>Click me</button>;
}
```

After using this hook:

```js
const SomeComponent = ({ value }) => {
  const valueRef = useValueRef(value);
  const handleClick = useCallback(() => {
    alert(valueRef.current)
  }, [valueRef]);

  return <button onClick={handleClick}>Click me</button>;
}
```

I am thinking to create a `useCallbackWithCheck` to assert that we don't recreate callback functions after initial render, so we can detect this type of wasted renderings.

Issue #4326, opened 2022-06-23
How do I test my bot running on localhost?


dluc:
With Bot Framework Emulator I can test a bot running locally on localhost, e.g. connecting to http://localhost:3978/api/messages.

How do I connect WebChat to the same localhost endpoint?

![image](https://user-images.githubusercontent.com/371009/175388750-0c96732e-4030-4116-a180-3603b15f6600.png)

compulim:
Web Chat cannot connect to the bot directly. It must connect to a chat service.

The chat service could be, for example:

- Azure Bot Services (ABS)
   - You can use a HTTP tunneling service, such as ngrok, to redirect the traffic from ABS to localhost
- Bot Framework Emulator
   - Emulator is emulating ABS and will expose itself

When you run it in Emulator, in the Log window, you should see "Emulator listening on http://...". IIRC, you should pass it like this, using the `domain` property.

```js
window.WebChat.renderWebChat({
  directLine: window.WebChat.createDirectLine({
    domain: 'http://localhost:.../v3/directline'
  })
});
```

Could you try this out and see? You can run a Fiddler, run Emulator, and you should see Web Chat inside Emulator is pointing to a similar URL. That's what you need in the `domain` field.

I am recently busy working on some issues in Web Chat so I cannot get you the exact steps to get the job done. But I am 95% sure this is possible. 🙏🏻 Will be great if you can post your findings here. I will try to post more if I can find any.

compulim:
![image](https://user-images.githubusercontent.com/1622400/175496167-b5c62754-983e-49f9-8c80-a35568019516.png)

You can check out their code here, https://github.com/microsoft/BotFramework-Emulator/blob/main/packages/app/client/src/state/sagas/chatSagas.ts#L740. This code initialize DirectLineJS chat adapter and plug into Web Chat.

dluc:
hi @compulim, I'm looking for a development workflow that doesn't require a deployment, until the bot is ready. E.g. I want to enable any developer in my team to create web chat bots and show a demo (locally, sharing the screen, recording a video), without depending on cloud resources, Azure access, etc.

Looks like I need to add extra .NET code to handle requests to "/v3/directline", if I want to run the code locally without external dependencies.

[...]

PR #4362, opened 2022-07-22
Add "send failed retry" prompt to live region

### LINKED ISSUE #4211: [Accessibility] Screen reader does not announce Send failed message information
> Copied from [Composer#64488](https://fuselabs.visualstudio.com/Composer/_workitems/edit/64488).

## Version

`4.15.1` (or `main`)

## Describe the bug

If screen reader users navigate on Live Chat Box screen and message get failed and message appear send failed and screen reader does not announce this message, then it would be difficult for users to know sent message get failed and need to try again to send the message.

## Steps to reproduce

1. Open Bot Framework Emulator installed on the system.
1. Bot Framework Emulator Application with Welcome screen opens.
1. Navigate on the welcome screen and select Open Bot button.
1. Open A bot Dialog opens, navigate on the dialog and enter inputs in edit fields and select Connect button.
1. Bot Gets connected and a new Tab "Live Chat "opens, Navigate on the tab.
1. Verify that screen reader announces send failed message or not.

## Expected behavior

Screen reader should be announcing Send failed message information under Live Chat Box.
So that users know send message get failed, it does not deliver to another person.

## Additional context

Related to [WCAG 2.1 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html).

Some texts from designer:

```
Message failed to send. Go to message to resend.

Your message failed to send: .......... Retry.
```

[Bug]

### LINKED ISSUE #4322: Improve error message around send failed/timeout
## Feature Request

Today, we need 2 signals to indicate a message is successfully sent. They need to be received in a timely manner:

- The `postActivity` call to the service return successful
- The activity is being echoed back from the `activity$`

And if either one of them are not fulfilled, we will report as "send timeout" without much details.

We could improve the error by saying which signal is not received timely.

[feature-request]

compulim:
> Fixes #4211. Fixes #4322.

### Breaking changes

-  The `activity.channelData.state` property is being deprecated in favor of the updated [`activity.channelData['webchat:send-status']`](https://github.com/microsoft/BotFramework-WebChat/blob/main/packages/core/src/types/WebChatActivity.ts) property, main differences:
   -  The new `'webchat:send-status'` property will become `"send failed"` when the chat adapter failed to send the activity or after passing a hardcoded 5 minute timeout
   -  Previously the `state` property would become `"send failed"` when the chat adapter failed to send the activity or after passing a timeout as defined in `styleOptions.sendTimeout`
   -  See PR [#4362](https://github.com/microsoft/BotFramework-WebChat/pull/4362) for details

### Changed

-  Resolves [#4322](https://github.com/microsoft/BotFramework-WebChat/issues/4322). Improved error messages for sending activities, by [@compulim](https://github.com/compulim) in PR [#4362](https://github.com/microsoft/BotFramework-WebChat/pull/4362)
-  Resolves [#4211](https://github.com/microsoft/BotFramework-WebChat/issues/4211). Added new `useSendStatusByActivityKey` hook to check the UI send status of an outgoing activity, by [@compulim](https://github.com/compulim) in PR [#4362](https://github.com/microsoft/BotFramework-WebChat/pull/4362)
   -  The send status returned by this hook is designed to display different UIs that reflect the "sending", "send failed" or "sent" status of the activity
   -  When modifying `styleOptions.sendTimeout` prop, the send status returned by this hook may transition from `"send failed"` to `"sending"`, and vice versa
   -  This is different from the send status provided by the chat adapter, namely `activity.channelData['webchat:send-status']`

### Fixed

-  Fixes [#4211](https://github.com/microsoft/BotFramework-WebChat/issues/4211). Screen reader should read when an activity was failed to send, by [@compulim](https://github.com/compulim), in PR [#4326](https://github.com/microsoft/BotFramework-WebChat/issues/4326), also fixed:
   -  The "send failed" status on the activity should show up as soon as the chat adapter failed to send the activity

## Description

When an activity was failed to send (e.g. send timeout, network error, etc.), a "Send failed, retry" prompt will be shown on the chat history. However, this retry prompt is not narrated through live region.

This pull request is to add the retry prompt to the live region to make sure screen reader users are aware that the message they sent is failing and they can retry.

The "send failed" prompt will only read for failing message that are also non-presentational (e.g. not `event` or `typing` activities).

While we are working on this PR, we discovered a few flaws in the current version of Web Chat. The flaws are documented in the [design](#design) section of this PR description:

- `channelData.state === 'send failed'` was intentionally ignored
   - Activities that are legitimately failed (not timed out) will not be reflected until timeout has passed
   - In other words, in airplane mode (no network), outgoing activities should show up as failed immediately
   - However, currently, it take 20s for an activity to show up as failed
- Outgoing activity may not have `channelData.state` set momentarily
   - While sending an activity, if the chat service echoed back but the `postActivity()` call did not return yet, the activity-in-transit should still have send status of `sending`, instead of unset
- `createDirectLineWithTranscript` does not work properly for outgoing activities because the `postActivitySaga` is not triggered
   - Some of our existing tests forcefully put `channelData.state` in the `createDirectLineWithTranscript.activity$` to emulate the effect of `postActivitySaga`, this behavior is not correct

### New `<ActivitySendStatus>` provider

A new context provider to provide UI-oriented information on outgoing activity send status. Its corresponding hook `useSendStatusByActivityKey` will retrieve a map of send status by activity key. Its return value has type of `Map<string, 'sending' | 'send failed' | 'sent'>`.

All existing code related to send status of outgoing activities are being refactored to use this new provider.

New code added to chat history live region will use this provider to make sure it has a consistent experience as the visible chat history.

This send status is different from the `channelData['webchat:send-status']`:

| `channelData['webchat:send-status']` | `useSendStatusByActivityKey` |
| - | - |
| `"sending"` | `"sending"` if not timed out, otherwise, `"send failed"` |
| `"send failed"` | `"send failed"` |
| `"sent"` | `"sent"` |

Notes: timeout is based on `styleOptions.sendTimeout` or `styleOptions.sendTimeoutForAttachments`.

Web developers can change `styleOptions` to modify send timeout on-the-fly and the `useSendStatusByActivityKey` will reflect the change immediately.

### New `activity.channelData['webchat:send-status']`

This is very similar to the existing `activity.channelData.state` with a few differences:

- Prefixed with `webchat:` for better namespace
- `sending` will become `send failed` if and only if:
   - Failed to send due to fatal error, such as network error
      - This is same as previous behavior
   - Timed out for a hardcoded 5 minutes
      - Previously, timed out for X minutes as defined in `styleOptions.sendTimeout`

`activity.channelData.state === 'send failed'` is not very helpful because:
   - If `styleOptions.sendTimeout` is increased, it will NOT revive the status back to `sending` because the `postActivitySaga` has already stopped
   - It did not consider prolonged send timeouts for activity with attachments
   - The previous field will continue to use the standard value (default 20s). That means, for activity with attachments, which has a timeout of 120s, will be prematurely marked as timed out after 20s

For the reasons above, we ignored both `sending` and `send failed` state. We only considered the `sent` and not-`sent` state
   - If it took more than 20s (or 120s) in the non-`sent` state, we will mark the activity as "send failed"
   - If the activity cannot be sent due to fatal error, such as network error, we CANNOT immediately mark it as send failed because we ignored the `send failed` state. The UI will need to wait for 20s to reflect the fatal error

The newer channel data field will speed up showing "Send failed" as soon as fatal error is detected.

This field is NOT recommended to be a factor of whether the "Send failed. Retry" prompt should be shown or not. Web developers should use the new `<ActivitySendStatus>` provider and its `useSendStatusByActivityKey` hook. This is because this channel data field is not based on `styleOptions.sendTimeout`. Thus, it is not a direct representation of whether the "Send failed. Retry" prompt should be shown or not.

### Added test facility `createDirectLineEmulator`

Previously, we use `createDirectLineWithTranscript` and inject static transcript with `activity.channelData.state`. However, this approach is not truly reflecting the reality:

-  Outgoing activities should be tracked by `postActivitySaga.ts`
-  Chat adapter/service should not send any activities to Web Chat with `channelData.state`

Thus, we created a new `createDirectLineEmulator`:

-  For static transcript with bot activities only, we can continue to use `createDirectLineWithTranscript`, otherwise;
-  We should use `createDirectLineEmulator` instead

### Fixing intermediate send status

Previously, when an outgoing activity is sent successfully, the `activities` reducer would receive the following actions:

1. `POST_ACTIVITY`, which triggers the `postActivitySaga`
1. `POST_ACTIVITY_PENDING`
1. `INCOMING_ACTIVITY`, when an echo back is received from the chat adapter/service
1. `POST_ACTIVITY_FULFILLED`

When `INCOMING_ACTIVITY` is processed by the `activities` reducer, it upsert an activity into the state with `channelData.state` unset. This is because the incoming activity from the chat adapter/service should not contain this channel data field.

However, after the `INCOMING_ACTIVITY` is handled by the `take` effect in `postActivitySaga`, the `POST_ACTIVITY_PENDING` is dispatched. The `POST_ACTIVITY_PENDING` set the `channelData.state` to `'sent'` for the activity in the `activities` state.

Thus, the `channelData.state` transitioned from `sending` to unset to `sent` in a very short period of time.

The send status for outgoing activities should never be unset.

The following table explains differences between the current `channelData.state` and newer `channelData['webchat:send-status']`, in chronological order:

| # | Action processed | `state` after process | `webchat:send-status` after process |
| - | - | - | - |
| 1 | `POST_ACTIVITY` | (Not in `activities` state) | (Not in `activities` state) |
| 2 | `POST_ACTIVITY_PENDING` | `sending` | `sending` |
| 3 | `INCOMING_ACTIVITY` | Unset | `sending` |
| 4 | `POST_ACTIVITY_FULFILLED` | `sent` | `sent` |

For backward compatibility, the `channelData.state` field has not been updated. Instead, the new `channelData['webchat:send-status']` is updated so it will always have a value for all outgoing activities.

## Specific Changes

(Please see [design](#design) section of this pull request)

## Review Checklist


__tests__/html/activityGrouping.groupingActivityStatus.html
```diff
@@ -187,15 +31,64 @@
 
         await pageConditions.webChatRendered();
 
-        const advanceButton = document.querySelector('.lolex__advance-button');
+        directLine.emulateIncomingActivity('Ullamco mollit sunt officia consectetur ex commodo.');
```
compulim, line 34: It was not exactly correct to emulate outgoing activities by simply using `activity$` to dump things in.

To correctly emulate outgoing activities, it must go through Redux saga. This will capture logic such as send timeout.

__tests__/html/activityStatus.sendFailed.noEchoBack.html
```diff
@@ -0,0 +1,118 @@
...
+          await activityChannelDataWebChatSendStatusBecame('sending');
+
+          // THEN: The message should have status of "Sending".
+          await activityStatusInnerTextContained('Sending');
+
+          // WHEN: After 20 seconds.
+          clock.tick(20000);
+
+          // THEN: `channelData.state` should be "send failed".
+          await activityChannelDataStateBecame('send failed');
```
compulim, line 90: This is the primary reason why we are no longer using the `channelData.state`. It expires after `styleOptions.sendTimeout` (normally 20s) and no way to turn it back.

This means, we cannot revive an activity after it become "send failed".

Thus, in previous version, we simply ignored "send failed" and assume it is "sending", or else, "sent". Thus, we are not recognizing the message failed to send due to reasons other than timeout.

packages/core/src/reducers/activities.ts
```diff
@@ -92,12 +99,9 @@ function upsertActivityWithSort(activities: WebChatActivity[], nextActivity: Web
...
   );
 
-  // Then, find the right (sorted) place to insert the new activity at, based on timestamp
-  // Since clockskew might happen, we will ignore timestamp on messages that are sending
-
+  // Then, find the right (sorted) place to insert the new activity at, based on sequence ID.
   const indexToInsert = nextActivities.findIndex(
-    ({ channelData: { state, 'webchat:sequence-id': sequenceId } = {} }) =>
-      (sequenceId || 0) > (nextSequenceId || 0) && state !== SENDING && state !== SEND_FAILED
+    ({ channelData: { 'webchat:sequence-id': sequenceId } = {} }) => (sequenceId || 0) > (nextSequenceId || 0)
```
compulim, line 104: All activities are already patched with `webchat:sequence-id`. Thus, we can just purely use sequence ID to sort activities, without any concern whether certain activities had been sent or not.

packages/core/src/reducers/activities.ts
```diff
@@ -134,33 +138,91 @@ export default function activities(
...
         activity = updateIn(activity, ['channelData', 'state'], () => SENDING);
+        activity = updateIn(activity, ['channelData', 'webchat:send-status'], () => 'sending');
 
         state = upsertActivityWithSort(state, activity);
       }
 
       break;
 
-    case POST_ACTIVITY_REJECTED:
+    case POST_ACTIVITY_IMPEDED:
```
compulim, line 151: "Impeded" = the activity past the `styleOptions.sendTimeout` period.

This is only needed for backward compatible with previous `channelData.state` field and should be removed after deprecation period.

packages/test/page-object/src/globals/testHelpers/createDirectLineEmulator.js
```diff
@@ -0,0 +1,127 @@
+import createDeferred from 'p-defer';
```
compulim, line 1: Previously, many of our tests relies on `createDirectLineWithTranscript`.

For newer tests which will send outgoing activities, this `createDirectLineEmulator` should be used instead.

CHANGELOG.md
```diff
@@ -33,11 +33,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
    -  Suggested actions container is not longer a live region. The suggested action buttons will now be narrated by the chat history live region
 -  Published NPM packages will now include `/dist`, `/lib`, and `/src` folders
    -  The `/dist` folder was previously missing from our NPM packages
+-  The `activity.channelData.state` property is being deprecated in favor of the updated [`activity.channelData['webchat:send-status']`](https://github.com/microsoft/BotFramework-WebChat/blob/main/packages/core/src/types/WebChatActivity.ts) property, main differences:
```
tdurnford, line 36:
```suggestion
-  The `activity.channelData.state` property is being deprecated in favor of the updated [`activity.channelData['webchat:send-status']`](https://github.com/microsoft/BotFramework-WebChat/blob/main/packages/core/src/types/WebChatActivity.ts) property. Main differences include:
```

CHANGELOG.md
```diff
@@ -33,11 +33,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+   -  Previously, the `state` property would become `"send failed"` when the chat adapter failed to send the activity, or after passing a timeout as defined in `styleOptions.sendTimeout`
```
tdurnford, line 37:
```suggestion
   -  Previously, we would set the `state` property to `"send failed"` when the chat adapter failed to send the activity, or after passing a timeout as defined in `styleOptions.sendTimeout`
```

CHANGELOG.md
```diff
@@ -33,11 +33,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+   -  The new `'webchat:send-status'` property will become `"send failed"` when the chat adapter failed to send the activity, or after passing a hardcoded timeout of 5 minutes
```
tdurnford, line 38:
```suggestion
   -  The new `'webchat:send-status'` property will be set to `"send failed"` when the chat adapter fails to send the activity or after passing a hardcoded timeout of 5 minutes
```

docs/HOOKS.md
```diff
@@ -1067,6 +1068,24 @@ This hook will return whether the typing indicator will be sent to the bot when
...
+
+This hook will return send status of all outgoing activities, either `"sending"`, `"send failed"`, or `"sent"`. The key of the `Map` is activity key. This is different than activity ID and can be obtained via the [`useGetKeyByActivity`](#usegetkeybyactivity) or [`useGetKeyByActivityId`](#usegetkeybyactivityid) hooks.
+
+`"send failed"` is not a terminator state. An activity marked as `"send failed"` could become other state again. However, `"sent"` is a terminator state.
+
+If `styleOptions.sendTimeout` or `styleOptions.sendTimeoutForAttachments` increased, an activity previously marked as `"send failed"` (due to timeout) could become `"sending"` again.
+
+If the activity failed to send (`"send failed"`), retrying the send will also change the send status back to `"sending"`.
+
+Send status is based on clock and could change very frequently. This will incur performance cost. Please use this hook deliberately.
```
tdurnford, line 1087: Based on clock?

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
```diff
@@ -1,104 +1,107 @@
...
-    // If no timestamp, we assume the "sending" will be timed out as "send failed".
-    sendTimeoutAt = !activitySent
-      ? new Date(activity.localTimestamp || new Date(0).toISOString()).getTime() + sendTimeout
-      : 0;
-  } else {
-    activitySent = true;
-    sendTimeoutAt = 0;
+const ActivityStatusContainerCore: VFC<InnerActivityStatusContainerProps> = memo(
+  ({ activity, hideTimestamp, nextVisibleActivity, sendStatus }) => {
+    const { activityStatusRenderer: createActivityStatusRenderer }: { activityStatusRenderer: RenderActivityStatus } =
```
tdurnford, line 25: Do we need to specify the type here? Is `useWebChatAPIContext` untyped?

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
```diff
@@ -1,104 +1,107 @@
...
+      useWebChatAPIContext();
+
+    return createActivityStatusRenderer({
+      activity,
+      hideTimestamp,
+      nextVisibleActivity, // "nextVisibleActivity" is for backward compatibility, please remove this line on or after 2022-07-22.
+      sameTimestampGroup: hideTimestamp, // "sameTimestampGroup" is for backward compatibility, please remove this line on or after 2022-07-22.
+      sendState: sendStatus === 'send failed' || sendStatus === 'sent' ? sendStatus : 'sending'
```
tdurnford, line 33: Is `sendStatus` typed? Should we use constants here?

```suggestion
      sendState: sendStatus === SEND_FAILED || sendStatus === SENT ? sendStatus : SENDING
```

tdurnford (dismissed):

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
compulim: Trying to type out `useWebChatAPIContext` and it break something else.

The water bucket is leaking... I will tackle it in a new issue #4367.

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
compulim: This is actually intentional. Feel free to comment:

- Let's stay away from the `Constants.SendState` thing we introduced before TypeScript
   - The constant can be replaced/typechecked by TypeScript without worry about typo
- I am unsure if we should go `enum ABC` or just `type ABC = 'one' | 'two' | 'three'`

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
```diff
@@ -1,104 +1,107 @@
...
 };
 
+const ActivityStatusContainer: VFC<ActivityStatusContainerProps> = memo(
+  ({ activity, hideTimestamp, nextVisibleActivity }) => {
+    const [sendStatusByActivityKey] = useSendStatusByActivityKey();
+    const getKeyByActivity = useGetKeyByActivity();
+
+    const key = getKeyByActivity(activity);
+
+    const sendStatus = (typeof key === 'string' && sendStatusByActivityKey.get(key)) || 'sent';
```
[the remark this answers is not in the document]
compulim: (Intentionally not using the older `Constants.SendState`.)

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
compulim: Fixed by using internal constant.

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
```diff
@@ -1,104 +1,107 @@
...
 ActivityStatusContainer.propTypes = {
   // PropTypes cannot fully capture TypeScript types.
   // @ts-ignore
   activity: PropTypes.shape({
-    channelData: PropTypes.shape({ state: PropTypes.string }),
     from: PropTypes.shape({ role: PropTypes.string }).isRequired,
     localTimestamp: PropTypes.string
   }).isRequired,
-  hideTimestamp: PropTypes.bool,
+  hideTimestamp: PropTypes.bool.isRequired,
```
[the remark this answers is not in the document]
compulim: Right, we should leave it `undefined`.

packages/api/src/hooks/useCreateActivityStatusRenderer.tsx
tdurnford: I'd prefer `type ABC = 'one' | 'two' | 'three'` :)

tdurnford (approved):

PR #4420, opened 2022-09-20
Fix ARIA role of transcript when connection is pending

### LINKED ISSUE #4393: [Accessibility] WCAG 1.3.1: Ensures elements with an ARIA role that require child roles contain them

- [ ] Yes, this is an Adaptive Card issue but it is specific to Web Chat.

### Do you have any screenshots?

image

Latest production

Bundle (webchat.js)

Web apps

Browser: Chrome (latest), Platform: Windows

Chat history

- [X] Yes, this is an accessibility issue.

### Please describe the bug

WCAG 1.3.1: Ensures elements with an ARIA role that require child roles contain them (.webchat--css-cujit-1l95nvm)

Initially, the chat history is empty with no messages. Thus, it has `role="feed"` but does not have any children marked as `role="article"` or `<article>`.

Thus, it violated WCAG 1.3.1.

_No response_

1. Talk to MockBot
2. Run "Accessibility Insights" with "FastPass"

### What is the expected and actual behavior?

Expected:
- No accessibility issues found by FastPass

Actual:
- FastPass is complaining about WCAG 1.3.1

### Adaptive Card JSON

_No response_

### Additional context

There are few potential solutions:

- Adds an invisible (screen reader only) label of `<article>` saying "Chat history is empty"
- Sets `role="feed"` only if the chat history contains 1 or more messages, leave the `role` attribute empty otherwise
- Sets `aria-busy="true"` if the chat history is empty

### Readings

- [Understanding Success Criterion 1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)

beyackle2:
> Fixes #4393.

### Fixed

-  Fixes [#4393](https://github.com/microsoft/BotFramework-WebChat/issues/4393). Renders `<section role="feed">` only if there are one or more activities contained within, by [@beyackle2](https://github.com/beyackle2) and [@compulim](https://github.com/compulim), in PR [#4420](https://github.com/microsoft/BotFramework-WebChat/pull/4420)

## Description

The `<section role="feed">` should not be emptied and should have at least one `<article>` or `role="article"`.

When the chat is started, it is initially empty. We should not render `<section role="feed">`.

## Design

We tried `<section role="feed" aria-busy="true">`, however, Accessibility Insights still say the element is violating "WCAG 1.3.1: Ensures elements with an ARIA role that require child roles contain them".

We have 2 options:

1. Render `<section>` without `role` attribute, and set `role="feed"` when it has contents
1. Do not render `<section role="feed">` if it is empty

We went on the second option as we worry the first one might cause issues in some screen readers.

## Specific Changes

- Not rendering `<section role="feed">` when it is empty

## Review Checklist


packages/component/src/BasicTranscript.tsx
```diff
@@ -869,7 +882,7 @@ type BasicTranscriptProps = {
   className?: string;
 };
 
-const BasicTranscript: VFC<BasicTranscriptProps> = ({ className }) => {
+const BasicTranscript: React.FC<BasicTranscriptProps> = ({ className }) => {
```
tdurnford, line 885:
```suggestion
const BasicTranscript: FC<BasicTranscriptProps> = ({ className }) => {
```

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+-  Fixes [#4393](https://github.com/microsoft/BotFramework-WebChat/issues/4393). Assigns correct ARIA roles to components in web chat when a conversation status is pending, by [@beyackle2](https://github.com/beyackle2), in PR [#4420](https://github.com/microsoft/BotFramework-WebChat/pull/4420)
```
tdurnford, line 25:
```suggestion
### Fixed

-  Fixes [#4393](https://github.com/microsoft/BotFramework-WebChat/issues/4393). Assigns correct ARIA roles to components in web chat when a conversation status is pending, by [@beyackle2](https://github.com/beyackle2), in PR [#4420](https://github.com/microsoft/BotFramework-WebChat/pull/4420)
```

beyackle2: I wasn't sure where this one belonged, since the section above "Fixed" is "Unreleased", and this change won't be released immediately.

packages/component/src/BasicTranscript.tsx
```diff
@@ -508,11 +509,16 @@ const InternalTranscript = forwardRef<HTMLDivElement, InternalTranscriptProps>(
...
+    const [{ connectivitystatus: connectivityStatus }] = useDebouncedNotifications();
+    const statusMessage = connectivityStatus?.message;
+    const isWaiting = !statusMessage || statusMessage === 'connecting' || statusMessage === 'reconnecting';
+
     return (
       <div
         // Although Android TalkBack 12.1 does not support `aria-activedescendant`, when used, it become buggy and will narrate content twice.
         // We are disabling `aria-activedescendant` for Android. See <ActivityRow> for details.
         aria-activedescendant={android ? undefined : activeDescendantId}
+        aria-busy={isWaiting}
```
compulim, line 521: Do we need `aria-busy` for `role="group"`? Or just the `role="feed"`?

If we only need it for `role="feed"`, then we can put the `useDebouncedNotifications()` inside `<InternalTranscriptScrollable>`.

packages/component/src/BasicTranscript.tsx
```diff
@@ -508,11 +509,16 @@ const InternalTranscript = forwardRef<HTMLDivElement, InternalTranscriptProps>(
       useCallback(() => focusByActivityKey(undefined), [focusByActivityKey])
     );
 
+    const [{ connectivitystatus: connectivityStatus }] = useDebouncedNotifications();
+    const statusMessage = connectivityStatus?.message;
+    const isWaiting = !statusMessage || statusMessage === 'connecting' || statusMessage === 'reconnecting';
```
compulim, line 514: Consider renaming it to `ariaBusy`.

packages/component/src/BasicTranscript.tsx
```diff
@@ -605,13 +615,15 @@ type InternalTranscriptScrollableProps = {
...
 };
 
 // Separating high-frequency hooks to improve performance.
 const InternalTranscriptScrollable: FC<InternalTranscriptScrollableProps> = ({
   children,
   onFocusFiller,
-  terminatorRef
+  terminatorRef,
+  isWaiting
 }) => {
```
compulim, line 627: Sorting.

```ts
const InternalTranscriptScrollable: FC<InternalTranscriptScrollableProps> = ({
  children,
  isWaiting,
  onFocusFiller,
  terminatorRef,
}) => {
```

(GitHub didn't allow me to use code suggestions.)

packages/component/src/BasicTranscript.tsx
```diff
@@ -731,7 +743,7 @@ const InternalTranscriptScrollable: FC<InternalTranscriptScrollableProps> = ({
         <section
           aria-roledescription={transcriptRoleDescription}
           className={classNames(activitiesStyleSet + '', 'webchat__basic-transcript__transcript')}
-          role="feed"
+          role={isWaiting ? undefined : 'feed'}
```
compulim, line 746: Should we set `aria-busy` than modifying the `role`? I am worried by how screen reader behave when changing role.

packages/component/src/BasicTranscript.tsx
```diff
@@ -743,6 +755,7 @@ const InternalTranscriptScrollable: FC<InternalTranscriptScrollableProps> = ({
 
 InternalTranscriptScrollable.propTypes = {
   children: PropTypes.any.isRequired,
+  isWaiting: PropTypes.bool.isRequired,
```
compulim, line 758: Maybe rename it to `aria-busy`.

```suggestion
  'aria-busy': PropTypes.bool.isRequired,
```

packages/component/src/BasicTranscript.tsx
beyackle2: That's a good idea. Reading over what `aria-busy` does, this looks like it would make it do exactly what we want.

packages/component/src/BasicTranscript.tsx
beyackle2: Fixed in latest commit.

packages/component/src/BasicTranscript.tsx
beyackle2: I've looked it up and `aria-busy` doesn't specify which roles it should or shouldn't be attached to. Since this div covers the whole transcript, I think it does make sense to mark it as busy when chat is loading.

packages/component/src/BasicTranscript.tsx
beyackle2: Done.

CHANGELOG.md
compulim: I think when TJ added his comment, this `### Fixed` was not in `CHANGELOG.md`.

A (few) recent PRs added the `### Fixed` section.

packages/component/src/BasicTranscript.tsx
```diff
@@ -602,13 +612,15 @@ InternalTranscript.propTypes = {
 };
 
 type InternalTranscriptScrollableProps = {
+  ariaBusy: boolean;
```
compulim, line 615:
```suggestion
  'aria-busy': boolean;
```

compulim (commented):
Need tests and this will be puuuuurfect. 😽

packages/component/src/BasicTranscript.tsx
beyackle2: Fixed, though it took a little more rearranging to make the linter happy.

compulim:
image

Looks like `aria-busy="true"` won't help with the accessibility issue. Few things we could try:

- Change `role` to something else until the first message come
- Inject an invisible "the chat history is empty" in the transcript before the first message

compulim:
@beyackle2 Few things I changed:

- Now, it renders `<section role="feed">` if there are any contents
- For tests, I am adding this DOM integrity test to `verifyDOMIntegrity.js`, which is a tiny version of `axe-core`
   - Ultimately, we should use `axe-core` and replace `verifyDOMIntegrity.js`
   - However, `axe-core` is using a non-MIT license so we need to do something extra before using it
   - Thus, we have the tiny `verifyDOMIntegrity.js` as a stopgap for now

tdurnford (approved):

PR #5135, opened 2024-04-15
Fluent: migrate to CSS Modules

OEvgeny:
## Changelog Entry

(No CHANGELOG for tooling update)

## Description

This PR removes emotion from the `fluent-theme` package to:
- Reduce bundled size. The package went from ~70+ KB down to ~41 KB
- Reduce runtime complexity: emotion runs in runtime and while it's fast, not having it should be even faster
- Unlock native CSS authoring capabilities via CSS modules

## Design

For now we don't introduce any additional tools to the build pipeline. The `tsup` bundler uses `esbuild` which is able to bundle CSS modules when configured to use `local-css` loader. We now use this approach to bundle CSS into a single file.

After the file is produced, we're using `build.onEnd` hook to inject the styles as a string into the JavaScript bundle.

During script/module initialization, we inject the string into a style tag, and inject the tag into the `document.head` only if the document is available.

Styles are available internally, but not yet exposed in the package exports.

For types inference from the CSS modules, the `typescript-plugin-css-modules` was added.

## Specific Changes

- Move all styles to the corresponding `.module.css` files
- Add `injectCSSPlugin` esbuild plugin and make related changes
- Configure esbuild to bundle CSS modules
- Fix workspace typescript path
- Unprefix CSS rules, use power of CSS to namespace styles
  * Before `.webchat-fluent__sendbox__attachment-drop-zone`
  * After: `:global(.webchat-fluent) .sendbox__attachment-drop-zone`
- Replace previous `useStyles` hook with the new one handling CSS modules
- Add `*.module.css` type definitions
- Add `typescript-plugin-css-modules` TypeScript plugin

## Review Checklist


package.json
```diff
@@ -176,6 +176,7 @@
     "serve-handler": "^6.1.5",
     "strip-ansi": "^6.0.1",
     "typescript": "^5.3.2",
+    "typescript-plugin-css-modules": "^5.1.0",
```
OEvgeny, line 179: Had to add the package to the root dependencies as I discovered TypeScript tries to find it there

packages/fluent-theme/src/components/Theme.module.css
```diff
@@ -0,0 +1,60 @@
+
+:global(.webchat-fluent).theme {
+  display: contents;
+
+  --webchat-colorNeutralForeground1: var(--colorNeutralForeground1, #242424);
```
compulim, line 5: Maybe this one later:

```suggestion
  --webchat--colorNeutralForeground1: var(--colorNeutralForeground1, #242424);
```

Or...

```suggestion
  --webchat-fluent--colorNeutralForeground1: var(--colorNeutralForeground1, #242424);
```

packages/fluent-theme/src/components/sendbox/ErrorMessage.module.css
```diff
@@ -0,0 +1,10 @@
+
```
compulim, line 1: No empty new line here.

packages/fluent-theme/src/styles/useStyles.ts
```diff
@@ -0,0 +1,14 @@
+import { useMemo } from 'react';
+
+function useStyles<T extends CSSModuleClasses>(styles: T): T {
+  // @ts-expect-error: entries/fromEntries don't allow to specify keys type
+  return useMemo(
+    () =>
+      Object.freeze(
+        Object.fromEntries(Object.entries(styles).map(([baseCls, resultCls]) => [baseCls, `${baseCls} ${resultCls}`]))
```
compulim, line 8:
```suggestion
        Object.fromEntries(Object.entries(styles).map(([baseClassName, resultClassName]) => [baseClassName, `${baseClassName} ${resultClassName}`]))
```

packages/fluent-theme/src/styles/injectStyle.ts
```diff
@@ -0,0 +1,9 @@
+export const injectedStyles = '@--INJECTED-STYLES-CONTENT--@';
+
+export default function injectStyles() {
+  if (globalThis.document) {
+    const style = document.createElement('style');
```
compulim, line 5: We will need to add `nonce` attribute.

I expect a prop from the upstreamer.

```ts
root.render(
  <FluentThemeProvider nonce="a1b2c3d">
     ...
  </FluentThemeProvider>
);
```

And then we grab the nonce and inject it here.

You can do it in another PR.

packages/fluent-theme/tsup.config.ts
```diff
@@ -21,10 +22,28 @@ const umdResolvePlugin = {
   }
 };
 
+const injectCSSPlugin = {
+  name: 'inject-css-plugin',
+  setup(build) {
+    build.onEnd(result => {
+      const js = result.outputFiles.find(f => f.path.match(/(\.js|\.mjs)$/u));
+      const css = result.outputFiles.find(f => f.path.match(/(\.css)$/u));
+      if (js && css && js.text.includes(injectedStylesPlaceholder)) {
```
compulim, line 31:
```suggestion
      if (css && js?.text.includes(injectedStylesPlaceholder)) {
```

packages/fluent-theme/tsup.config.ts
```diff
@@ -21,10 +22,28 @@ const umdResolvePlugin = {
...
+        const cssText = css.text;
+        js.contents = Buffer.from(js.text.replace(`"${injectedStylesPlaceholder}"`, JSON.stringify(cssText)));
```
compulim, line 33:
```suggestion
        js.contents = Buffer.from(js.text.replace(`"${injectedStylesPlaceholder}"`, JSON.stringify(css.text)));
```

compulim (approved):
Approve with comments.

packages/fluent-theme/src/components/Theme.module.css
OEvgeny: Let's address in the next PR

packages/fluent-theme/src/styles/injectStyle.ts
OEvgeny: Sure, what should it do?

compulim (approved):

PR #5136, opened 2024-04-17
Fluent: use grid for sendbox layout

OEvgeny:
## Changelog Entry

-  (Experimental) Added `botframework-webchat-fluent-theme` package for applying Fluent UI theme to Web Chat, by [@compulim](https://github.com/compulim) and [@OEvgeny](https://github.com/OEvgeny)
   -  Disabled send button and hid message length when telephone keypad is shown, in PR [#5136](https://github.com/microsoft/BotFramework-WebChat/pull/5136)

## Description

The PR enables advanced layout for the experimental Fluent send box experience, also fixing some small issues found along the way.

## Design

We now use named areas to manipulate the positioning for `telephone-keypad`, `text-area`, `attachment` and `controls`. These areas can then be remapped using CSS to achieve different layouting options.

To avoid double gaps we conditionally add the `attachment` area only when the corresponding to the area class is present.

## Specific Changes

-  Disable `Send` button when telephone keypad is shown
-  Hide message length when telephone keypad is shown
-  Rework sendbox grid implementation
-  Add tests

## Review Checklist

### DIFF
--- a/packages/fluent-theme/src/components/sendbox/index.module.css
+++ b/packages/fluent-theme/src/components/sendbox/index.module.css
@@ -12,14 +12,14 @@
   border-radius: var(--webchat-borderRadiusLarge);
   border: 1px solid var(--webchat-colorNeutralStroke1);
   display: grid;
+  font-family: var(--webchat-fontFamilyBase);
+  font-size: 14px;
   gap: 6px;
   grid-template:
     [telephone-keypad-start] 'text-area'
     var(--webchat-sendbox-attachment-area-active) [telephone-keypad-end]
     'controls' / [telephone-keypad] 1fr
   ;
-  font-family: var(--webchat-fontFamilyBase);
-  font-size: 14px;
   line-height: 20px;
   padding: 8px;
   position: relative;


packages/fluent-theme/src/components/sendbox/index.module.css
```diff
@@ -3,20 +3,31 @@
...
+    'controls' / [telephone-keypad] 1fr
+  ;
   font-family: var(--webchat-fontFamilyBase);
   font-size: 14px;
-  gap: 6px;
   line-height: 20px;
   padding: 8px;
   position: relative;
 
+  &:has(.sendbox__attachment--in-grid) {
```
compulim, line 27: Instead of `in-grid`, should we call it `visible`, `shown`, `hidden` (the other way) instead?

packages/fluent-theme/src/components/sendbox/index.module.css
```diff
@@ -3,20 +3,31 @@
...
+
+  --webchat-sendbox-attachment-area-active: ;
 }
 
 :global(.webchat-fluent) .sendbox__sendbox {
   background-color: var(--webchat-colorNeutralBackground1);
   border-radius: var(--webchat-borderRadiusLarge);
   border: 1px solid var(--webchat-colorNeutralStroke1);
   display: grid;
+  gap: 6px;
```
compulim, line 15: Sort.

packages/fluent-theme/src/components/sendbox/index.module.css
```diff
@@ -25,6 +36,22 @@
       - we are using colorCompoundBrandForeground1Hover */
     box-shadow: inset 0 -6px 0 -3px var(--webchat-colorCompoundBrandForeground1Hover);
   }
+
+  > .sendbox__text-area--in-grid {
```
compulim, line 40: Maybe we don't need "immediate child" because we are BEM?

```suggestion
  .sendbox__text-area--in-grid {
```

packages/fluent-theme/src/components/sendbox/index.tsx
```diff
@@ -119,24 +119,25 @@ function SendBox(
...
       <div className={cx(classNames['sendbox__sendbox'])} onClickCapture={handleSendBoxClick}>
-        <TelephoneKeypadSurrogate
-          autoFocus={true}
-          isHorizontal={false}
-          onButtonClick={handleTelephoneKeypadButtonClick}
-        />
         <TextArea
           aria-label={isMessageLengthExceeded ? localize('TEXT_INPUT_LENGTH_EXCEEDED_ALT') : localize('TEXT_INPUT_ALT')}
-          className={classNames['sendbox__sendbox-text']}
+          className={cx(classNames['sendbox__sendbox-text'], classNames['sendbox__text-area--in-grid'])}
```
compulim, line 124: If we always apply `--in-grid`, why don't we merge both class together?

packages/fluent-theme/src/components/sendbox/index.module.css
OEvgeny: We want to know that not only the element is visible, but the corresponding area is in use

packages/fluent-theme/src/components/sendbox/index.module.css
OEvgeny: We want to indicate that this selector will work only on direct children

packages/fluent-theme/src/components/sendbox/index.tsx
OEvgeny: You mean just removing the `--in-grid` prefix?
E.g.:
```
  > .sendbox__attachment {
    grid-area: attachment;
  }

  > .sendbox__sendbox-controls {
    grid-area: controls;
  }

  > .sendbox__telephone-keypad {
    grid-area: telephone-keypad;
  }
```

According to BEM this is essentially a modifier for the components to enable grid, so it uses the modifier postfix

packages/fluent-theme/src/components/sendbox/index.module.css
compulim: It's good. I thought if it will be too fragile. I guess not because this is a grid and supposedly to be direct children.

compulim (approved):

PR #5172, opened 2024-05-07
Perf: allow useMemoAll to cache results independently from deps

OEvgeny:
> Fixes #

## Changelog Entry

- Improved performance for `useActivityWithRenderer`, in [PR], by [@OEvgeny](https://github.com/OEvgeny)

## Description

Thanks to the optimization, now subsequent activity re-renders are cached. The downside is that the `useMemoAll` hook cached results are not discarded when `useMemo` dependencies are updated.

## Design

Initially the implementation was caching functions globally (see the [first commit](https://github.com/microsoft/BotFramework-WebChat/commit/2dd2fd74eb11da4dedaced1d530ae205b37b2096)). Now we cache per hook call to better align with `useMemo`.

Note that if we aim to maintain full parity with `useMemo`, we shouldn't make this optimisation. But it'd make the implementation similar to defining and using the memoized function directly in `useMemo` hook making this hook obsolete.

The updated implementation is rather not as trivial and solves use-cases for caching results even after dependencies change,

## Specific Changes

- Updated `useMemoAll` hook implementation
- Renamed: `useMemoize` -> `useMemoAll` as suggested
- Updated tests
- Added activity render count test

## -

-

## Review Checklist


packages/component/src/hooks/internal/useMemoAll.spec.jsx
```diff
@@ -0,0 +1,155 @@
...
+
+  expect(expensiveSum).toHaveBeenCalledTimes(3);
+  expect(expensiveSum.mock.calls[2]).toEqual([3, 6]);
+
+  // After the run, only 1 + 2 = 3 and 3 + 6 = 9 is cached. 2 + 4 is dropped.
+
+  // Start another run with previous cache
+  await render(
+    sum => {
+      expect(sum(2, 4)).toBe(6); // Not cached, return 6
```
compulim: Because 2 + 4 was dropped in previous run, this seems inconclusive that the deps is causing cache to drop.

Maybe, `expect(sum(1, 2)).toBe(3)` and it should call it the 4th times.

packages/component/src/hooks/internal/useMemoAll.spec.jsx
```diff
@@ -0,0 +1,146 @@
...
+  });
+
+  expect(expensiveSum).toHaveBeenCalledTimes(3);
+  expect(expensiveSum.mock.calls[2]).toEqual([3, 6]);
+
+  // After the run, only 1 + 2 = 3 and 3 + 6 = 9 is cached. 2 + 4 is dropped.
+
+  // Start another run with previous cache
+  await render(sum => {
+    expect(sum(2, 4)).toBe(6); // Not cached, return 6
```
compulim, line 116: Because 2 + 4 was dropped during previous run (1 + 2 and 3 + 6), it is inconclusive that 2 + 4 is not cached due to deps change.

Maybe try 1 + 2 or 3 + 6, which was not dropped.

__tests__/html/renderActivity.performance.html
```diff
@@ -0,0 +1,107 @@
...
+              { skipWait: true }
+            );
+
+          // THEN: activity re-renders the activity
+          expect(timesActivityRendered.get('activity-8')).toBe(activity9Renders + 1);
+          // THEN: activity gets updated on screen
+          pageElements.focusedActivity().scrollIntoView();
+          await host.snapshot();
+        },
+        { skipCheckAccessibility: true }
```
compulim, line 103: We should keep check accessibility unless we know something would definitely break.

compulim:
Literally we need to select one of 2 routes:

* `useMemo(fn, deps)`, which change to `fn` will not re-run. Change to `deps` will cause a re-run
* `useSyncExternalStore(subscribe, getSnapshot)`, which change to `subscribe` or `getSnapshot` will always cause a re-run

I think we should follow `useMemo`.

What I observed from React API:

* `useSyncExternalStore` do not accept `DependencyList` so it invalidate _solely_ based on changes to `subscribe` and `getSnapshot`
* In contrast, `useMemo` accept `DependencyList` so it invalidate _solely_ based on `deps` and ignore changes to `fn`

OEvgeny:
Instead of `useMemoAll` we probably could go with `useMemo` plus the memoized function to retrieve activity related data.

packages/component/src/providers/ActivityTree/private/useActivitiesWithRenderer.ts
```diff
@@ -43,7 +44,7 @@ export default function useActivitiesWithRenderer(
 
       return Object.freeze(activitiesWithRenderer);
     },
-    [activities]
+    [createActivityRenderer, activities]
```
compulim, line 47:
```suggestion
    [activities, createActivityRenderer]
```

compulim (approved):
Nitpicking.

compulim (approved):

PR #5176, opened 2024-05-08
Add base package

OEvgeny:
## Changelog Entry

(No changelog entry for the internal change)

## Description

This PR introduces a new private `botframework-webchat-base` package to centralize shared utilities and types across the Web Chat project. It also updates the build process for all packages to use `tsup`, as otherwise the reference to the package is included into depending packages by babel.

## Design

The new base package serves as a foundation for shared functionality, promoting code reuse and maintainability. The build process changes aim to simplify and standardize the compilation across all packages, reducing complexity and potential inconsistencies.

### Limitations

Due to how babel build works, only the packages which are using the new build approach powered by `tsup` can consume the base and styles packages. This limitation applies only to the packages we're publishing.

#### Package Status List

- 🚀 base
- 🔒 test/dev-server
- 🔒 test/harness
- 🔒 test/web-server
- ✅ core
- ✅ styles
- 🔒 support/cldr-data-downloader
- 🔒 support/cldr-data
- ✅ api
- 🔒 isomorphic-react
- 🔒 isomorphic-react-dom
- ✅ component
- ⚠️ directlinespeech
- ✅ bundle
- 🔒 test/page-object
- ✅ fluent-theme
- 🔒 test/fluent-bundle

#### Legend

- 🚀 New package with `tsup` introduced in this PR
- ✅ Existing package migrated to `tsup` in this PR
- ⚠️ Package needs update to `tsup` to use the base package
- 🔒 Private package (not published)

## Specific Changes

- Added new `packages/base` directory with initial setup
- Moved `withResolvers` utility from `core` to `base` package
- Updated `package.json` files across multiple packages:
  - Added `botframework-webchat-base` as a development dependency
  - Updated build scripts to use `tsup`
  - Removed Babel and separate TypeScript compilation steps
- Updated `tsconfig.json` files to remove emission-related options
- Modified import statements in affected files to use the new base package
- Updated webpack configurations to reflect new build output structure
- Adjusted GitHub Actions workflow to use new build output paths

## -

## Review Checklist


packages/base/babel.config.json
compulim, line 1: Maybe we don't need `babel` to emit CJS and keep only `tsup` for ESM as we are bundling into other packages?

We need to think how to bundle in other packages... say, CJS in `api` package is under `/lib` and how can we "bundle" there...?

Or are we brave enough to jump into `tsup` for both CJS or ESM. 🤷🏻‍♂️

OEvgeny: No need to ask twice.

Removed babel, and made tsup to emit both cjs and esm. As we don't have babel we don't have `lib` folder. In addition now typescript is used during build/start to only report issues as tsup sometimes doesn't provide a clear output on failures.

packages/base/package.json
compulim, line 1: Also mark this package as `private: true` so we don't accidentally publish it.

compulim:
Let's make this `base` package internal and bundled into other packages.

For a cleaner approach, we should use `tsup` for CJS build and no longer need Babel.

packages/component/package.json
```diff
@@ -131,6 +127,8 @@
     "@types/react": "^16.14.60",
     "babel-plugin-istanbul": "^6.1.1",
     "babel-plugin-transform-inline-environment-variables": "^0.4.4",
+    "botframework-webchat-base": "0.0.0-0",
```
compulim, line 130: Double verify, are we sure `-base` and `-styles` should be development dependencies?

packages/core/src/actions/connect.ts
compulim, line 1: Can we keep it as `.js` until we add types to it?

packages/component/package.json
compulim: I like both of them bundled, I just want to verify if the code is good.

packages/styles/src/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "moduleResolution": "Bundler",
     "noEmit": true,
     "skipLibCheck": true,
-    "target": "ESNext"
+    "target": "ESNext",
```
compulim, line 9: May not need this.

compulim (dismissed):

packages/component/package.json
OEvgeny: Otherwise npm won't allow to install packages depending on them. We either have to bundle or publish it seems.

packages/core/src/actions/connect.ts
OEvgeny: Unfortunately there is a problem with generating d.ts from js files. It only happens for CJS, so I had to update extensions to avoid broken .d.ts for CJS.

packages/styles/src/tsconfig.json
OEvgeny: This may be required though. By telling TS we're aiming for the latest supported by TS version, we ensure it doesn't assume any polyfills to be added.

packages/component/package.json
compulim: Bundle is good, not too worry about code dupe yet, it's better than today.

Also no need to update package dependency is good.

packages/core/src/actions/connect.ts
compulim: Oh wow, you figured it out. ❤️

packages/styles/src/tsconfig.json
compulim: I mean the extra comma. 🤣

OEvgeny, line 9:
```suggestion
    "target": "ESNext"
```

OEvgeny: Fixed

compulim (approved):

PR #5183, opened 2024-05-10
Perf: Transcript and Activity render speed

OEvgeny:
> Fixes #

## Description

Improve performance by more carefully memoizing activity-related components.

## -

-

## Review Checklist

### DIFF
--- a/packages/component/src/BasicTranscript.tsx
+++ b/packages/component/src/BasicTranscript.tsx
@@ -11,7 +11,7 @@ import {
 } from 'react-scroll-to-bottom';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import React, { forwardRef, Fragment, useCallback, useMemo, useRef } from 'react';
+import React, { forwardRef, Fragment, memo, useCallback, useMemo, useRef } from 'react';
 
 import type { ActivityComponentFactory } from 'botframework-webchat-api';
 import type { ActivityElementMap } from './Transcript/types';
@@ -48,6 +48,7 @@ import useStyleSet from './hooks/useStyleSet';
 import useStyleToEmotionObject from './hooks/internal/useStyleToEmotionObject';
 import useUniqueId from './hooks/internal/useUniqueId';
 import useValueRef from './hooks/internal/useValueRef';
+import useMemoAll from './hooks/internal/useMemoAll';
 
 const {
   useActivityKeys,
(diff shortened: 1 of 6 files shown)


packages/component/src/Transcript/ActivityRow.tsx
```diff
@@ -56,6 +56,15 @@ const ActivityRow = forwardRef<HTMLLIElement, ActivityRowProps>(({ activity, chi
   // We are doing it in event capture phase to prevent descendants from stopping event propagation to us.
   const handleMouseDownCapture: MouseEventHandler = useCallback(() => focusSelf(false), [focusSelf]);
 
+  const focusTrapChildren = useMemo(
+    () => (
+      <div className="webchat__basic-transcript__activity-body" ref={bodyRef}>
+        {children}
+      </div>
+    ),
+    [children]
```
OEvgeny, line 65:
```suggestion
    [bodyRef, children]
```

packages/component/src/Transcript/FocusTrap.tsx
```diff
@@ -69,4 +69,4 @@ FocusTrap.propTypes = {
   onLeave: PropTypes.func.isRequired
 };
```
OEvgeny, line 71:
```suggestion
FocusTrap.displayName = 'FocusTrap';

```

packages/component/src/BasicTranscript.tsx
```diff
@@ -155,15 +142,20 @@ const InternalTranscript = forwardRef<HTMLDivElement, InternalTranscriptProps>(
...
 
+    const createAvatarRendererMemoized = useMemoized(
+      (activity: WebChatActivity) => createAvatarRenderer({ activity }),
+      [createAvatarRenderer]
+    );
+
     // Flatten the tree back into an array with information related to rendering.
     const renderingElements = useMemo(() => {
-      const renderingElements: RenderingElement[] = [];
+      const renderingElements: Array<ReactNode> = [];
```
compulim, line 152: Should we use...

```suggestion
      const renderingElements: ReactNode[] = [];
```

I am always confused when to use `Array<T>` vs. `T[]`.

packages/component/src/BasicTranscript.tsx
```diff
@@ -200,27 +188,23 @@ const InternalTranscript = forwardRef<HTMLDivElement, InternalTranscriptProps>(
...
-                hideAllTimestamps || indexWithinSenderAndStatusGroup !== entriesWithSameSenderAndStatus.length - 1,
-              key,
-              renderActivity,
-              renderActivityStatus,
-              renderAvatar,
-              showCallout
-            });
+            renderingElements.push(
+              <TranscriptActivity
+                activity={activity}
```
compulim, line 193: I like that there are clear "setup-time props" and "render-time props" difference now. 😄

compulim (commented):
Excellent. You now know the internals of our chat history. It's one of the most complex parts of Web Chat.

Let's timebox some time and understand why some snapshots need to be updated. Our snapshots tell us a lot more than we understand. They are our teachers. 👨🏻‍🏫

__tests__/html/fluentTheme/attachMultipleFiles.html
```diff
@@ -63,9 +63,10 @@
           { type: 'image/jpeg' }
         );
 
-        await pageObjects.runHook(({ useSendBoxAttachments }) =>
-          useSendBoxAttachments()[1]([{ blob: blob1 }, { blob: blob2 }])
-        );
+        await pageObjects.runHook(({ useSendBoxAttachments }) => {
```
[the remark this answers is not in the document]
OEvgeny: This popped-up as an infinite re-render loop after my changes. I used `ref`, so we always pass the stable value to the hook.

Took some time to investigate, and I don't see this as a bug in new rendering process, rather we just hit the issue after touching how rendering works.

Perhaps we could improve `useSendBoxAttachments`, but overall I believe it should receive stable inputs.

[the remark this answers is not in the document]
compulim: Oh you are right. The `runHook` will render a component (to run the hook). Then the `useSendBoxAttachments()` triggered a state change and re-render. So the `<RunHook>` get re-rendered again.

PR #5190, opened 2024-05-17
Perf: improve useMemoized callback execution time

OEvgeny:
## Changelog Entry

- Improved performance for `useMemoized`, in PR [5190](https://github.com/microsoft/BotFramework-WebChat/pull/5190), by [@OEvgeny](https://github.com/OEvgeny)

## Description

This PR reduces time spent in the memoized function created by `useMemoized` by 3-18 times. As this function is one of the primitives we use to improve performance, we should keep it as fast and simple as possible.

## Design

The main idea is to reduce unnecessary object allocations and reuse existing objects where possible.

Changes are:
- use a single lookup instead of combined array lookup
- reuse cached object if found
- do not create unnecessary objects
- explicit branching for cached / non-cached conditions

## Specific Changes

- Updated `memoizedFn` implementation

## -

-  [x] ~I have added tests and executed them locally~

## Review Checklist

-  [x] ~Tests reviewed (coverage, legitimacy)~

### DIFF
--- a/packages/component/src/hooks/internal/useMemoized.ts
+++ b/packages/component/src/hooks/internal/useMemoized.ts
@@ -39,7 +39,7 @@ export default function useMemoized<TFinal, TArgs>(fn: Fn<TArgs, TFinal>, deps: 
         const lookupCached = ({ args: cachedArgs }: { args: TArgs[] }) =>
           args.length === cachedArgs.length && args.every((arg, index) => Object.is(arg, cachedArgs[+index]));
 
-        const cached = nextCache.find(lookupCached) ?? cache.find(lookupCached);
+        const cached = cache.find(lookupCached) ?? nextCache.find(lookupCached);
         if (cached) {
           cached.args = args;
           nextCache.push(cached);
--- a/packages/component/src/hooks/internal/useMemoized.ts
+++ b/packages/component/src/hooks/internal/useMemoized.ts
@@ -36,20 +36,27 @@ export default function useMemoized<TFinal, TArgs>(fn: Fn<TArgs, TFinal>, deps: 
         const fn = fnRef.current;
         const cache = cacheRef.current;
         const nextCache = nextCacheRef.current;
-        const lookupCached = ({ args: cachedArgs }: { args: TArgs[] }) =>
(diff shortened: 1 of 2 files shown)


packages/component/src/hooks/internal/useMemoized.ts
```diff
@@ -33,17 +33,30 @@ export default function useMemoized<TFinal, TArgs>(fn: Fn<TArgs, TFinal>, deps:
...
+        const cache = cacheRef.current;
+        const nextCache = nextCacheRef.current;
+
+        const cached = cache.find(
           ({ args: cachedArgs }) =>
-            args.length === cachedArgs.length && args.every((arg, index) => Object.is(arg, cachedArgs[+index]))
-        ) || { result: fn(...args) };
+            // index is guarented to be a number here
+            // eslint-disable-next-line security/detect-object-injection
+            args.length === cachedArgs.length && args.every((arg, index) => Object.is(arg, cachedArgs[index]))
```
compulim, line 44: Could consider this:

```suggestion
            args.length === cachedArgs.length && args.every((arg, index) => Object.is(arg, cachedArgs[+index]))
```

packages/component/src/hooks/internal/useMemoized.ts
```diff
@@ -33,17 +33,30 @@ export default function useMemoized<TFinal, TArgs>(fn: Fn<TArgs, TFinal>, deps:
...
+        );
+        if (cached) {
+          cached.args = args;
+          nextCache.push(cached);
+          return cached.result;
+        }
 
-        nextCache.push({ args, result });
+        const newCached = {
```
compulim, line 52:
```suggestion
        const nextCached = {
```

Or if it looks too close to `nextCache` vs. `nextCached`, we could also consider `nextEntry` or `nextItem`.

packages/component/src/hooks/internal/useMemoized.ts
OEvgeny: The conversion is unnecessary in this case. To me this is more like a tooling issue, so I disabled the specific check.

compulim (approved):

Issue #4850, opened 2023-09-07
Copying some `styleOptions` to CSS variables


compulim:
### Is your feature request related to a problem?

(This is not related to idiosyncratic styling.)

Copying some `styleOptions` to CSS variables `--webchat__color--accent` will enable us to use CSS media queries, which will make it easy to implement dark theme.

We should also consider a customization story for overriding certain CSS variables in a specific component. For example, modifying `--webchat__padding--regular` for suggested actions.

We should not put runtime style options, such as `sendTimeout`, in CSS variables. `styleOptions` (or JSON) is still the best way to put variables read by JavaScript. This is because JavaScript will not receive any events when CSS variables changed. However, when `styleOptions` changes, React will re-render us.

### Describe the suggestion or request in detail

For naming, let's follow CSS BEM with hyphen naming. CSS BEM does not have naming system for CSS variables. However, we could use this format `--namespace__value-type--modifier`.

```css
--webchat__color--accent: rgb(0, 99, 177);
--webchat__padding--regular: 10px;
```

When converting from JavaScript to CSS, make sure numbers have `"px"` appended.

```js
{
  '--webchat__padding--regular': paddingRegular + 'px'
}
```

### Additional context

CSS variables is not supported in IE11. We tried a polyfill. However, it requires a different variable name, for example, `-webchat__color--accent`, instead of `--webchat__color--accent`.

Because `-xyz` is not a CSS-compliant way to specify CSS variables, we should not use the polyfill.

As IE11 population is diminishing, it should be okay to let Web Chat looks ugly on IE11. In the future we need to go back on this decision, we may be able to use a programmatic way to inject CSS variables into our emotion stylesheet and enable this only for IE11.

The limitation would be the resulting stylesheet will not change when media queries change.

PR #5236, opened 2024-07-05
Add `bubbleImageMaxHeight` and `bubbleImageMinHeight`

### LINKED ISSUE #5235: Use `object-fit` for `<CroppedImage>` so we don't need a fixed height
### Is your feature request related to a problem?

The bubble height for image is fixed at 240px (configurable). When displaying wide image, say with aspect ratio (AR) of 21:9), there are a lot of padding on top and bottom.

This is because the image bubble has fixed height.

### Describe the suggestion or request in detail

Use `object-fit` to resize the image so its height can be honored.

### Describe alternatives you have considered

_No response_

### Additional context

Maybe this CSS, or simpler than this.

image

Also, consider deprecating `bubbleImageHeight` and offers both `bubbleMaxImageHeight` and `bubbleMinImageHeight` instead.

compulim:
> Fixes #5235.

### Breaking changes

-  `styleOptions.bubbleImageHeight` is being deprecated in favor of `styleOptions.bubbleImageMaxHeight` and `styleOptions.bubbleImageMinHeight`. The option will be removed on or after 2026-07-05
   -  Image will be shown at minimium height of `180` and maximum height of `240`, instead of fixed height of `240`
   -  To revert to previous behavior, set both values to `240`

### Added

-  Added `styleOptions.bubbleImageMaxHeight` and `styleOptions.bubbleImageMinHeight` for variable image height, in PR [#5236](https://github.com/microsoft/BotFramework-WebChat/pull/5236), by [@compulim](https://github.com/compulim)

## Description

Previously, to support IE11, we could not use `object-fit`. For simplicity, we make images with fixed height.

As we discontinued support of IE11, we can now making image height variable (min/max) and relies `object-fit` to honor the aspect ratio.

## Design

The image resize logic is "enlarge image to meet fixed width, then crop/pad top/bottom to meet height."

Current design is "fixed height", it will crop the image if it is too tall, or pad the image if it is too short.

This PR will enable variable height by min/max. When min is set to `0` and max is set to `Infinity`, it means "fixed width and make the image as short/tall as it fit."

When the image is taller than max, it will be cropped. When the image is shorter than min, it will be padded.

When min/max are set to the same number, say, `240`, it will always crop/pad the image at height of 240px and this is same as today's behavior and enable backward compatibility.

If the deprecating `styleOptions.bubbleImageHeight` is set, it will propagate to `min`/`max` value and is backward compatible.

## Specific Changes

- Added `styleOptions.bubbleImageMaxHeight` and `styleOptions.bubbleImageMinHeight`, both defaults to `240`
- Deprecated `styleOptions.bubbleImageHeight` and changed default to `undefined`, which will honor `min`/`max`
- Updated some VRT due to sub-pixel differences

## Review Checklist


packages/api/src/StyleOptions.ts
```diff
@@ -150,7 +150,28 @@ type StyleOptions = {
...
+   * @deprecated Use `bubbleImageMaxHeight` and `bubbleImageMinHeight` instead. To mimick behavior before deprecation, set both options to 240px.
+   */
+  bubbleImageHeight?: number | undefined;
+
+  /**
+   * Specifies the maximum height of the bubble for image, default to 240px.
+   *
+   * New in 4.18.0.
+   */
+  bubbleImageMaxHeight?: number | undefined;
```
OEvgeny, line 166: Does it make sense to introduce these properties as deprecated from the start? This could:
- highlight our vision to migrate to CSS variables for stylistic changes
- communicate upcoming fundamental changes as early as possible
- provide more clear timeline for the upcoming change

OEvgeny (approved):

tdurnford (dismissed):

packages/api/src/StyleOptions.ts
compulim: IMO, the "`styleOptions` -> CSS variables" feature won't deprecate `styleOptions` until few years later. It will remain as an overlapping feature for extended periods.

The roadmap is in this issue, https://github.com/microsoft/BotFramework-WebChat/issues/4850. But I agree we don't have a good way to list out all roadmaps.

I badly want to do the CSS variables.

OEvgeny (approved):

PR #5251, opened 2024-08-02
feat: centralized announcements approach

OEvgeny:
> Fixes

## Changelog Entry

- Introduced centralized announcements approach via the new `usePushToLiveRegion` hook, in PR [#5251](https://github.com/microsoft/BotFramework-WebChat/pull/5251), by [@OEvgeny](https://github.com/OEvgeny)
- Added keyboard shortcut for the "New Messages" button, in PR [#5251](https://github.com/microsoft/BotFramework-WebChat/pull/5251), by [@OEvgeny](https://github.com/OEvgeny)

## Description

This PR introduces a centralized approach for handling announcements in Web Chat with the new `usePushToLiveRegion` hook. It addresses the issue of multiple live regions potentially interfering with each other and provides a more consistent way of managing announcements across different components.

The PR also adds a shortcut for the New Messages button, improving accessibility by announcing the keyboard shortcut for this functionality when the button becomes visible.

## Design

This design builds upon and refines the previous implementation of transcript live region announcements in Web Chat.

The live region functionality has been moved up to the Composer level to allow more components to access and utilize the live region for announcements.

The `usePushToLiveRegion` hook, which utilizes the `useQueueStaticElement` hook, was added to provide a way to push memoized JSX fragments to the live region.

The previous transcript live region implementation was refactored according to the design agreements recently reached.

## Specific Changes

- Introduced `usePushToLiveRegion` hook for centralized announcement management
- Updated components to use `usePushToLiveRegion` instead of individual live regions
- Added `Alt+Shift+U` shortcut for the New Messages button
- Added `aria-keyshortcuts` attribute to the New Messages button
- Updated tests to reflect the new announcement behavior
- Refactored touched HTML test files for consistency and improved readability

## -

-

## Review Checklist


packages/bundle/src/index-minimal.ts
```diff
@@ -4,6 +4,7 @@
 import { StrictStyleOptions, StyleOptions } from 'botframework-webchat-api';
 import * as decorator from 'botframework-webchat-api/decorator';
 import { Constants, createStore, createStoreWithDevTools, createStoreWithOptions } from 'botframework-webchat-core';
+import * as internal from 'botframework-webchat-component/internal';
```
OEvgeny, line 7: As we need this for Fluent, using dedicated bundle would mean UMD users need to include it in order for Fluent theme to work.

packages/component/src/providers/LiveRegionTwin/private/LiveRegionTwinComposer.ts
```diff
@@ -0,0 +1,29 @@
...
+          `botframework-webchat: LiveRegionTwinContext.${key} can only be used inside of <LiveRegionTwinComposer>.`
+        );
+      }
+    }
+  ) as unknown as LiveRegionTwinContextType
+);
+
+const LiveRegionTwinComposer = memo(createLiveRegionTwinComposer(LiveRegionTwinContext));
+
+LiveRegionTwinComposer.displayName = 'LiveRegionTwinComposer';
```
OEvgeny, line 27:
```suggestion
const LiveRegionTwinComposer = createLiveRegionTwinComposer(LiveRegionTwinContext);
```

OEvgeny: This avoids double memoization

packages/component/src/providers/LiveRegionTwin/private/LiveRegionTwinComposer.ts
```diff
@@ -0,0 +1,29 @@
+import { createContext, memo } from 'react';
```
OEvgeny, line 1:
```suggestion
import { createContext } from 'react';
```

packages/component/src/providers/LiveRegionTwin/private/LiveRegionTwinComposer.ts
compulim: Right.

packages/component/src/Middleware/ScrollToEndButton/ScrollToEndButton.js
```diff
@@ -1,29 +1,55 @@
...
+  const liveRegionText = localize(
+    'TRANSCRIPT_LIVE_REGION_NEW_MESSAGES_ALT',
+    localizeAccessKey(newMessagesAccessKey),
+    text
+  );
+
+  // Setup and announce new messages button shortcuts
+  useFocusAccessKeyEffect(scrollToEndButtonBehavior === 'any' ? '' : newMessagesAccessKey, focusRef);
+
+  useLiveRegion(
```
compulim, line 34: Whenever the scrollToEndButtonBehavior change, we will queue the live region text, is this correct behavior?

packages/component/src/Transcript/LiveRegionTranscript.tsx
```diff
@@ -1,47 +1,28 @@
...
-import isPresentational from './LiveRegion/isPresentational';
 import LiveRegionActivity from '../LiveRegion/LiveRegionActivity';
-import LiveRegionSendFailed from './LiveRegion/SendFailed';
-import LiveRegionTwinComposer from '../providers/LiveRegionTwin/LiveRegionTwinComposer';
 import tabbableElements from '../Utils/tabbableElements';
 import useLocalizeAccessKey from '../hooks/internal/useLocalizeAccessKey';
-import useQueueStaticElement from '../providers/LiveRegionTwin/useQueueStaticElement';
-import useStyleToEmotionObject from '../hooks/internal/useStyleToEmotionObject';
 import useSuggestedActionsAccessKey from '../hooks/internal/useSuggestedActionsAccessKey';
+import { useQueueStaticElement } from '../providers/LiveRegionTwin';
```
compulim, line 10: I would love to see the newer `usePushToLiveRegion()` (or as I renamed to `useLiveRegion()`) to be able to superseded `useQueueStaticElement()` in all cases.

The only exception is the usage in this component.

`<LiveRegionTranscript>` need a render loop and inspect the DOM element to see if certain live region text need to be added based on the rendered content.

To use the new `useLiveRegion()` hook in this component, we will need to introduce another render loop, then it would call the `useLiveRegion`.

It seems the new `useLiveRegion()` is not very capable on this scenario. Makes me think if the new approach is right or not.

When we want to push something to live region, it could be at anytime, inside or outside of render loop. In this component, we are pushing things to live region outside of render loop.

When pushing things during render loop, the `useLiveRegion()` feels like pure syntactic sugar on `useMemo()` + `useQueueStaticElement()`, than having a true purpose.

What do you think?

packages/fluent-theme/src/components/sendBox/private/useSubmitError.ts
```diff
@@ -37,9 +38,21 @@ const useSubmitError = ({
...
+    if (error === 'empty' && hasValue) {
+      setError(undefined);
+    }
+  }, [error, hasValue]);
+
+  const checkError = useCallback(() => {
+    setError(submitErrorRef.current);
+  }, [submitErrorRef]);
+
+  return useMemo<Readonly<[RefObject<SendError | undefined>, string | undefined, () => void]>>(
```
compulim, line 53: The return value of this hook feels like a random choice of stuff.

packages/fluent-theme/src/components/sendBox/SendBox.tsx
```diff
@@ -50,7 +50,7 @@ function SendBox(props: Props) {
   const sendMessage = useSendMessage();
   const makeThumbnail = useMakeThumbnail();
   const errorMessageId = useUniqueId('sendbox__error-message-id');
-  const [errorRef, errorMessage] = useSubmitError({ message, attachments });
+  const [errorRef, errorMessage, checkError] = useSubmitError({ message, attachments });
```
compulim, line 53: Adding `checkError` here seems like a weird mix... I think the original idea of `useSubmitError` is to able to synchronously know what errors we have.

Then, the `checkError` makes the check asynchronous (outside of render loop). I think the `useSubmitError` would need some redesigns.

packages/component/src/Middleware/ScrollToEndButton/ScrollToEndButton.js
```diff
@@ -1,29 +1,55 @@
...
   const text = localize(scrollToEndButtonBehavior === 'any' ? 'TRANSCRIPT_MORE_MESSAGES' : 'TRANSCRIPT_NEW_MESSAGES');
 
+  const liveRegionText = localize(
+    'TRANSCRIPT_LIVE_REGION_NEW_MESSAGES_ALT',
+    localizeAccessKey(newMessagesAccessKey),
+    text
+  );
+
+  // Setup and announce new messages button shortcuts
+  useFocusAccessKeyEffect(scrollToEndButtonBehavior === 'any' ? '' : newMessagesAccessKey, focusRef);
```
compulim, line 32: If `scrollToEndButtonBehavior` is `false`, we won't show the button. I think we shouldn't need `useFocusAccessKeyEffect` in that case.

packages/fluent-theme/src/components/sendBox/private/useSubmitError.ts
```diff
@@ -37,9 +38,21 @@ const useSubmitError = ({
     [localize]
   );
 
-  return useMemo<Readonly<[RefObject<SendError | undefined>, string | undefined]>>(
-    () => Object.freeze([submitErrorRef, submitErrorRef.current && errorMessageStringMap.get(submitErrorRef.current)]),
-    [errorMessageStringMap, submitErrorRef]
+  const hasValue = !!message?.trim();
+
+  useEffect(() => {
+    if (error === 'empty' && hasValue) {
```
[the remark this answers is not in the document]
OEvgeny: Yea, we're resetting it after value has updated. This is an improvement for today's UX. I don't have a better idea yet. I can probably remove it from this PR if you don't want it :rofl:

packages/fluent-theme/src/components/sendBox/private/useSubmitError.ts
OEvgeny: It only looks like. The ref contains the actual error state, while the string contains the error state we'd like to show. The ref is required to be ref so we don't accidentally re-render on the change.

Sorry, I'm not seeing better design at this point. I probably can rename things to make this clear.

packages/fluent-theme/src/components/sendBox/SendBox.tsx
OEvgeny: Idea a bit changed. Maybe the name for the `checkError` is not correct... We'd like to postpone setting error on submit for most cases. It leads to better UX.

The idea was to provide both outputs to the component:
- the actual error, so `onSubmit` could rely on it
- the error we'd like to show

The `checkError` synchronizes these two states.

packages/fluent-theme/src/components/sendBox/private/useSubmitError.ts
OEvgeny: Updated. It's not perfect still, but should look a bit better.

compulim (approved):

PR #5265, opened 2024-08-21
Livestreaming informative message

### LINKED ISSUE #4876: WebChat: Support for ‘stream’ output in ChatGPT style while using Azure OpenAI Completion
Bot responses with Azure OpenAI Completion, but we need the response to be displayed in a live streaming format.

### Describe the suggestion or request in detail

Allow Application that uses Bot framework webchat to stream out the response like ChatGPT web chat does.

### LINKED ISSUE #4939: SSR (Server Sent Event support)/Streaming support in WebChat
### Is your feature request related to a problem?

If we have to implement functionality like https://chat.openai.com/ where answer is rendered word (or few words) at a time instead of getting the complete answer in one shot, would it be possible to achieve it with BotFramework WebChat? If we get streamed answer from an LLM (say as SSR), which we can intercept and  and send "a", "a b", "a b c", "a b c d" and so on, on the WebSocket, is there some way in which we can make WebChat render it like how we see it in https://chat.openai.com/?

### Describe the suggestion or request in detail

_No response_

### Describe alternatives you have considered

_No response_

### Additional context

_No response_

compulim:
> Resolves #4876 and resolves #4939.

### Added

-  Resolves [#4876](https://github.com/microsoft/BotFramework-WebChat/issues/4876) and [#4939](https://github.com/microsoft/BotFramework-WebChat/issues/4939). Added support of informative message in livestreaming, by [@compulim](https://github.com/compulim), in PR [#5265](https://github.com/microsoft/BotFramework-WebChat/pull/5265)

## Description

This pull request will add support for informative message for livestreaming session.

Informative message is a side-channel message that usually appear at the start of the livestreaming session. The message is usually describe how the bot is preparing for the livestreaming session, e.g. "Searching your document library..."

Also, in this pull request, we are improving interoperability by making `Activity` shape check stricter and aligned with other services such as Teams.

### Livestreaming activity type

| Livestream type | Decoration | Description |
| - | - | - |
| "informative message" | "preparing" | Optional side-channel informative message describing the current response, e.g. "Searching your document library". Always replace past informative messages. May interleave with interim activities. |
| "interim activity" | "ongoing" | Current response, could be empty, partial-from-start, or complete response. More activities are expected. Future interim activities always replace past interim activities, enable erasing or backtracking response. |
| "final activity" | "completing" | Complete-and-final response, always replace past interim activities and remove all informative messages. This activity indicates end of the session, all future activities must be ignored. |

### Stricter `Activity` shape for livestreaming

Previously, we are relaxed on the mandatory fields for the `Activity` object to consider the activity as part of a livestreaming session.

To improve interoperability, we are bringing full rules to the `Activity` object. The implementation of the rule can be found at [getActivityLivestreamingMetadata.ts](https://github.com/microsoft/BotFramework-WebChat/pull/5265/files#diff-cfe9600573dbf04168f477c308bbd6a04dbfc330b0d5fc36f8951a39ee4bfd2b).

`getActivityLivestreamingMetadata()` will return an object if the activity is part of a livestreaming session. Otherwise, it will return `undefined` and indicate the activity is *not* part of a livestreaming session. It is recommended to use this function to get livestreaming metadata for the activity.

- `activity.channelData.streamSequence` must be a `number`, starting from `1` and monotonic increasing integer
   - This number represents the order sequence of the activity in the livestreaming session
- `activity.id` must be a `string` (empty or non-empty)
- `activity.text` must be a `string` (empty or non-empty)
- `activity.channelData.streamType` must be either `"informative"`, `"streaming"` or `"final"`
   - For `"informative"` (sidechannel informative message) and `"streaming"` (interim activity)
      - `activity.type` must be `"typing"`
      - `activity.channelData.streamId` is an optional `string`, representing the ID of the current livestreaming session
   - It must be the activity ID of the first activity in the livestreaming session
      - We are unable to enforce this rule on client-side. However, service may enforce this rule
   - For `"final"` (final activity of the session)
      - `activity.channelData.streamId` must be a `string`, representing the ID of the current livestreaming session
         - In other words, final activity must not be the first activity of the livestreaming session
      - `activity.type` must be `"message"`

For the explanation of `informative`, `streaming` and `final`, please refer to `getActivityLivestreamingMetadata()` for details.

## Specific Changes

- Existing tests were updated to reflect the stricter requirements on the `Activity` object
- Changed `ActivityDecoratorRequest`
   - `state` is too vague if we have a few orthogonal decorations to apply to (e.g. timestamp, border, etc.)
   - `decorateForLivestreaming` is dedicated for indicating what kind of decorations should apply to the livestreaming activity
- Removed extraneous exposed port in Docker Compose
- Added `docs/LIVESTREAMING.md` to describe how to use livestreaming

## Review Checklist


packages/api/src/decorator/private/activityDecoratorRequest.ts
```diff
@@ -1,6 +1,23 @@
...
+  /**
+   * Decorate the activity as it participate in a livestreaming session.
+   *
+   * - `"completing"` - decorate as the livestreaming is completing
+   * - `"ongoing"` - decorate as the livestreaming is ongoing
+   * - `"preparing"` - decorate as the livestreaming is being prepared
+   * - `undefined` - not participated in a livestreaming session
+   */
+  decorateForLivestreaming: 'completing' | 'ongoing' | 'preparing' | undefined;
+
```
OEvgeny, line 11: How we would add a new type here? Imagine we'd like to decorate for reaction. Would it be a new `decorateForReaction` prop?

Most of fields are named the way they describe a value compared to the new one describing the action to be done (decorate something). Not against the naming, but it'd be nice to have an insight on why we introduce such.

packages/api/src/hooks/useActiveTyping.ts
```diff
@@ -36,15 +36,16 @@ function useActiveTyping(expireAfter?: number): readonly [Readonly<Record<string
...
   useEffect(() => {
+    const timeToRender = earliestExpireAt && earliestExpireAt - Date.now();
+
     if (timeToRender && isFinite(timeToRender)) {
       const timeout = setTimeout(forceRender, Math.max(0, timeToRender));
 
       return () => clearTimeout(timeout);
     }
-  }, [clearTimeout, forceRender, setTimeout, timeToRender]);
+  }, [clearTimeout, Date, earliestExpireAt, forceRender, setTimeout]);
```
OEvgeny, line 49: This change probably causes an additional re-render, but looking good to me.

packages/api/src/providers/ActivityTyping/ActivityTypingComposer.tsx
```diff
@@ -47,7 +38,12 @@ const ActivityTypingComposer = ({ children }: Props) => {
...
         } = activity;
 
-        if (type === 'typing' && (role === 'bot' || role === 'user')) {
+        const livestreamingMetadata = getActivityLivestreamingMetadata(activity);
+
+        if (type === 'message' || livestreamingMetadata?.type === 'final activity') {
+          nextTyping.delete(id);
+          changed = true;
+        } else if (type === 'typing' && (role === 'bot' || role === 'user')) {
           const currentTyping = nextTyping.get(id);
```
OEvgeny, line 47: Just throwing some ideas: seems this could also lead to more re-renders due to this change might cause more frequent updates to the typing state

packages/component/src/Activity/StackedLayout.tsx
```diff
@@ -102,7 +101,12 @@ const StackedLayout: FC<StackedLayoutProps> = ({
   const fromUser = activity.from.role === 'user';
   const messageBackDisplayText: string = (isMessage && activity.channelData?.messageBack?.displayText) || '';
 
-  const activityDisplayText = isMessage ? messageBackDisplayText || activity.text : '';
+  const isLivestreaming = getActivityLivestreamingMetadata(activity);
```
OEvgeny, line 104:
```suggestion
  const isLivestreaming = !!getActivityLivestreamingMetadata(activity);
```
Otherwise it's an object which may cause re-renders and would need memoization.

packages/api/src/hooks/useActiveTyping.ts
compulim: It's actually the otherwise. I may be wrong somewhere, so let me explain how it happened.

Before the change in this file, render was added/removed. Then, it caused the `timeToRender` to be inaccurate by 1 ms.

Says, t=900, `earliestExpireAt` is 1000, which means it should wake up at t=1000. `timeToRender` is computed to be 100ms. So, everything works perfectly.

But after some changes somewhere else, `earliestExpireAt` is still 1000. But it woke up at t=999 instead of t=1000 (either now or `timeToRender` is wrong). It woke up 1 ms too early and t=999 didn't reach `expireAt`. So the logic did not expire some typing indicators.

Instead of passing `timeToRender`, which is a relative number. I am passing `earliestExpireAt` which is an absolute number. And it would calculate `timeToRender` inside the `useEffect`, so the timeout value should be more accurate.

Also, I think passing an absolute number would reduce the chance that this `useEffect` got re-render. `earliestExpireAt` is based on solely of all typing entries. `timeToRender` is based on all typing entries, *plus* `Date.now()`. So, `earliestExpireAt` should have a lesser chance to cause re-render.

packages/api/src/decorator/private/activityDecoratorRequest.ts
compulim: In our offline chat, I thought you want to name it `decorateForXXX`. I feel weird too. 🤣

I guess we would name it `livestreaming` or `livestreamingState`. What's your thought?

packages/api/src/providers/ActivityTyping/ActivityTypingComposer.tsx
compulim: Unlikely I guess, here is why.

For `type === 'message' || livestreamingMetadata?.type === 'final activity'`, it means, when receiving a message activity or livestreaming final activity, then, it should remove the typing entry.

However, all final activity *must be* type of message. So, the extraneous `||` here is for defensive programming.

Looking at the whole thing, it would just switched the position of the if-statement to the top rather than the bottom.

Then, the other if-statement is `type === 'typing'`, so it would be orthogonal to `type === 'message'`. So switching the position of the if-statement shouldn't matter.

packages/api/src/decorator/private/activityDecoratorRequest.ts
OEvgeny: Sorry for confusion, I'm ok with both specific field names proposed. Alternatively we could go with an array of decorations (`decoration: []` field) maybe.

OEvgeny: Sounds good!

OEvgeny (approved):

NagarroShobharaja:
Hello All,

i tried implementing the solution using the below code let me know what i'm doing it wrong

else if (turnContext.Activity.ChannelId.Equals(Channels.Directline))
            {
                var oldActivity = new Bot.Schema.Activity();

                oldActivity.Type = "typing";
                oldActivity.Text = "Searching your document library...";
                oldActivity.ChannelData = new ChannelData()
                {
                    streamSequence = 1,
                    streamType = "informative"
                };
                var activityid = await turnContext.SendActivityAsync(oldActivity);

                Thread.Sleep(1000);

                oldActivity.Type = "typing";
                oldActivity.Text = "A quick";
                oldActivity.ChannelData = new ChannelData()
                {
                    streamId = activityid.Id,
                    streamSequence = 2,
                    streamType = "streaming"
                };
                await turnContext.SendActivityAsync(oldActivity);

                Thread.Sleep(1000);

                oldActivity.Type = "typing";
                oldActivity.Text = "A quick brown fox";
                oldActivity.ChannelData = new ChannelData()
                {
                    streamId = activityid.Id,
                    streamSequence = 3,
                    streamType = "streaming"
                };
                await turnContext.SendActivityAsync(oldActivity);

                Thread.Sleep(1000);

                oldActivity.Type = "message";
                oldActivity.Text = "A quick brown fox jumped over the lazy dogs.";
                oldActivity.ChannelData = new ChannelData() { streamId = activityid.Id, streamType = "final" };
                await turnContext.SendActivityAsync(oldActivity);
            }

PR #5312, opened 2024-09-26
feat: Add white-label CSS decorator

OEvgeny:
> Fixes

## Changelog Entry

- (Experimental) Added CSS decorator support into Web Chat white-label experience, in PR [#5312](https://github.com/microsoft/BotFramework-WebChat/pull/5312), by [@OEvgeny](https://github.com/OEvgeny)
   - Introduced `WebChatDecorator` component for adding animated borders to activities, in PR [#5312](https://github.com/microsoft/BotFramework-WebChat/pull/5312)
   - Added new style options `borderAnimationColor1`, `borderAnimationColor2`, and `borderAnimationColor3` for customizing decorator colors, in PR [#5312](https://github.com/microsoft/BotFramework-WebChat/pull/5312)

## Description

This PR introduces a white label CSS decorator for Web Chat, allowing developers to add animated borders to activities. The new `WebChatDecorator` component provides a customizable way to enhance the visual feedback for different activity states, such as "preparing" and "completing".

## Design

The decorator is designed to be easily integrated into existing Web Chat implementations. It uses CSS animations and custom properties to achieve the desired visual effects while maintaining flexibility for customization.

The new `botframework-webchat-styles` package handles shared helpers to work with CSS styles.

## Specific Changes

- Added `WebChatDecorator` component in `packages/component/src/decorator/`
- Implemented `BorderFlair` and `BorderLoader` components for different animation states
- Created new style options: `borderAnimationColor1`, `borderAnimationColor2`, `borderAnimationColor3`
- Introduced `botframework-webchat-styles` package for centralized style management
- Updated `Composer` component to use the new decorator
- Refactored style injection mechanisms for better performance
- Implemented Firefox-compatible animations using `@keyframes`
- Updated tests to reflect new decorator functionality

## -

## Review Checklist


packages/component/.eslintrc.yml
```diff
@@ -5,3 +5,6 @@ extends:
 # This package is expected to run inside a web browser.
 env:
   browser: true
+
+rules:
```
compulim, line 9: Maybe put at root `.eslintrc.typescript.yml`?

packages/component/package.json
```diff
@@ -42,7 +52,8 @@
     "./dist/**/*",
     "./*.js",
     "./lib/**/*",
-    "./src/**/*"
+    "./src/**/*",
+    "./*.js"
```
compulim, line 56: Already have it on line 53. 😎

packages/component/src/Styles/useInjectCustomProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+      fontSizeSmall,
+      markdownExternalLinkIconImage,
+      paddingRegular,
+      primaryFont,
+      subtle,
+      timestampColor
+    } = styleOptions;
+
+    // eslint-disable-next-line no-magic-numbers
+    const randomClass = `wc-${Math.ceil(Math.random() * Number.MAX_SAFE_INTEGER).toString(36)}`;
```
compulim, line 32: We can't use `Math.random()` despite we have no crypto concern here.

Need to use `math-random` or anything under Web Crypto API.

packages/component/.eslintrc.yml
OEvgeny: I guess you're referring to `.eslintrc.production.yml`

packages/component/src/Styles/useInjectCustomProperties.ts
```diff
@@ -0,0 +1,58 @@
+import { hooks } from 'botframework-webchat-api';
+import { makeCreateStyles } from 'botframework-webchat-styles';
```
compulim, line 2: Should we have a new package?

Should we create a new private package for isolation, but bundle it in when compiling?

The other side to think... who can use `bf-wc-styles`? Can it be used without `bc-wc-component`?

Should we put all `styleOptions` and related stuff (say, rectification), inside the new `bf-wc-styles` package?

I like we say, `bf-wc-styles/react`. Good to use named import for different ecosystems.

packages/component/src/Styles/useInjectCustomProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+import { useMemo } from 'react';
+import CustomPropertyNames from './CustomPropertyNames';
+import useInjectStyles from '../hooks/internal/useInjectStyles';
+import useNonce from '../hooks/internal/useNonce';
+
+const { useStyleOptions } = hooks;
+
+const webchatCustomPropertiesClass = 'webchat__css-custom-properties';
+
+export default function useInjectCustomProperties() {
```
compulim, line 12: This seems like a private function that is only use by `bf-wc-component/Composer`.

Can we call it `useCustomPropertiesClassName(): readonly [string]`?

Resonating with my question above... should this put inside `bf-wc-styles` package?

packages/styles/src/private/makeCreateStyles.ts
```diff
@@ -0,0 +1,15 @@
+function createStyle(content: string) {
+  const style = document.createElement('style');
+  style.append(document.createTextNode(content));
+  return style;
+}
+
+export default function makeCreateStyles(...contents: string[]) {
+  return function createStyles() {
+    if (!globalThis.document) {
```
compulim, line 9: Should we throw if `document` is not there? I know SSR, but this function seems like it will create (detached) styles.

Could also be, `safeMakeCreateStyles`, `makeCreateSafeStyles`, `safeMakeCreateDetachedStyles`...

packages/component/src/Styles/useInjectCustomProperties.ts
OEvgeny: I don't want styles to depend on api. I think api -> component relationship is better meaning simpler hence left it here.

packages/component/src/Styles/useInjectCustomProperties.ts
OEvgeny: At the moment the `styles` package is internal. It is dedicated for providing necessary tools for applying styles. Think of it as `emotion`, but with smaller scope and tightened to our selection of tools.

Even in case we expose it, it seems we would need a better name for it (without `botframework-webchat` prefix) as while being our set of tools, it doesn't limit its usage in any way.

We can make it another package for styling Web Chat, but I don't think it's necessary at this point.

OEvgeny: In case we'll need a `component2` package which leverages `core`, `api` etc for new experience, we can think on increasing the scope for `styles` to cover things we want to share between the `component` and `component2`.

packages/fluent-theme/src/styles/useStyles.ts
```diff
@@ -1,19 +1,2 @@
...
-          ])
-        )
-      ),
-    [styles]
-  );
-}
-
-export default useStyles;
+// TODO: refactor after initial approval gained
+export { useStyles as default } from 'botframework-webchat-styles/react';
```
OEvgeny, line 2: Can handle this in a follow-up PR

packages/styles/src/env.d.ts
```diff
@@ -0,0 +1,7 @@
+// CSS modules
```
compulim, line 1: Should we name this file `CSSModulesClasses.d.ts`?

packages/styles/src/private/makeCreateStyles.ts
```diff
@@ -0,0 +1,15 @@
+function createStyle(content: string) {
+  const style = document.createElement('style');
+  style.append(document.createTextNode(content));
+  return style;
+}
+
+export default function makeCreateStyles(...contents: string[]) {
```
compulim, line 7: Should we call this `makeCreateStyleElements`?

packages/component/src/decorator/private/Decorator.tsx
```diff
@@ -13,8 +15,14 @@ const middleware: DecoratorMiddleware[] = [
     (next => request => (request.livestreamingState === 'preparing' ? BorderLoader : next(request)))
 ];
 
-function FluentThemeDecorator({ children }: Readonly<{ readonly children?: ReactNode | undefined }>) {
-  return <DecoratorComposer middleware={middleware}>{children}</DecoratorComposer>;
+const styles = createStyles();
```
compulim, line 18: If we always create styles on importing, then should we `makeCreateStyles` -> `createStyles` instead?

packages/styles/src/build/private/injectCSSPlugin.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const stylesPlaceholderQuoted = JSON.stringify(stylesPlaceholder);
+
+  return {
+    name: `inject-css-plugin(${stylesPlaceholder})`,
+    setup(build) {
+      build.onEnd(({ outputFiles = [] }) => {
+        for (const file of outputFiles) {
+          if (file.path.match(/(\.js|\.mjs)$/u)) {
+            const js = file;
```
compulim, line 35: Do we need this line?

packages/component/src/Styles/useInjectCustomProperties.ts
compulim: I was thinking `styles` depends on `component`, thus I was thinking about it that way. 🤣

I think it makes sense to have a new package. If it lives outside of the `botframework-webchat` prefix, then we should have a different version number not tie to Web Chat.

packages/component/src/Styles/useCustomPropertiesClassName.ts
```diff
@@ -0,0 +1,64 @@
...
+  ${CustomPropertyNames.MinHeightImageBubble}: ${bubbleImageMinHeight}px;
+  ${CustomPropertyNames.PaddingRegular}: ${paddingRegular}px;
+}
+`;
+    const [style] = makeCreateStyles(contents)();
+    style.dataset.webchatInjected = 'component';
+    return [Object.freeze([style]), Object.freeze([`${webchatCustomPropertiesClass} ${randomClass}`] as const)];
+  }, [styleOptions]);
+
+  useInjectStyles(styles, nonce);
```
compulim, line 61: Should we move `useNonce()` inside `useInjectStyles()`?

packages/styles/package.json
```diff
@@ -0,0 +1,87 @@
...
+        "types": "./dist/botframework-webchat-styles.react.d.mts",
+        "default": "./dist/botframework-webchat-styles.react.mjs"
+      },
+      "require": {
+        "types": "./dist/botframework-webchat-styles.react.d.ts",
+        "default": "./dist/botframework-webchat-styles.react.js"
+      }
+    },
+    "./tsconfig.json": "./src/ts-config/config.json",
+    "./env": {
```
compulim, line 39: Maybe call this `"./dts"`? What do you think?

packages/component/src/decorator/private/BorderFlair.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { Fragment, memo, useCallback, useState, type ReactNode } from 'react';
 import cx from 'classnames';
+import { useStyles } from 'botframework-webchat-styles/react';
```
compulim, line 3: IntelliSense is not working for me on `classNames.`, is it working for you?

packages/styles/src/index.ts
```diff
@@ -0,0 +1 @@
+export { default as makeCreateStyles } from './private/makeCreateStyles';
```
compulim, line 1: In the future, when we want to make it better, I want to reduce a confusing part:

- `makeCreateStyles` create `HTMLStyleElement[]` and it contains styles from every `*.module.css` in the bundle
- `useStyles` contains class names for only one `*.module.css`, not every in the bundle

Can we pick either one concept?

packages/styles/src/react/private/useStyles.ts
```diff
@@ -0,0 +1,19 @@
...
+
+function useStyles<T extends CSSModuleClasses>(styles: T): T {
+  // @ts-expect-error: entries/fromEntries don't allow to specify keys type
+  return useMemo(
+    () =>
+      Object.freeze(
+        Object.fromEntries(
+          Object.entries(styles).map(([baseClassName, resultClassName]) => [
+            baseClassName,
+            `${baseClassName} ${resultClassName}`
```
compulim, line 11: Question... why do we need to append the key to the class name? Is it only for React?

packages/component/src/decorator/private/Decorator.tsx
```diff
@@ -1,8 +1,10 @@
 import { DecoratorComposer, type DecoratorMiddleware } from 'botframework-webchat-api/decorator';
```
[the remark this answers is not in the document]
OEvgeny: It is optional and exported. I think it may be considered a breaking change to include it. I marked it as experimental. I'd expect we to not rush and include it by default maybe in 4.20.x

packages/styles/src/env.d.ts
```diff
@@ -0,0 +1,7 @@
...
+type CSSModuleClasses = { readonly [key: string]: any };
```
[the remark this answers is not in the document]
OEvgeny: There is a way to export values:

https://github.com/css-modules/css-modules/blob/master/docs/values-variables.md#exporting-values-variables
 
We don't use it currently and seems the TypeScript plugin doesn't support it, so I guess, we're good either way.

packages/styles/src/env.d.ts
OEvgeny: Going with `env.d.ts` allows us adding other style related globals here in the future.

The idea is: you reference `bf-wc-styles/env` in your typescript env file and you get the environment supporting CSS modules.

packages/styles/src/private/makeCreateStyles.ts
OEvgeny: I don't see the need in extra verbosity. Let me know if you prefer more verbose one

packages/component/src/decorator/private/Decorator.tsx
OEvgeny: I think this was discussed in our early design discussions.

I can't fully understand the proposal. The `makeCreateStyles` allows us to share the logic behind `createStyles` function for each of `xxx/createStyles` module.

packages/styles/src/build/private/injectCSSPlugin.ts
OEvgeny: Not really, but it helped readability

packages/component/src/Styles/useCustomPropertiesClassName.ts
OEvgeny: I'm not sure. I think it is good as an argument. What if we expose it allowing someone creating a style with nonce? This feels like a bad design security-wise.

packages/component/src/decorator/private/BorderFlair.tsx
OEvgeny: Yes, it does:
Screenshot from 2024-09-27 09-54-54

packages/styles/src/index.ts
OEvgeny: Yea, unfortunately for now this reflects how we bundle things. The design was influenced the actual implementation of the bundling CSS and such

packages/styles/src/react/private/useStyles.ts
OEvgeny: Per convention we have in `component` package the classes look like: `webchat__block__element--modifier <generated-class-we-apply-styles-to>`.

In Web Chat we had to specify the `webchat__block__element--modifier` by hand. This reflects the convention and automates the part, so we in most cases we only need to: `className={classNames['something']}`.

We do so to provide clear naming for classes we are using, so they can be styled outside of Web Chat.

packages/component/tsup.config.ts
```diff
@@ -1,10 +1,22 @@
...
 
 export default defineConfig({
   ...baseConfig,
+  loader: {
+    ...baseConfig.loader,
+    '.css': 'local-css'
+  },
+  esbuildPlugins: [
+    injectCSSPlugin({ stylesPlaceholder: componentStyleContentPlaceholder }),
+    injectCSSPlugin({ stylesPlaceholder: decoratorStyleContentPlaceholder })
```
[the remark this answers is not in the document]
OEvgeny: Both are needed :smile:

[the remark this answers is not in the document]
OEvgeny: We have one per bundle

packages/component/src/Styles/useInjectCustomProperties.ts
OEvgeny: As it is internal for now, we can make this later. Our setup is pretty lightweight and may indeed be useful for making CSS modules to be available today.

This needs the following improvements to be useful for community:
- The build plugin should use [`unplugin`](https://unplugin.unjs.io/) so it supports most of bundlers, not only esbuild.
- We should probably switch from `style` elements to [`CSSStyleSheet`](https://web.dev/constructable-stylesheets/) for handling styles as this is something we expect to get from `import style from '*.css' with { type: 'css' };`
- We should improve `makeCreateStyles` and `createStyles` to make it less confusing at least

A few of these are going along with current Web Chat goals:
- We use only esbuild/tsup
- The `CSSStyleSheet` is not yet very useful for our needs: we have to convert back and forth to style elements.

packages/component/src/decorator/private/BorderFlair.tsx
compulim: image

On my side... don't know why, it's a brand new Codespaces.

packages/component/src/decorator/private/Decorator.tsx
compulim: We can talk about it offline and in another PR.

packages/component/src/Styles/useCustomPropertiesClassName.ts
compulim: Agree with your points. Some days in the future, we should put more eyes on the `useNonce` and audit all routes and potentially making less calls to `useNonce` to reduce attack surface.

compulim (approved):
Solid new package.

packages/component/src/Styles/useInjectCustomProperties.ts
compulim: Different POV, but all are good stuff.

packages/component/src/decorator/private/BorderFlair.tsx
OEvgeny: Note to future travellers: make sure to switch to the workspace version of TypeScript

PR #5334, opened 2024-10-25
Add code block copy button

compulim:
> Fixes #

### Added

-  Added copy button to code blocks, in PR [#5334](https://github.com/microsoft/BotFramework-WebChat/pull/5334), by [@compulim](https://github.com/compulim)

## Description

For code blocks inside Markdown (via triple-backtick or `<pre>`), we will add a copy button on the top right corner of it.

image

## Design

We are introducing internal-use Web Components.

As we go through sanitization, DOM become string and it lose event handler. In other words, we cannot listen to "click" event on the button after sanitization.

Our solution is to use Web Components so we can preserve logic during sanitization.

Alternatively, we could have a catch-all click handler on Markdown, but it is not trivial as Adaptive Cards requires a HTML string instead of DOM too. That means, we will need to put a catch-all container in 2+ different places.

## Specific Changes

- Added new `CustomElementsProvider` to host internal-use hash-secured Web Components
- Added new code block copy button `<webchat-xxxxx--code-block-copy-button>` where `xxxxx` is a random hash
- Chaining `package.json` via `touch` so changes to `core` will trigger rebuild of `api`, and so on

## Review Checklist


OEvgeny (dismissed):

packages/component/src/providers/CustomElements/customElements/registerCodeBlockCopyButton.ts
```diff
@@ -0,0 +1,90 @@
...
+      );
+
+      const buttonElement = this.ownerDocument.createElement('button');
+
+      buttonElement.classList.add('webchat__code-block-copy-button', styleSet.codeBlockCopyButton);
+      buttonElement.dataset.testid = testIds.codeBlockCopyButton;
+      buttonElement.type = 'button';
+
+      buttonElement.append(copyIconImageElement);
+      buttonElement.append(copiedIconImageElement);
```
OEvgeny, line 40: Using html module would be a nice touch here I guess.

OEvgeny (approved):

OEvgeny (approved):

packages/component/src/providers/CustomElements/customElements/registerCodeBlockCopyButton.ts
```diff
@@ -9,7 +9,8 @@ type Strings = Readonly<{
 type StyleSet = ReturnType<typeof useStyleSet>[0];
 
 export default function registerCodeBlockCopyButton(hash: string, strings: Strings, styleSet: StyleSet): string {
-  const name = `webchat-${hash}__code-block-copy-button`;
+  // Allowed tag names are ASCII alphanumeric only, https://html.spec.whatwg.org/multipage/syntax.html#tag-name.
+  const name = `WebChatCodeBlockCopyButton${hash}`;
```
OEvgeny: I think it should have at least one hyphen, and AFAIK html names are case insensitive

packages/component/src/providers/CustomElements/customElements/registerCodeBlockCopyButton.ts
```diff
@@ -1,12 +1,8 @@
...
 
-type ObservedAttributes = 'className' | 'data-alt-copied' | 'data-alt-copy';
+type ObservedAttributes = 'class' | 'data-alt-copied' | 'data-alt-copy';
 
-const observedAttributes: readonly ObservedAttributes[] = Object.freeze([
-  'className',
-  'data-alt-copied',
-  'data-alt-copy'
-]);
+const observedAttributes: readonly ObservedAttributes[] = Object.freeze(['class', 'data-alt-copied', 'data-alt-copy']);
```
OEvgeny, line 5: Welcome back to html world :rofl:

OEvgeny (dismissed):

packages/component/src/providers/CustomElements/customElements/registerCodeBlockCopyButton.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export default function registerCodeBlockCopyButton(hash: string): string {
+  // Allowed tag names are specified here, https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define#valid_custom_element_names
+  const name = `webchat-${hash}--code-block-copy-button`;
+
+  customElements.define(
+    name,
+    // Using anonymous class because one class can only be registered with one tag name.
+    // Web Components enforce 1:1 relationship.
+    class extends CodeBlockCopyButtonElement {
```
OEvgeny, line 117: If you don't want to create a chain of inheritance, it should be possible to wrap the class creation into a function and call here.

On the other hand this approach allows ckecking if we derived from the initial class.

Overall, there shouldn't be a problem keeping 1-1 relationship between Web Chat instances using React primitives. I assume something else needs our attention.

OEvgeny (dismissed):

OEvgeny (approved):

packages/component/src/providers/CustomElements/customElements/registerCodeBlockCopyButton.ts
compulim: All lowercase, at least one hyphen. :)

PR #5380, opened 2024-11-14
Add scrollbar to math

compulim:
### Fixed

-  Fixed long math formula should be scrollable, in PR [#5380](https://github.com/microsoft/BotFramework-WebChat/pull/5380), by [@compulim](https://github.com/compulim)

## Description

Adding scrollbar to very long math formula.

image

## Design

Scrollbar in Fluent UI and Copilot variant themes is custom-styled. In white-label UX, we are using `scrollbar-width: thin`.

### Browser quirks

image

When the math formula is very long, browser may calculate the `<math>` size wrongly and pushed it a bit left and clipped it on `overflow-x: auto`.

## Specific Changes

- Updated `Theme.module.css` to include Math `<figure>` for thin scrollbar style

## Review Checklist


__tests__/html2/markdown/math/layout.scroll.html.snap-1.png
OEvgeny, line 1: Thin scrollbars seem to be implemented in Chrome 121+.

I think this is missing some paddings, but this means we diverge from browser styles for content.

packages/component/src/Styles/StyleSet/RenderMarkdown.ts
```diff
@@ -79,6 +82,10 @@ export default function createMarkdownStyle() {
         flexDirection: 'column'
       },
 
+      '& :is([data-math-type=block], [data-math-type=inline]) > span': {
```
OEvgeny, line 85: I'd look into removing the span

OEvgeny:
As we have now formulas scrollable, shouldn't they be focusable / scrollable via keyboard?

packages/component/src/Styles/StyleSet/RenderMarkdown.ts
compulim: Should this be something handled on your side of work right now? Say, `mathChunk` should not render any tags.

compulim:
> As we have now formulas scrollable, shouldn't they be focusable / scrollable via keyboard?

Right... it's not caught by `axe-core`.

packages/component/src/Styles/StyleSet/RenderMarkdown.ts
OEvgeny: On the extension side we do render only root plus what's KaTeX returned. Looks like we need to adjust the renderer by opting to `renderToDomTree` instead of `renderToString`:

```
var renderToString = function renderToString(expression, options) {
  var markup = renderToDomTree(expression, options).toMarkup();
  return markup;
};
```

I can look into it.

compulim: Not quite... :rofl:

image

__tests__/html2/markdown/math/layout.scroll.html.snap-1.png
OEvgeny: Some brainstorming:

https://codepen.io/OEvgeny/pen/PoMLBzr

I think we should still increase the space. Without inner space the scrollbar looks not in place. We maybe can just define default styles for figure to say we expect them to be present

compulim: I am always impressed by the quality of your samples. :wink:

compulim: Incorporated.

OEvgeny (approved):

PR #5394, opened 2024-12-16
fix: dropping file out of drop zone makes it stuck in visible state

### LINKED ISSUE #5390: Dropzone sendbox stuck in droppable state when drop outside of Dropzone

No

No

Latest production

Bundle (webchat.js)

Web apps

Browser: Edge (latest)

Send box

### Which theme pack does this issue affect?

Fluent UI

### What is the public URL for the website?

_No response_

### Please describe the bug

Image

Dropping a file outside of the dropbox stuck in drop state, 
you can drop the file anywhere in the page not the Dropzone, here its the sendbox where the file should be dropped in.

In this scenario I dropped the file outside of the sendbox, then I am stuck at this state. 
This is a UI bug, alternatively I can refresh my chat or drop again.

```js
no
```

1. Navigate to ...
2. Click on ...
3. Type "..." in the send box

### What do you expect?

Allow adding a prop to specify the dragenter, dragleave, dragend events.

- document.addEventListener('dragenter', handleDragEnter, false);
- document.addEventListener('dragleave', handleDragLeave, false);
- document.addEventListener('dragend', handleDragEnd, false);

https://github.com/microsoft/BotFramework-WebChat/blob/c3857bf9fb4ba963c50ef32cd03bf9c70f1da70e/packages/fluent-theme/src/components/dropZone/DropZone.tsx

is adding a drag event on the entire document.

In sendbox maybe add a ref prop so that you can only drop at the send box and dropping outside will just reset dropZoneState back to false.
https://github.com/microsoft/BotFramework-WebChat/blob/c3857bf9fb4ba963c50ef32cd03bf9c70f1da70e/packages/fluent-theme/src/components/sendBox/SendBox.tsx#L8

### What actually happened?

Dropping a file outside of dropzone dropZoneState stays at droppable, even though no drag event.

### Do you have any screenshots or recordings to repro the issue?

Yes, you can drop the file anywhere outside the sendbox
Image

### Adaptive Card JSON

```json

```

### Additional context

_No response_

OEvgeny:
> Fixes #5390

### Fixed

-  Fixed [#5390](https://github.com/microsoft/BotFramework-WebChat/issues/5390). Fixed drop zone remaining visible when file is dropped outside of the zone, in PR [#5394](https://github.com/microsoft/BotFramework-WebChat/pull/5394), by [@OEvgeny](https://github.com/OEvgeny)

## Description

When dragging a file over the application and dropping it outside of the designated drop zone, the drop zone overlay remained visible instead of being hidden. This PR fixes the issue by properly handling the drop event at the document level, ensuring the drop zone state is reset regardless of where the file is dropped.

## Specific Changes

- Added document-level `drop` event listener with capture phase
- Removed `false` parameter from event listeners as it's the default behavior
- Added cleanup for the drop event listener in the useEffect cleanup function

## Review Checklist


compulim (commented):
Quick question. Do we need to handle specially if "drop outside of window"? I guess "drop" may not fire as well.

OEvgeny:
> Quick question. Do we need to handle specially if "drop outside of window"? I guess "drop" may not fire as well.

In browsers I tested dragend works well and drop also works when standard file handler is involved. The thing is: we need to listen to the drop outside of our drop zone elements, i.e. "global drop" as `dragend` is called only when no drop occurred.

compulim (approved):

PR #5430, opened 2025-02-25
[WIP] Conclude livestream with empty content

compulim:
> Fixes #

## -

-

## Review Checklist


__tests__/html2/basic/avatarBackgroundColor.html
```diff
@@ -0,0 +1,64 @@
...
+          WebChat: { ReactWebChat }
+        } = window; // Imports in UMD fashion.
+
+        const { directLine, store } = testHelpers.createDirectLineEmulator();
+
+        const styleOptions = {
+          botAvatarBackgroundColor: 'red',
+          botAvatarInitials: 'B',
+          userAvatarBackgroundColor: 'blue',
+          userAvatarInitials: 'TJ'
```
OEvgeny, line 27: :rofl:

__tests__/html2/hooks/useActiveTyping.variable.html
```diff
@@ -78,10 +77,7 @@
         // 3. t=10: Typed letter "b"
         // 4. t=10: Scheduled another typing indicator at t=3000
         // 5. t=3000: Send typing activity
-        clock.tick(3000);
-
-        // TODO: Understand why we need to yield after clock.tick and see if we could eliminate it.
-        await 0;
+        await directLine.actPostActivity(() => ponyfill.tick(3000));
```
OEvgeny, line 80: Did you understood?

__tests__/html2/livestream/activityOrder.html
```diff
@@ -37,15 +37,15 @@
           return false;
         };
 
-        const App = () => (
-          <Composer directLine={directLine} store={store}>
-            <BasicWebChat />
-            <Monitor />
-          </Composer>
+        render(
+          createElement(
```
OEvgeny, line 41: I'm ok with `createElement` but it's also common to use `h`

packages/api/src/providers/ActivityTyping/private/useReduceActivities.ts
```diff
@@ -0,0 +1,46 @@
+import { type WebChatActivity } from 'botframework-webchat-core';
+import { useActivities } from '../../../hooks';
+import useMemoWithPrevious from './useMemoWithPrevious';
+
+type Entry<T> = Readonly<{
+  activity: WebChatActivity;
+  value: T | undefined;
+}>;
+
+export default function useReduceActivities<T>(
```
OEvgeny, line 10: :heart:

OEvgeny (approved):

__tests__/html2/basic/avatarBackgroundColor.html
compulim: Direct port from old tests. 🤣

__tests__/html2/basic/longURL.keepAll.html
```diff
@@ -0,0 +1,9 @@
+<!doctype html>
+<html>
+  <head>
+    <script>
+      location = './longURL?break=keep-all&text=箸より重いものを持ったことがない箸より重いものを持ったことがない';
```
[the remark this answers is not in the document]
compulim: Done.

__tests__/html2/hooks/useActiveTyping.variable.html
compulim: Nope. But after I move from DLJS to DL Emulator, it is no longer needed. If I keep it, it will add 10ms to the test case.

__tests__/html2/livestream/raceBetweenTypingAndContentlessLivestream.html
```diff
@@ -0,0 +1,104 @@
...
+    <script crossorigin="anonymous" src="/test-harness.js"></script>
+    <script crossorigin="anonymous" src="/test-page-object.js"></script>
+    <script crossorigin="anonymous" src="/__dist__/webchat-es5.js"></script>
+  </head>
+  <body>
+    <main id="webchat"></main>
+    <script type="importmap">
+      {
+        "imports": {
+          "@testduet/wait-for": "https://esm.sh/@testduet/wait-for"
```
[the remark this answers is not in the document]
compulim: Done.

packages/api/src/providers/ActivityTyping/private/useReduceActivities.ts
```diff
@@ -0,0 +1,46 @@
...
+      });
+
+      // Returns the original array if nothing changed.
+      return changed ? nextState : state;
+    },
+    [activities, fn]
+  );
+
+  // eslint-disable-next-line no-magic-numbers
+  return state.at(-1)?.value;
```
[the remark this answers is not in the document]
compulim: If someone inserted an activity in the middleware of the array, we can start recalculation from the insertion point.

Assumes, we have a reducer which find the sum of all values.

The array is `[1, 2, 3, 4]`. The state is `[1, 3, 6, 10]`.

Then, someone added 2.5 between 2 and 3, so the array become `[1, 2, 2.5, 3, 3, 4]`.

Then, we only need to recompute from 2.5 because we remember the state for value 2 (which has value of 3). That means, the new state `[_, _, 5.5, 8.5, 12.5]` (`_` denotes state copied from previous array.)

That's why we keep every state but just return the very last value.

__tests__/html2/livestream/activityOrder.html
compulim: Let's keep it for now. I am wondering if we can go JSX Runtime with 16.8.6 by rebuilding `_jsx` function.

Thinking about using esm.sh/run with `<script type="text/babel">`.

packages/api/src/providers/ActivityTyping/private/useReduceActivities.ts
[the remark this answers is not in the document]
OEvgeny: Makes sense, so we need every value to be present in order we could do: `value: (prevValue = fn(prevValue, activity, index, activities))`

Another question. Looking at the code:
```js
      const nextState = activities.map<Entry<T>>((activity, index) => {
        const entry = state[+index];

        if (Object.is(entry?.activity, activity)) {
          prevValue = entry?.value;

          // Skips the activity if it has been reduced in the past render loop.
          return entry;
        }

        changed = true;

        return Object.freeze({
          activity,
          value: (prevValue = fn(prevValue, activity, index, activities))
        });
      });
  ```

What happens if we have a single activity updated? e,g.

```
before: a1, a2, a3, a4...
after: a1, a2', a3, a4...
```

OEvgeny (dismissed):
Approved with another question

OEvgeny (approved):

PR #5489, opened 2025-05-28
Move suggested actions from Redux store to React hooks

compulim:
### Changed

- Ported `useSuggestedActions` to use React hooks as backend instead of Redux store, in PR [#5489](https://github.com/microsoft/BotFramework-WebChat/pull/5489), by [@compulim](https://github.com/compulim)

## Description

In the future, we will deprecate the Redux store and move all state management into native React. This will enable us to work with the next generation of chat adapter which we incubated during our experimental work with Azure Communication Services.

The chat adapter work will make Web Chat cleaner on responsibility and reduce burden on handling different protocols.

This pull request is our first attempt to move suggested actions to React native state management, while keeping backward compatibility through data replication.

## Design

Each area of concern will be hosted in their own React context. For instance, `suggestedActions` and `suggestedActionsOriginActivity` is now in the new `<SuggestedActionsComposer>`.

#### Source of truth

The source of truth will be React context. This will help us to deprecate Redux store more easily in some future.

#### Redux action -> React context

All existing Redux action will be captured by "action sink" and sink into React context. The context component will then use React native state management (`useState` and `useReducer`) to save the value.

#### React context -> Redux action

When the state in React has changed, the new value will be replicated to Redux store via a special `WEB_CHAT_INTERNAL/*` action.

### Data consumption

Our React hooks will start reading from React context, instead of Redux `useSelector()`. The impact should be minimal as we should have enough tests to convert the surface area of our hooks. However, further minimize risks, we should continue to invest test on hooks.

Data changes from React will be replicated to Redux, web developers can continue to run their apps based on Redux `dispatch()` and `getState()`.

### Performance impact

As data will need to replicate from React to Redux, there will be both space and time penalty.

To minimize performance impact, web developers who do not use the Redux store can opt in and disable the data replication. This feature will be built in future pull requests.

## Specific Changes

- Created new `redux-store` package
- Moved `suggestedActions` reducer to the new `<SuggestedActionsComposer>` component
- Redirected `useSuggestedActions()` hook to use the new value from `<SuggestedActionsComposer>` component
- Created new `react-valibot` package to house `reactNode` and `validateProps`

## Review Checklist


packages/redux-store/src/suggestedActions/useSuggestedActionsHooks.ts
```diff
@@ -0,0 +1,7 @@
+import { useContext } from 'react';
+
+import SuggestedActionsContext, { type SuggestedActionsContextType } from './private/SuggestedActionsContext';
+
+export default function useSuggestedActionsHooks(): SuggestedActionsContextType {
+  return useContext(SuggestedActionsContext);
```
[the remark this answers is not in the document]
OEvgeny: > Given all best practices (using useRef, useCallback, and useMemo effectively), generally, I think the performance dip come from 2 things: useState (including useReducer) and useEffect. If useEffect is unavoidable, then majority performance dip lies within useState/useReducer. So let's focus on this first.

These are noticeable, they are usually inside of the component's code. I think our components are mostly fine in this matter: we have some blind spots, but usually not as drastically hitting perf as context updates.

When component depends on the context, when context updates either legit the update or not, it updates every component. In this sense the context is the same as prop, but it is hidden from the observer's eyes inside of hooks.

Maybe we could have a naming convention or something to mark such hooks, it could simplify the hide and seek game process.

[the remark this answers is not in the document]
compulim: <3

__tests__/html/hooks.useCreateActivityStatusRenderer.html
```diff
@@ -51,6 +51,7 @@
         );
 
         await pageConditions.webChatRendered();
+        await pageConditions.uiConnected();
```
compulim, line 54: Without this, we would send the incoming activity before `connectionStatus$` turn into `ONLINE`.

__tests__/html2/hooks/private/renderHook.js
```diff
@@ -67,10 +67,7 @@ export default function renderHook(
     return { rerender: render, unmount: () => ReactDOM.unmountComponentAtNode(element) };
   };
 
-  const { rerender: baseRerender, unmount } = render(
-    React.createElement(TestComponent, { renderCallbackProps: initialProps }),
-    renderOptions
-  );
+  const { rerender: baseRerender, unmount } = render({ renderCallbackProps: initialProps });
```
compulim, line 70: `initialProps` was never working before.

packages/test/page-object/src/globals/testHelpers/createDirectLineEmulator.js
```diff
@@ -3,10 +3,10 @@ import random from 'math-random';
 import updateIn from 'simple-update-in';
 
 import createDeferredObservable from '../../utils/createDeferredObservable';
+import withResolvers from '../../utils/withResolvers';
```
OEvgeny, line 6: Can use from base package now

packages/core/src/index.ts
```diff
@@ -46,6 +44,20 @@ import onErrorResumeNext from './utils/onErrorResumeNext';
...
 import warnOnce from './utils/warnOnce';
 
+export {
+  CLEAR_SUGGESTED_ACTIONS,
+  default as clearSuggestedActions,
+  clearSuggestedActionsActionSchema,
+  type ClearSuggestedActionsAction
+} from './actions/clearSuggestedActions';
+
+export {
```
compulim, line 54: We didn't expose the `setSuggestedActions` previously. 🤣

packages/api/src/hooks/useSuggestedActions.ts
```diff
@@ -1,37 +1,5 @@
...
-
-        clearSuggestedActions();
-      },
-      [clearSuggestedActions]
-    ),
-    { activity }
-  ];
+export default function useSuggestedActions() {
+  return useSuggestedActionsHooks().useSuggestedActions();
 }
```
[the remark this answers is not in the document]
OEvgeny: See the PR https://github.com/compulim/BotFramework-WebChat/pull/5

packages/component/src/index.ts
```diff
@@ -65,9 +65,9 @@ const hooks = {
   ...componentHooks
 };
 
-const buildTool = process.env.build_tool;
-const moduleFormat = process.env.module_format;
-const version = process.env.npm_package_version;
+const buildTool = (process as any).env.build_tool;
+const moduleFormat = (process as any).env.module_format;
+const version = (process as any).env.npm_package_version;
```
OEvgeny, line 70:
```suggestion
declare var process: {
  env: {
    build_tool: string;
    module_format: string;
    npm_package_version: string;
  }
};

const buildTool = process.env.build_tool;
const moduleFormat = process.env.module_format;
const version = process.env.npm_package_version;
```

compulim: This is good. I was using `declare global {}` and causing conflicts.

OEvgeny (approved):

PR #5513, opened 2025-07-10
fix: treat activity with abstract as non contentless activity

OEvgeny:
> Fixes #

## -

-

## Review Checklist


packages/component/src/Activity/StackedLayout.tsx
```diff
@@ -287,7 +288,7 @@ const StackedLayout = ({
         renderBubbleContent={renderBubbleContent}
         showCallout={showCallout}
       >
-        {!isCollapsible && attachmentChildren}
+        {!isCollapsible && attachmentChildren.length > 0 && attachmentChildren}
```
compulim, line 291: Should it be...?

```suggestion
        {!isCollapsible && attachmentChildren?.length > 0}
```

packages/core/src/utils/getActivityLivestreamingMetadata.ts
```diff
@@ -47,8 +47,8 @@ const livestreamingActivitySchema = union([
       streamType: literal('informative')
     }),
     id: string(),
-    // Informative message must have "text".
-    text: string(),
+    // Informative may not have "text", but should have abstract instead (checked later)
```
compulim, line 50: If it should have abstract, should we add a check here (later)?

OEvgeny: I haven't figured the way as it makes everything in union kinda harder to work with. We do have the check later.

I'll look into how to improve this.

packages/component/src/Activity/StackedLayout.tsx
OEvgeny: Given `attachmentChildren` is a memo and we never return anything besides `[]` I think we can leave this the way it is.

compulim: Oh right, it's returning `attachmentChildren`.

PR #5515, opened 2025-07-16
Introduction of `polymiddleware`

compulim:
### Breaking changes

- `activityMiddleware` is being deprecated in favor of [`polymiddleware`](./docs/MIDDLEWARE.md). It will be removed on or after 2027-08-16, related to PR [#5515](https://github.com/microsoft/BotFramework-WebChat/pull/5515)

### Added

- Added `polymiddleware`, a new [universal middleware for every UIs](./docs/MIDDLEWARE.md), by [@compulim](https://github.com/compulim) in PR [#5515](https://github.com/microsoft/BotFramework-WebChat/pull/5515)
   - Added `polymiddleware` to `<ThemeProvider>`
   - Currently supports activity middleware and the new error box middleware
- New internal packages, by [@compulim](https://github.com/compulim) in PR [#5515](https://github.com/microsoft/BotFramework-WebChat/pull/5515)
   - `@msinternal/botframework-webchat-api-middleware` for middleware branch of API package
   - `@msinternal/botframework-webchat-debug-theme` package for enabling debugging scenarios
   - `@msinternal/botframework-webchat-react-hooks` for helpers for React hooks

### Removed

- `activityMiddleware` is being deprecated in favor of [`polymiddleware`](./docs/MIDDLEWARE.md). It will be removed on or after 2027-08-16, related to PR [#5515](https://github.com/microsoft/BotFramework-WebChat/pull/5515)

## Description

Web Chat is becoming an UI orchestrator.

We are introducing the very first step: a new universal middleware called "polymiddleware".

As we enable different types of UI customization, we also introduced many middleware types, just to name a few: activity, activity grouping activity status, attachment, attachment for screen reader, card action.

It is increasingly difficult to manage: different API signature to suit different needs, multiple props to pass, not theme pack friendly.

In this pull request, we are introducing polymiddleware, an universal middleware that can handle all types of UI customization, including customizations we didn't introduce yet.

```ts
import ReactWebChat, { createStoreWithOptions } from 'botframework-webchat';
import { type WebChatActivity } from 'botframework-webchat-core';
import {
   activityComponent,
   createActivityPolymiddleware,
   type ActivityPolymiddlewareProps
} from 'botframework-webchat/middleware';
import React, { memo } from 'react';
import { render } from 'react-dom';

type ChannelMessageProps = ActivityPolymiddlewareProps & {
   readonly activity: WebChatActivity & {
      type: 'message';
   };
};

const ChannelMessage = memo<ChannelMessageProps>(function ChannelMessage({ activity }) {
   return (
      <div className="channel-message">
         <div className="channel-message__body">{activity.text}</div>
      </div>
   );
});

const polymiddleware = [
   createActivityPolymiddleware(next => request => {
      const { activity } = request;

      if (activity.from.role === 'channel' && activity.type === 'message') {
         return activityComponent(ChannelMessage, { activity });
      }

      return next(request);
   })
];
```

## Design

Polymiddleware is very similar to traditional middleware.

### `init` as the enhancer selector

A normal middleware API: `init => next => request => next(request)`. FWIW, enhancer is `next => request => next(request)`.

A polymiddleware for activity: `init => init === 'middleware' && (next => request => next(request))(props)`. It returns `false` if it is initialized with "a topic it is not interested", otherwise, return the enhancer.

### Handler result: render function vs. component

We settled to use render function for handler result.

- Render function can perform "bind props" with proper memoization
   - When binding props to a component, the component must be re-rendered
   - When binding props with same contents to the same component, it introduce wasted rendering
- Render function can be converted into component via `<Fragment>{renderFn()}</Fragment>`

Notes: "bind props" is for prepopulating props and conceptually same as the `Function.bind()` (bind arguments) in JavaScript. When the component is rendered, it will use the props bound.

To reduce confusion, all polymiddleware must use a set of APIs to instantiate the polymiddleware and its result.

### Versionable APIs

For backward and forward compatibility, all polymiddleware are encapsulated behind set of APIs to enable versioning. The following example show the activity polymiddleware APIs:

```ts
function MyActivityComponent(props) {
  return <div>{props.activity.text}</div>;
}

const polymiddleware = [
  createActivityPolymiddleware(request => next => {
    const props = { activity: request.activity };

    return activityComponent(MyActivityComponent, props);
  })
];
```

Every type of middleware will have their own set of APIs. Therefore, they can be individually versioned and reduce frictions for their specific use case.

### Bridging legacy activity middleware

Legacy activity middleware has a few characteristics:

- Request: `activity`, `renderAttachment`
- Props: `hideTimestamp`, `renderActivityStatus`, `renderAvatar`, `showCallout`

Except `activity`, other request/props can be now obtained through hooks. However, legacy activity middleware still requires them.

We built the bridge middleware that would obtain those values through hooks and call legacy middleware. However, due to signature differences, the legacy middleware cannot call `next(request)`. Thus, the legacy activity middleware cannot have downstreamers of modern activity polymiddleware, which is actually okay.

### Theme-ability

A new `polymiddleware` props is introduced to `<ThemeProvider>` to enable cascade. The priority:

- Polymiddleware closer to the leaf node is preferred: "deeper win"
- Polymiddleware always win over legacy middleware: "newer win"

### Error boundary

Error boundary is now a top-level decorative polymiddleware. It is purely decorative means it will always render downstreamers and is a perfect catch-all for errors catchable by React error boundary.

The new error boundary polymiddleware is opt-in on a per type basis set by Web Chat team.

### New error box middleware and a new debug theme

Before this PR, we are using `node_env` to show/hide error box (a.k.a. red box.) This is only useful for web developers who import Web Chat via NPM and running the code in dev server. It is not available on bundle because bundle is always running in production mode.

In this PR, we are introducing a new (internal) debug theme and a new error box middleware. The theme will add the red box as a error box middleware.

When the theme is enabled and error is caught, the red box will be shown if the debug theme is installed.

The debug theme is currently for internal use. In future, we will have some ways to enable the debug theme in your environment.

## Specific Changes

- Deprecating `activityMiddleware` in favor of the new `polymiddleware`
   - Bridge middleware is introduced in `polymiddleware` and it will call the deprecated `activityMiddleware`
- Added new `debug-theme` to add error box, instead of using `node_env === 'development'`
- Added new `ErrorBoxPolymiddleware` and added as a prioritized decorative middleware
- Bumped `react-chain-of-responsibility`

## Review Checklist


packages/component/src/Middleware/Activity/private/LegacyActivityBridge.tsx
```diff
@@ -83,23 +88,22 @@ const TranscriptActivity = ({ activity, renderActivity }: TranscriptActivityProp
     showCallout = true;
   }
 
-  const children = useMemo(
+  const node = useMemo(
     () =>
-      renderActivity({
+      render(renderAttachment, {
```
OEvgeny: Does this mean we can remove the whole `*activityWithRenderer` thing? Huge if true.

packages/middleware/src/PolyMiddlewareComposer.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+type PolyMiddlewareComposerProps = Readonly<InferInput<typeof polyMiddlewareComposerPropsSchema>>;
+
+function PolyMiddlewareComposer(props: PolyMiddlewareComposerProps) {
+  const { children, middleware } = validateProps(polyMiddlewareComposerPropsSchema, props);
+
+  const activityPolyMiddleware = useMemo(() => extractActivityPolyMiddleware(middleware), [middleware]);
+
+  return (
+    <ActivityPolyMiddlewareProvider middleware={activityPolyMiddleware}>{children}</ActivityPolyMiddlewareProvider>
```
OEvgeny: Doesn't it make sense to split poly middleware provider from the RCoR provider? This way we could have all middleware in a single provider until we get to the proxy component rendering. We'll need to wrap RCoR Proxy into the RCoR provider, but this way we get leaner tree in case when multiple PolyMiddlewareComposers are used.

packages/middleware/src/internal/createActivityPolyMiddlewareFromLegacy.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    return ({ activity }) => {
+      // TODO: `nextVisibleActivity` is deprecated and should be removed.
+      const legacyResult = legacyHandler({ activity, nextVisibleActivity: undefined as any });
+
+      if (!legacyResult) {
+        return undefined;
+      }
+
+      return activityComponent(bridgeComponent, { activity, render: legacyResult });
```
OEvgeny: This might be hooks mismatch due to:
- on the first pass `legacyResult` is `false`
- on the second pass it flips, so we call the `activityComponent`

packages/middleware/src/private/templateMiddleware.tsx
```diff
@@ -0,0 +1,153 @@
...
+    Request,
+    Props,
+    string
+  >();
+
+  type TemplatedEnhancer = ReturnType<InferOrganicMiddleware<typeof Provider>>;
+  type TemplatedMiddleware = (init: string) => TemplatedEnhancer;
+
+  const createMiddleware = (enhancer: TemplatedEnhancer): TemplatedMiddleware => {
+    const factory: TemplatedMiddleware = init => (init === name ? enhancer : BYPASS_ENHANCER);
```
OEvgeny: Nice trick to not filter out things :rofl:

compulim: Yea, but may be some time in the future we would want to filter out instead of running 100x `BYPASS_ENHANCER`.

packages/middleware/src/PolyMiddlewareComposer.tsx
compulim: To have a single provider, I think the best move is to move `init`/`type` inside `request`, and have a wrapper enhancer wrapping the userland enhancer.

The wrapper enhancer would bypass the request if the `init`/`type` doesn't match.

In this way, we still keep the original `init`, `request`, and `props` working as they are designed. Init for init-time, request for build-time, props for render-time.

packages/component/src/Middleware/Activity/private/LegacyActivityBridge.tsx
compulim: I don't think so... `activityWithRenderer` run the enhancer and give us an idea which activities should be rendered or hidden.

It's an intended 2-pass mechanism.

To get rid of this, we need another mechanism to filter out "which activities will be rendered or hidden." Such as rules like "always render message activity and drop event activity."

OEvgeny: Why don't we move that part into the enhancer (before the user ones)?

__tests__/html/focusManagement.disableHeroCard.obsolete.html
```diff
@@ -15,14 +15,29 @@
...
+          hooks: { useActivities }
         }
       } = window;
 
       run(async function () {
         const store = testHelpers.createStore();
         const directLine = WebChat.createDirectLine({ token: await testHelpers.token.fetchDirectLineToken() });
 
+        const AutoDisableHeroCardContent = ({ activity, attachment }) => {
+          const [activities] = useActivities();
```
compulim, line 28: This is the correct way, use hooks to detect if the activity is the most recent one.

We should not relies on middleware because it could be skipped due to caching.

packages/api/src/middleware/private/templateMiddleware.tsx
```diff
@@ -58,15 +59,30 @@ function templateMiddleware<Request, Props extends {}>(name: string) {
...
-  const { Provider, Proxy } = createChainOfResponsibility<Request, Props>();
+  const { Provider, Proxy } = createChainOfResponsibility<Request, Props, string>();
+
+  type TemplatedProviderProps = {
+    readonly children?: ReactNode | undefined;
+    readonly middleware: readonly Middleware[];
+  };
+
+  // eslint-disable-next-line prefer-arrow-callback
+  const TemplatedProvider = memo(function TemplatedProvider({ children, middleware }: TemplatedProviderProps) {
```
compulim, line 73: For better typing for existing stuff, we are binding props to `Provider` with `{ init: name }`.

packages/component/src/Composer.tsx
```diff
@@ -441,6 +447,15 @@ const InternalComposer = ({
     [sendBoxToolbarMiddlewareFromProps, theme.sendBoxToolbarMiddleware]
   );
 
+  const polyMiddlewareArray = useMemo<readonly PolyMiddleware[]>(
+    () =>
+      Object.freeze([
+        // TODO: Add <FallbackComponent>.
```
compulim: NTM: Add test for red box here.

packages/component/src/Middleware/Activity/private/LegacyActivityBridge.tsx
compulim: I mean, that part (decide whether to show/hide activity) is in the enhancer already.

We run the enhancer, returning an array of (visible) activities. Then we run the grouping middleware. Then we have the 2D array. Then finally, we render everything out, IIRC.

So, it's a multi-pass and we can't get rid of `useActivityWithRenderer` idea... we can't do everything in a single pass I believe, but I could be wrong.

packages/component/src/providers/RenderingActivities/useActivityRendererMap.ts
```diff
@@ -0,0 +1,9 @@
+import { type ActivityPolyMiddlewareRenderer } from '@msinternal/botframework-webchat-middleware';
+import { type WebChatActivity } from 'botframework-webchat-core';
+import { useRenderingActivitiesContext } from './private/RenderingActivitiesContext';
+
+export default function useActivityRendererMap(): readonly [
```
compulim, line 5: I know we should reduce unnecessary hooks, let's do it separately.

__tests__/html2/middleware/activity/hooks/useBuildRenderActivityCallback/simple.html
```diff
@@ -34,40 +34,25 @@
         const {
           testHelpers: { createDirectLineEmulator },
           WebChat: {
-            Components: { Composer },
-            hooks: { useActivities, useBuildRenderActivityCallback }
+            hooks: { useActivities, useBuildRenderActivityCallback },
+            ReactWebChat
           }
```
OEvgeny, line 37: Doesn't this work with just Composer anymore?

compulim: Ha, I retouched this test case so many times it is not testing anything related to poly middleware now. 🤣

It works with `<Composer>`, but the key part is that if we want to render timestamp in the test, we need to pass `renderActivityStatus()` in the props.

We were doing exactly the same thing as if yesterday we are still on the `useCreateActivityRenderer()` hook. But we don't have much tests on old `uCAR()`.

So... why not testing using `<Composer>` is because I also want to make sure we are passing `renderActivityStatus()` properly. We are auto-building this prop and other legacy props in (the descendants of) `<BasicTranscript>`, so no one need to send this in. I want to make sure we are sending this in properly. So we need to use `<ReactWebChat>`/`<BasicTranscript>` than `<Composer>`.

OEvgeny: My concern is that we may hit the case when we're not able to run render with only Composer. I did that mistake recently and there are not much tests failing :rofl:

packages/react-valibot/src/validateProps.ts
```diff
@@ -42,6 +42,12 @@ export default function validateProps<const TSchema extends BaseSchema<unknown,
   props: InferInput<TSchema>,
   isolationMode?: IsolationMode | undefined
 ): InferInput<TSchema> | InferOutput<TSchema> {
+  // When an error boundary caught an error, React will render some components again with `props` of `undefined`.
+  if (typeof props === 'undefined') {
```
OEvgeny, line 46: Thanks for fixing this. The error is very cryptic :rofl:

packages/middleware/src/internal/createActivityPolyMiddlewareFromLegacy.tsx
compulim: No able to repro this now. Guess if it's related to error boundary thing we talk offline.

packages/api-middleware/src/private/createErrorBoundaryMiddleware.ts
```diff
@@ -41,14 +42,14 @@ export default function createErrorBoundaryMiddleware<Request, Props extends {}>
 
       // Thrown before render, show the red box immediately.
       return reactComponent<
-        // TODO: [P*] Fix unknown.
+        // Need to be unknown here because memo() hid the generic type.
```
OEvgeny: You can supply the type to memo e.g. memo<T>() if it helps.

__tests__/html2/middleware/errorBox/simple.noDebugProvider.html.snap-1.png
OEvgeny, line 1: Maybe add some CSS, something like:
```css
.webchat__transcript:empty {
  display: grid;
}

.webchat__transcript:empty::after {
  content: 'No activities';
  place-self: center;
}
```

Maybe even add it to the debug theme itself.

__tests__/html2/middleware/activity/hooks/useBuildRenderActivityCallback/simple.html
compulim: Right, added test `renderViaComposer.html`.

__tests__/html2/middleware/errorBox/simple.noDebugProvider.html.snap-1.png
compulim: Added a red box.

It's actually rendering the `<SUT>` component without any content.

__tests__/html2/samples/middleware/channelMessage.tsx
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable prefer-arrow-callback */
+
+// Use code-splitting to speed up development time, just dummy import is sufficient.
+// esbuild will split out botframework-webchat* packages into their own chunks.
+await import('botframework-webchat');
+await import('botframework-webchat/middleware');
+
+// #region Sample code
```
compulim, line 8: In some future, we should copy this region into the sample readme.

docs/MIDDLEWARE.md
```diff
@@ -0,0 +1,122 @@
+# Middleware
```
compulim, line 1: Need more time to write this up, maybe in another PR.

packages/api-middleware/src/private/createErrorBoundaryMiddleware.ts
compulim: No, it doesn't. 🤣

We don't know TypeScript enough.

packages/bundle/src/boot/bundle/full-es5.ts
```diff
@@ -16,7 +17,8 @@ const { createStoreWithDevTools: _createStoreWithDevTools, ...finalES5 } = es5;
 window['WebChat'] = Object.freeze({
```
compulim, line 13: These "boot" files are still quite ugly. Need another push to clean them up.

packages/react-valibot/package.json
```diff
@@ -41,7 +41,7 @@
     "precommit:eslint": "../../node_modules/.bin/eslint --report-unused-disable-directives --max-warnings 0",
     "precommit:typecheck": "tsc --project ./src --emitDeclarationOnly false --esModuleInterop true --noEmit --pretty false",
     "preversion": "cat package.json | jq '(.localDependencies // {} | to_entries | map([if .value == \"production\" then \"dependencies\" else \"devDependencies\" end, .key])) as $P | delpaths($P)' > package-temp.json && mv package-temp.json package.json",
-    "start": "npm run build -- --watch"
+    "start": "npm run build -- --onSuccess=\"touch ../api/src/index.ts ../api-middleware/src/index.ts ../component/src/index.ts ../redux-store/src/index.ts\" --watch"
```
compulim, line 44: In future, we should have a way to detect dependents.

package.json
```diff
@@ -26,21 +26,25 @@
...
     "./packages/fluent-theme"
   ],
   "scripts": {
     "audit-all": "find -name package-lock.json | xargs dirname | xargs -I {} sh -c 'cd {} && npm audit'",
     "biome": "biome check ./packages",
     "browser": "node ./packages/test/harness/src/host/dev/index http://localhost:5001/__tests__/html/",
     "browser:watch": "node-dev --no-notify --respawn ./packages/test/harness/src/host/dev/index http://localhost:5001/__tests__/html/",
-    "build": "npm run build --if-present --workspaces",
+    "build": "npm run build:production && npm run build:html2-samples",
+    "build:html2-samples": "cd __tests__/html2/samples/ && esbuild --bundle --format=esm --outbase=. --outdir=./dist/ --minify **/*.tsx --splitting",
```
compulim, line 47: Will be great if we have a way to write testable sample without esbuild. The testable sample should be machine copyable to `MIDDLEWARE.md`.

packages/api-middleware/src/PolymiddlewareComposer.tsx
```diff
@@ -0,0 +1,84 @@
...
+      // Checks for array equality, return previous version if nothing has changed.
+      return prevActivityEnhancers.length === activityEnhancers.length &&
+        activityEnhancers.every((middleware, index) => Object.is(middleware, prevActivityEnhancers.at(index)))
+        ? prevActivityEnhancers
+        : activityEnhancers;
+    },
+    [polymiddleware]
+  );
+
+  const activityPolymiddleware = useMemo(() => activityEnhancers.map(enhancer => () => enhancer), [activityEnhancers]);
```
OEvgeny, line 50: You can save on the memo call, but I'm ok with this

packages/api-middleware/src/PolymiddlewareComposer.tsx
```diff
@@ -0,0 +1,84 @@
...
+  //    - If props.middleware changed and only one middleware changed, we could cache other middleware types
+  // 2. Single <Provider>
+  //    - `useBuildRenderCallback()` should pre-build (apply) the callback (multiple middleware -> single enhancer)
+  //    - The callback will be invalidated on middleware (of target type) change
+  //    - <Proxy> will need to be rebuilt, as it use a different `useBuildRenderCallback()`
+
+  return (
+    <ActivityPolymiddlewareProvider middleware={activityPolymiddleware}>
+      <ErrorBoxPolymiddlewareProvider middleware={errorBoxPolymiddleware}>{children}</ErrorBoxPolymiddlewareProvider>
+    </ActivityPolymiddlewareProvider>
```
OEvgeny, line 79: We'll end up with lots of nested providers, but I'm ok with that

packages/api-middleware/src/activityPolymiddleware.tsx
```diff
@@ -0,0 +1,92 @@
...
+  object({
+    activity: custom<Readonly<WebChatActivity>>(value => safeParse(object({}), value).success)
+  }),
+  readonly()
+);
+
+type ActivityPolymiddlewareProxyProps = Readonly<InferInput<typeof activityPolymiddlewareProxyPropsSchema>>;
+
+// A friendlier version than the organic <Proxy>.
+const ActivityPolymiddlewareProxy = memo(function ActivityPolymiddlewareProxy(props: ActivityPolymiddlewareProxyProps) {
```
OEvgeny, line 44: You already have custom proxy. Not sure what prevents you from pulling middleware here and calling `useBuildRenderActivityCallback` instead of introducing another layer of wrapping things.

packages/api-middleware/src/activityPolymiddleware.tsx
```diff
@@ -0,0 +1,92 @@
...
+            createErrorBoundaryMiddleware({
+              createMiddleware: createActivityPolymiddleware,
+              reactComponent: activityComponent,
+              where: 'Activity polymiddleware'
+            }),
+            ...middleware
+          ]),
+        [middleware]
+      )}
+    >
```
OEvgeny:
```suggestion
  const middleware = useMemo(
        () =>
          Object.freeze([
            createErrorBoundaryMiddleware({
              createMiddleware: createActivityPolymiddleware,
              reactComponent: activityComponent,
              where: 'Activity polymiddleware'
            }),
            ...middleware
          ]),
        [middleware]
      );

  return (
    <Provider
      // Decorates middleware with <ErrorBoundary>.
      middleware={middleware}
    >
```

I'm ok with the original. It just looks a bit cumbersome

packages/api-middleware/src/errorBoxPolymiddleware.tsx
```diff
@@ -0,0 +1,68 @@
...
+    error: unknown(),
+    where: string()
+  }),
+  readonly()
+);
+
+type ErrorBoxPolymiddlewareProxyProps = Readonly<InferInput<typeof ErrorBoxPolymiddlewareProxyPropsSchema>>;
+
+// A friendlier version than the organic <Proxy>.
+const ErrorBoxPolymiddlewareProxy = memo(function ErrorBoxPolymiddlewareProxy(props: ErrorBoxPolymiddlewareProxyProps) {
```
OEvgeny, line 45: Maybe fold this into template if you have to repeat it?

packages/api-middleware/src/internal/useMemoWithPrevious.ts
```diff
@@ -0,0 +1,17 @@
+// TODO: [P0] Dedupe and put it under api-tools.
```
OEvgeny: Maybe handle P0? We need a better way to handle P0, otherwise it may stay for years.

packages/api-middleware/src/private/createErrorBoundaryMiddleware.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+      >(
+        ErrorBoundaryForRenderFunction,
+        // TODO: [P0] Fix any, consider using React context to add extraneous props.
+        {
+          errorBoundaryRenderFunction: () => {
+            throw error;
+          },
+          errorBoundaryWhere: where
+        } as any
```
OEvgeny: You need to properly cast this `as unknown as Something` at least

packages/bundle/src/boot/actual/middleware.ts
```diff
@@ -0,0 +1,36 @@
+// TODO: [P1] Unsure why export * will not export anything for /__tests__/html2/samples.
+
+// ```
+// ▲ [WARNING] Import "activityComponent" will always be undefined because the file "../../../packages/bundle/dist/botframework-webchat.middleware.mjs" has no exports [import-is-undefined]
+// ```
+
+// export * from 'botframework-webchat-api/middleware';
```
OEvgeny: Anyway it is a good idea to explicitly state what you're exporting

packages/bundle/src/boot/exports/middleware.ts
```diff
@@ -0,0 +1 @@
+export * from '../actual/middleware';
```
OEvgeny, line 1: Why not directly exporting here from the /middleware?

OEvgeny (approved):

packages/api-middleware/src/PolymiddlewareComposer.tsx
compulim: Let's keep this one, I tried but not trivial to do to the equality check in `useMemoWithPrevious` unless we save the `enhancer` function as a hidden variable inside `() => enhancer`.

packages/api-middleware/src/PolymiddlewareComposer.tsx
compulim: Yea, this need some bold changes as laid out in the comment, change both `useBuildRenderXXXCallback` and `XXXProxy` (to use the new `useBuildRenderXXXCallback`.)

Maybe when we reach 5-10 types of polymiddleware, we should start looking at this. But at least, I have test to make sure "changing error box in `polymiddleware` should not re-render activities."

packages/api-middleware/src/activityPolymiddleware.tsx
compulim: It's just time. 🤣 But I agree we could extract a specific type of middleware in the `uBRXC` and `XXXProxy`.

One thing I didn't think through, if we extract it in `uBRXC` (which `XXXProxy` depends on), that means we need to traverse the whole `polymiddleware` prop on every new instance of `XXXProxy`. Then it could hurt performance unless we have a `<XXXProvider>` to cache that result?

Any thoughts?

packages/api-middleware/src/errorBoxPolymiddleware.tsx
compulim: This is primarily for renaming prop to make `<XXXProxy>` more usable than `<Proxy request={}>`. Maybe a "rename prop" function? 🤣 But I bet it doesn't help much in terms of number of line of code.

packages/api-middleware/src/private/createErrorBoundaryMiddleware.ts
compulim: Will try it again today. IIRC, it's about the signature inside `errorBoundaryRenderFunction`, not about the props.

packages/bundle/src/boot/exports/middleware.ts
compulim: I am thinking about the hierarchy of the `boot` folder, definitely worth another clean up task:

- `/exports/` is named exports. More-or-less purely `export * from '../actual/xxx`
- `/bundle/` is bundle. More-or-less `window.WebChat['xxx'] = import('../actual/xxx')`

Then, `/actual/` is the actual code. My goal is to keep exports and bundle exactly the same.

Any thoughts?

packages/api-middleware/src/activityPolymiddleware.tsx
OEvgeny: Maybe use useMemoized for such cases. I didn't think about this either.

That said this might be implemented for decorator as well

packages/api-middleware/src/errorBoxPolymiddleware.tsx
OEvgeny: I mean you can supply the name inside of the template

packages/bundle/src/boot/exports/middleware.ts
OEvgeny: Can't we make bundle a special case of exports maybe?

I mean why we need actual?

OEvgeny (dismissed):

__tests__/html2/middleware/activity/legacyActivityMiddleware/hideTimestamp.html
```diff
@@ -0,0 +1,90 @@
...
+          createElement(ReactWebChat, {
+            activityMiddleware,
+            directLine,
+            store
+          })
+        );
+
+        await pageConditions.uiConnected();
+
+        // WHEN: Two messages arrives: "Hello, World!" and "Aloha!"
```
xTEddie: nit: "Two messages arrive:"

__tests__/html2/middleware/activity/legacyActivityMiddleware/showCallout.html
```diff
@@ -0,0 +1,105 @@
...
+              botAvatarInitials: 'WC',
+              // Intentionally show avatar from below.
+              bubbleNubOffset: 'bottom'
+            }
+          })
+        );
+
+        await pageConditions.uiConnected();
+
+        // WHEN: Two messages arrives: "Hello, World!" and "Aloha!"
```
xTEddie: nit: "Two messages arrives" -> "Two messages arrive"

__tests__/html2/middleware/activity/invalidMiddleware.createdOutsideOfFactory.html
```diff
@@ -0,0 +1,96 @@
...
+          createElement(ReactWebChat, {
+            directLine,
+            polymiddleware,
+            store
+          })
+        );
+
+        await pageConditions.uiConnected();
+
+        // WHEN: A messages arrive "Hello, World!".
```
xTEddie: nit: "A messages arrive" -> "A message arrives"

packages/api-middleware/src/private/templatePolymiddleware.check.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  expect(warn).toHaveBeenNthCalledWith(1, expect.stringContaining('must be an array of function'));
+});
+
+test('should throw if middleware is not a function', () => {
+  const template = templatePolymiddleware('Check' as any);
+
+  expect(() => template.createMiddleware(1 as any)).toThrow('enhancer must be of type function');
+});
+
+test('should not warn if middleware return false', () => {
```
xTEddie: nit: "return" -> "returns"

packages/api-middleware/src/private/templatePolymiddleware.check.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const template = templatePolymiddleware('Check' as any);
+
+  jest.spyOn(console, 'warn').mockImplementation(warn);
+
+  template.extractEnhancer([template.createMiddleware(() => false as any)]);
+
+  expect(warn).toHaveBeenCalledTimes(0);
+});
+
+test('should not warn if middleware return function', () => {
```
xTEddie: nit: "return" -> "returns"

__tests__/html2/middleware/activity/legacyActivityMiddleware/showCallout.html
compulim: Fixed.

packages/api-middleware/src/private/templatePolymiddleware.check.test.tsx
compulim: Fixed.

packages/api-middleware/src/private/templatePolymiddleware.check.test.tsx
compulim: Fixed.

__tests__/html2/middleware/activity/legacyActivityMiddleware/hideTimestamp.html
compulim: Fixed.

__tests__/html2/middleware/activity/invalidMiddleware.createdOutsideOfFactory.html
compulim: Fixed.

packages/api-middleware/src/PolymiddlewareComposer.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const errorBoxPolymiddleware = useMemo(() => errorBoxEnhancers.map(enhancer => () => enhancer), [errorBoxEnhancers]);
+
+  // Didn't thoroughly think through this part yet, but I am using the first approach for now:
+
+  // 1. <XXXProvider> for every type of middleware
+  //    - If props.middleware changed and only one middleware changed, we could cache other middleware types
+  // 2. Single <Provider>
+  //    - `useBuildRenderCallback()` should pre-build (apply) the callback (multiple middleware -> single enhancer)
+  //    - The callback will be invalidated on middleware (of target type) change
+  //    - <Proxy> will need to be rebuilt, as it use a different `useBuildRenderCallback()`
```
xTEddie: nit: "it use" -> "it uses"

packages/api/src/hooks/Composer.tsx
```diff
@@ -87,7 +91,9 @@ import observableToPromise from './utils/observableToPromise';
 
 // PrecompileGlobalize is a generated file and is not ES module. TypeScript don't work with UMD.
```
xTEddie: nit: "TypeScript don't" -> "TypeScript doesn't"

packages/api-middleware/src/activityPolymiddleware.tsx
compulim: `useMemoized` doesn't help because every `useBuildRenderXXXCallback` and `<Proxy>` is unique and unrelated to each other.

Consider... `<BasicTranscript>` is rendering a lot of `<ActivityProxy>`. Without context, each proxy component don't know each other, so we can't cache the list of extracted middleware.

But maybe the "extraction" is not as expensive as we thought, I don't know the number yet.

It seems safer to have extraction done in `<Provider>` as it's O(n), than doing it here which is similar to O(n^2) (every activity added, will instantiate <Proxy> which loop over every middleware.)

packages/api-middleware/src/PolymiddlewareComposer.tsx
compulim: Fixed.

packages/api/src/hooks/Composer.tsx
compulim: Fixed.

packages/api-middleware/src/private/createErrorBoundaryMiddleware.ts
compulim: Fixed used `as unknown as`, not elegant, but that's how much we can go given 2 things:

- Generic type argument is used in a class component
- `memo()` hide a lot of typing details

I have a better/cleaner way in my head but need time to brew.

packages/api-middleware/src/activityPolymiddleware.tsx
OEvgeny: All you have to do is to lift `getMiddlewareOfType` function on the same context the middleware array comes from. The function wrapped with `useMemoized` should be able to cache results for multiple Proxy instances. Each Proxy would use it from the context instead of fetching the list of middleware.

packages/bundle/src/boot/exports/middleware.ts
compulim: Cleaned up the boot code a bit so they are kind of uniform and symmetric.

Both `bundle/*` and `exports/*` will call `addVersion()` but the `buildInfo` is slightly different. So `bundle/*` cannot depends on `exports/*`, otherwise, it will inject version twice.

Plus, `bundle` will export as `window.WebChat.middleware`, while we have `exports/full` and `exports/middleware` in separated named exports.

It is symmetric now:

- `actual/full-es5` depends on `actual/full` depends on `actual/minimal`
- `(bundle|exports)/minimal` depends on `actual/minimal`
- `(bundle|exports)/full` depends on `actual/full`
- `(bundle|exports)/full-es5` depends on `actual/full-es5`

I could reduce more code if `(bundle|exports)/full-es5` depends on `(bundle|exports)/full` instead of `actual/full-es5`, but this breaks symmetricity.

packages/api-middleware/src/errorBoxPolymiddleware.tsx
compulim: Don't quite get it. Any pseudo code?

packages/api-middleware/src/activityPolymiddleware.tsx
compulim: So it will be more like lifting the extraction logic to the `<PolymiddlewareProvider>` and caching every type of extracted polymiddleware in `PolymiddlewareContext`, instead of having individual `<ActivityPolymiddlewareProvider>` and `<ErrorBoxPolymiddlewareProvider>`, right?

Then, in `useBuildRenderActivityCallback()`, we will call `useContext(PolymiddlewareContext).polymiddlewareMap.get('activity')` to get the array of extracted activity polymiddleware.

In this case, we are centralizing all middleware changes into a single `PolymiddlewareContext`. It means, every time someone added a new `ErrorBoxPolymiddleware`, all `<ActivityProxy>` will get triggered because they depends on `useBuildRenderActiivtyCallback`, and ultimately depends on `useContext(PolymiddlewareContext)`.

So any middleware changes will cause all proxy to re-evaluate.

OEvgeny: I see, yea. Doesn't sound good when updating middleware. I guess it'll still won't cause any content re-renders as we'll memo the enhancer as well, so it is just every Proxy re-render.

__tests__/html/hooks.useCreateActivityRenderer.js
```diff
@@ -1,5 +0,0 @@
-/** @jest-environment ./packages/test/harness/src/host/jest/WebDriverEnvironment.js */
-
```
OEvgeny, line 2: TIL: you can just remove tests :rofl: (joking)

__tests__/html2/middleware/activity/invalidMiddleware.createdOutsideOfFactory.html
```diff
@@ -0,0 +1,96 @@
...
+            middleware: { activityComponent },
+            ReactWebChat
+          }
+        } = window;
+
+        const { directLine, store } = createDirectLineEmulator();
+
+        spyOn(console, 'warn').mockImplementation(() => {});
+
+        const Dummy = memo(function Dummy() {
```
OEvgeny, line 48: Even Dummy is memoized. Appreciate the dedication

packages/api-middleware/src/errorBoxPolymiddleware.tsx
OEvgeny: I meant something like this, but now I got it, it needs to be a hook in order to get the request to be properly memoized.

```ts
templatePolymiddleware<{ readonly activity: WebChatActivity }, { readonly children?: never }>('activity', (props) => useMemo(() => ({ activity: props.activity }), [props.activity]));

```

OEvgeny (approved):

__tests__/html2/middleware/activity/invalidMiddleware.createdOutsideOfFactory.html
compulim: Ha, I learnt that... maybe I should or should not `memo()` in test to mimic real world scenario... I am not sure yet, so I added it first. 🤣

packages/api-middleware/src/errorBoxPolymiddleware.tsx
compulim: Yea... we are renaming/destructuring `request` for friendliness. I was thinking to write a function but it doesn't help too much but simply abstracting obvious logic, feels overengineering if I go that route.

compulim:
Thank @OEvgeny for a lot of brainstorming sessions for the polymiddleware. Despite you didn't write any code, your ideas and visions goes into this PR. ❤️

PR #5571, opened 2025-09-08
Centralize build info

compulim:
### Added

- Unified build info and injection of `<meta>` tag, in PR [#5571](https://github.com/microsoft/BotFramework-WebChat/pull/5571), by [@compulim](https://github.com/compulim)

## Description

Today, build info is aggregated and injected by the bundle package at once.

Tomorrow, build info will be exported by individual package on import: `api`, `bundle`, `core`, `component`, `debug-theme`, `fluent-theme`.

This pull request is part of the clean up of bundle package, which is part of the C2 modernization workstream.

### Foundation change for C2 modernization

Today, the bundle package includes entry points and code (e.g. Adaptive Cards.)

Tomorrow, the bundle package will only includes entry point. The code will be split into another package. This enables multiple entry points:

- Slim CJS/ESM for NPM/esbuild
   - External dependencies are not bundled
- IIFE for UMD in browser (`<script>`)
   - External dependencies are bundled
   - React is polyfilled from `globalThis.React` and fallback to bundled React
- 🆕 Fat CJS/ESM in browser (`<script type="module">`)
   - External dependencies are bundled
 
| Type | Format | Delivered via | Bundling deps | Override React |
| - | - | - | - | - |
| Slim import | CJS/ESM | NPM | No | No |
| UMD | IIFE | <nobr>CDN: `<script>`</nobr> | Yes | `globalThis.React` fallback to bundled |
| 🆕 Fat import | ESM | <nobr>CDN: `<script type="module">`</nobr> | Yes | `globalThis.React` fallback to bundled |

## Design

Every package will have their own build info.

## Specific Changes

- Makes `tsup` output atomically
   - Outputs to `dist.tmp` and copy to `dist` on success
   - Moves `--onSuccess` into `tsup.config.ts`
- Centralize creating build info and injecting `<meta>` tag into `@msinternal/botframework-webchat-base` package
   - Use `define` instead of `env` for build info
      - `globalThis.WEB_CHAT_BUILD_INFO_BUILD_TOOL` is `tsup`
      - `globalThis.WEB_CHAT_BUILD_INFO_MODULE_FORMAT` is `commonjs`, `esmodules` or `global`
      - `globalThis.WEB_CHAT_BUILD_INFO_VERSION`
   - Use `change-case/kebab-case` for injecting into `<meta>`
      - `change-case` needs to be `devDependencies` as it is ESM only
- Added temporary `tsup` config (`tsup.config.bundled_*.mjs`) to `.gitignore`
- Cleaned up `.gitignore`
- Updated `lint-staged` config
- Added missing `precommit:eslint:*` and `precommit:typecheck:*`
- Removed overlapping tests related to `<meta>` and boot
   - Removed `<meta name="botframework-webchat:ui:version">`
   - Removed `<meta name="botframework-webchat:bundle:variant">`
   - Renamed to `<meta name="botframework-webchat">` from `<meta name="botframework-webchat:bundle">`

## Review Checklist

-  [x] ~Tests reviewed (coverage, legitimacy)~


tsup.base.config.ts
```diff
@@ -114,9 +114,15 @@ const baseConfig: Options & {
...
+
+  // Writing output to /dist.tmp/ and copy everything back to /dist/ for better atomicity.
+
   // "onSuccess" runs before DTS, we need to wait until *.d.ts are emitted.
-  onSuccess: `while [ -z "$(find ./dist.tmp -name '*.d.ts' -print -quit)" ]; do sleep 0.2; done; mkdir -p ./dist/ && cp ./dist.tmp/* ./dist/`,
+
+  // All instances of tsup will try to copy at the same time and could fail with "cp: cannot create regular file './dist/...': File exists".
+  // We will retry.
+
+  onSuccess: `while [ -z "$(find ./dist.tmp -name '*.d.ts' -print -quit)" ]; do sleep 0.2; done; mkdir -p ./dist/; until cp ./dist.tmp/* ./dist/; do sleep 0.5; done`,
```
OEvgeny: @copilot any way to do it better?

compulim: Two parts:

- Wait until `*.d.ts`/`*.d.mts`
   - tsup run `onSuccess` in parallel with Rollup generating DTS
   - I don't know if there is a better solution, sounds like a bug to me...
      - Filed as https://github.com/egoist/tsup/issues/1363
- Retry copy
   - For every output, tsup config can write to different folders `dist.tmp/(cjs|esm|iife)` and copy-merge them back to `dist/`, so they don't collide
   - But then, every tsup config will have their own version of `onSuccess`...
      - `onSuccess` is `string` but not `string[]`... so it is a bit messy if we go this route
      - I was thinking to move to `Promise<void>` via [`tinyexec`](https://npmjs.com/package/tinyexec) which [`tsup` is using](https://github.com/egoist/tsup/blob/8b6907df5d944e5e4a639a297875a35818ba9fa5/src/index.ts#L355), a bit complex so I didn't go this route yet

packages/api/src/buildInfo.ts
```diff
@@ -0,0 +1,9 @@
+import { createBuildInfo } from '@msinternal/botframework-webchat-base/utils';
+
+const buildInfo = createBuildInfo('botframework-webchat:api');
+
+buildInfo.set('buildTool', globalThis.WEB_CHAT_BUILD_INFO_BUILD_TOOL);
```
compulim, line 5: Cannot "define" in `tsd`, we need to use `globalThis`.

packages/base/package.json
```diff
@@ -68,6 +68,7 @@
     "@msinternal/botframework-webchat-tsconfig": "^0.0.0-0",
     "@tsconfig/strictest": "^2.0.5",
     "@types/node": "^24.1.0",
+    "change-case": "^5.4.4",
```
compulim: `change-case` is ESM only, we need to bundle using `devDependencies`.

packages/base/tsup.config.ts
```diff
@@ -1,24 +1,34 @@
...
+// TODO: [P1] Compute this automatically.
+const DEPENDENT_PATHS = [
+  'api/src/index.ts',
+  'api-middleware/src/index.ts',
+  'bundle/src/boot/exports/full.ts',
+  'component/src/index.ts',
+  'core/src/index.ts',
+  'debug-theme/src/index.ts',
+  'fluent-theme/src/index.ts'
+];
```
OEvgeny, line 14: Maybe worth adding this to the PR?

tsup.base.config.ts
OEvgeny: Base config is already doing a lot. I think if we do `createConfig` as a function instead, we could add more logic in there...

packages/base/src/utils/setMetaTag.ts
```diff
@@ -0,0 +1,35 @@
...
+      const meta = document.querySelector(`html meta[name="${encodeURI(name)}"]`) || document.createElement('meta');
+
+      meta.setAttribute('name', name);
+      meta.setAttribute(
+        'content',
+        typeof contentMap === 'string'
+          ? contentMap
+          : Array.from(contentMap.entries())
+              .map(([key, value]) => {
+                const name = kebabCase(encodeURIComponent(key));
```
OEvgeny, line 25: Maybe I'm too old-fashioned, but for me this doesn't look nor secure to do via an external lib, nor I see it is pragmatic given complexity/size.

Given we want to only convert ascii and given we have limited set of keys, it should be safe to consider:
```js
  s.replace(/([ _A-Z])/g, '-$1').replace(/[A-Z]/g, c => c.toLowerCase()).replace(/([\-]+)/g, '-');
```

compulim: Removed `change-case` and moved to custom code instead, small footprint.

packages/base/tsup.config.ts
compulim: Not exactly trivial but can vibecode it. Maybe wait for next PR... people says it is a bug in tsup not picking up files.

- For every packages inside `/package.json/workspaces`, filter a list of path that their `package.json` has `dependencies`/`devDependencies` pointing to this package
- Get the filename of `package.json/exports/./default`

tsup.base.config.ts
compulim: Done, have a new `applyConfig`.

I was thinking/testing 2 signatures:

- `applyConfig(config => ({ ...config, env: { ...config.env, MY_STUFF: '...' } }))`
   - Aligned a bit more to tsup `defineConfig` approach, where it is `defineConfig(config => ({ ... })): void`
   - Caller has more controls, but also more responsibility to merge things correctly
- `mergeConfig({ env: { MY_STUFF: '...' } })`
   - Caller handle less merging, but also has less ability to override the config

I like `applyConfig` more.

For `onSuccess` concatenation, I am doing a "reconfigure after apply". It's rare but I kinda like this (now).

packages/base/src/utils/setMetaTag.ts
```diff
@@ -0,0 +1,39 @@
+/// <reference lib="dom" />
+
+function kebabCase(value: string): string {
+  return value
+    .replace(/([ _A-Z])/gu, '-$1')
+    .replace(/[A-Z]/gu, c => c.toLowerCase())
+    .replace(/([-]+)/gu, '-');
```
OEvgeny: Sorry, that didn't mean to be direct suggestion 😄 

If you want one which actually works well, consider:

```suggestion
    .replace(/([a-z0-9]|(?=[A-Z]))([A-Z])/gu, '$1-$2')
    .toLowerCase()
    .replace(/[^a-z0-9]+/gu, '-')
    .replace(/^-+|-+$/gu, '');
```

OEvgeny (dismissed):

PR #5572, opened 2025-09-09
Localized file check-in by OneLocBuild Task: Build definition ID 12805: Build ID 2235131

olprod:
This is the pull request automatically created by the OneLocBuild task in the build process to check-in localized files generated based upon translation source files (.lcl files) handed-back from the downstream localization pipeline. If there are issues in translations, visit https://aka.ms/icxLocBug and log bugs for fixes. The OneLocBuild wiki is https://aka.ms/onelocbuild and the localization process in general is documented at https://aka.ms/AllAboutLoc.


compulim:
Filed a bug to https://ceapex.visualstudio.com/CEINTL/_workitems/edit/1120244.

Track that bug to see the translation progress.

PR #5584, opened 2025-09-17
[WIP] fluent-theme should depends on bundle only

compulim:
> Fixes #

## -

-

## Review Checklist

### DIFF
--- a/packages/bundle/src/boot/exports/internal.ts
+++ b/packages/bundle/src/boot/exports/internal.ts
@@ -1,0 +1,11 @@
+// We should review exports in this file to make sure 1P = 3P.
+export { type ActivityMiddleware, type TypingIndicatorMiddleware } from 'botframework-webchat-api';
+export {
+  createActivityBorderMiddleware,
+  createActivityGroupingMiddleware,
+  DecoratorComposer,
+  type DecoratorMiddleware
+} from 'botframework-webchat-api/decorator';
+export { WebChatDecorator } from 'botframework-webchat-component/decorator';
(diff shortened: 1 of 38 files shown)


packages/bundle/src/boot/exports/internal.ts
```diff
@@ -0,0 +1,11 @@
+// We should review exports in this file to make sure 1P = 3P.
```
OEvgeny, line 1: We're lucky there is no any context involved as otherwise we'd have multiple context instances between `bundle` and `bundle/internal`

packages/fluent-theme/tsup.config.ts
```diff
@@ -83,7 +83,7 @@ export default defineConfig([
     },
     entry: { 'botframework-webchat-fluent-theme.development': './src/bundle.ts' },
     esbuildPlugins: [
-      ...(config.esbuildPlugins || []),
+      // ...(config.esbuildPlugins || []),
```
OEvgeny: Should we remove or uncomment?

packages/bundle/src/boot/exports/internal.ts
compulim: Oh yea... it's lucky that we don't have to export `<*Context>` from `api`, and re-exporting it here. That's a good thinking and future catch.

OEvgeny (approved):

packages/fluent-theme/tsup.config.ts
compulim: 6 hours later... uncommented. What a ride. 🤣

https://github.com/unjs/unplugin/issues/546

PR #5592, opened 2025-09-24
Add support for <script type="module"> (a.k.a. fat module)

compulim:
### Added

- (Experimental) Added support for importing via `<script type="module">`, by [@compulim](https://github.com/compulim) in PR [#5592](https://github.com/microsoft/BotFramework-WebChat/pull/5592)

## Description

In this pull request, we are adding support of `<script type="module">`, in additional to existing IIFE and ESM for bundlers.

The new "ESM for `<script>`" loading mechanism enables web developers to load and customize Web Chat in pure HTML/JS, without any bundlers. The ESM package is also code-split into hundreds of chunk and can be downloaded efficiently over HTTP/2+.

In future, "ESM for `<script>`" will enable lazy-loading and many modern scenarios.

### ESM for `<script>`

> Notes: your import map must include `react` and `react-dom`. Both are peer dependencies of Web Chat.

Once this feature leaves the experimental phase, it will become the recommended mechanism over IIFE/UMD. We will migrate our samples and documentations to use ESM for `<script>`.

```html
<script type="importmap">
  {
    "imports": {
      "botframework-webchat": "https://.../static/botframework-webchat.js",
      "react": "https://esm.sh/react@18",
      "react-dom": "https://esm.sh/react-dom@18"
    }
  }
</script>
<script type="module">
  import { renderWebChat } from 'botframework-webchat';

  renderWebChat({
    ...
  }, document.getElementById('webchat'));
</script>
```

#### Sovereign cloud and air-gapped environment

In the previous code snippet, we rely on [esm.sh](https://esm.sh/) to load React into browser. As [esm.sh](https://esm.sh/) is an Internet-facing service, this will not work for sovereign cloud and air-gapped environment.

For conveniences, Web Chat is re-exporting React 16.8.6 in ESM format. This will help web developers hosting Web Chat in limited environment.

For most scenarios, web devs should use production-grade CDN/ESM services for serving React, such as [esm.sh](https://esm.sh/) and [esm.run](https://esm.run/).

```html
<script type="importmap">
  {
    "imports": {
      "botframework-webchat": "https://.../static/botframework-webchat.js",
      "react": "https://.../static/react.js",
      "react-dom": "https://.../static/react-dom.js"
    }
  }
</script>
```

### IIFE/UMD

> Notes: this is unchanged.

Loading via IIFE/UMD is no longer recommended. ESM for `<script>` is preferred.

```html
<script src="https://.../dist/botframework-webchat.js"></script>
<script>
  window.WebChat.renderWebChat({
    ...
  }, document.getElementById('webchat'));
</script>
```

In IIFE, Web Chat will use React from `window.React` if available, otherwise, it will use a bundled version of React 16.8.6. Internally, we call this mechanism "isomorphic React".

### ESM for bundler

> Notes: this is unchanged.

```sh
npm install botframework-webchat
```

```ts
import { renderWebChat } from 'botframework-webchat';

renderWebChat({
  ...
}, document.getElementById('webchat'));
```

Then, build the project with a bundler.

### CommonJS in ESM

CommonJS can be loaded by ESM code. However, due to technical limitations, CommonJS code cannot be re-exported as ESM.

We did a special esbuild plugin to convert CommonJS into ESM for re-exporting. However, this is fragile and very limited to a few packages, such as `html-react-parser` and `react-dom`, which load `react` via `require()`, they will be converted to `import` statement.

For best compatibility, CommonJS packages are bundled in ESM packages. Thus, CommonJS packages are duplicated on a per ESM package basis. There are [some tricks done by esm.sh](https://github.com/esm-dev/esm.sh/blob/5cc3937618bdb4e65ade4df4b64de532cd456b51/server/build.go#L1262) to re-export CommonJS as ESM forcibly. We may follow their implementation in future releases.

### Re-packing

Some CommonJS packages are re-packed, so they can be imported properly by ESM packages.

To reduce footprint, `object-assign` is re-packed by exporting `Object.assign()`.

### Supporting CDN and static CMS

We expect web developers to copy-deploy the `/static/` folder to their CDN and static content management system (CMS).

### Browser has stricter rules for ESM than bundler

Bundler allows this statement: `export default fn()` and `fn()` will be expanded on execute, enabling `import { something }` scenarios.

However, browser do not allow the same statement because `fn()` is not declarative, thus, `import { something }` would fail.

Thus, we need to re-pack some packages to export them declaratively.

### Import maps

If we would like to enable web devs to override some of our dependencies via import map, we will need to publish an import map of about 50-100 packages. This is proven difficult to maintain. Thus, currently, we are limited our import map to 3 entries: `botframework-webchat`, `react`, and `react-dom`.

### Loading React

React must be singular in the system. Loading multiple versions of React will fail the rule of hooks.

Thus, we designed this loading mechanism which allow web devs to select their version of React while maintaining singularity:

- ESM for `<script>`: load React via import map
- ESM for bundler: load React via `package.json/peerDependencies`
- IIFE/UMD: load React via `window.React` if available, fallback to internal bundled React

### Why multiple chunks than a single bundle?

Currently, there are about 465 chunks of 6.6 MiB minified.

We want to load packages once-and-only-once and shared when possible. With each package in its own (set of) chunks, JS code loading the same package will refer to same set of memory. This will reduce duplicated packages in memory and reduce footprint.

However, due to technical limitations, this only applies to ESM. Packages exported in CommonJS will be bundled inside ESM packages, thus, they will be duplicated on a per ESM package basis.

The total size is larger than IIFE/UMD bundle, probably because we duplicated CommonJS packages. We will explore options to reduce the differences.

## Specific Changes

-

## Review Checklist


package.json
```diff
@@ -32,6 +32,15 @@
     "./packages/isomorphic-react",
     "./packages/isomorphic-react-dom",
     "./packages/component",
+    "./packages/repack/adaptivecards",
```
OEvgeny, line 35: Consider single `repack` package with multiple entry points for reduced boilerplate code.

packages/bundle/esbuild.static.mjs
```diff
@@ -0,0 +1,266 @@
...
+/* eslint-disable no-magic-numbers */
+/* eslint-disable no-restricted-globals */
+
+import { build, context } from 'esbuild';
+import { resolve as importMetaResolve } from 'import-meta-resolve';
+import { dirname, resolve } from 'path';
+import { readPackageUp } from 'read-pkg-up';
+import { fileURLToPath, pathToFileURL } from 'url';
+
+// const isomorphicReactPlugin = {
```
OEvgeny, line 14: Can we keep it but not commented, just eslint-ignore the unused variable.

This way we can keep the code and we have to adjust it rather have it dead commented.

OEvgeny: Removing it is also an option, but I bet it may make sense to have it in some form given react is still UMD/CJS in 2025.

packages/bundle/esbuild.static.mjs
```diff
@@ -0,0 +1,266 @@
...
+      name: 'static-builder',
+      setup(build) {
+        // eslint-disable-next-line require-unicode-regexp
+        build.onResolve({ filter: /^[^.]/ }, async args => {
+          // "external" field only works if the plug-in give up (return undefined.)
+          if (args.path === 'react' || args.path === 'react-dom') {
+            return undefined;
+          }
+
+          // Only ESM can be externalized, CJS cannot be externalized because require() is not guaranteed to be at top-level.
```
OEvgeny, line 181: I assume this is esbuild limitation. We can try work around it by using different transform from CJS to ESM (there are a lot of options available now).

Maybe we can bail if `require` is found not in top-level or at least fall back to CJS (which enforces duplicating code).

OEvgeny (approved):

packages/bundle/esbuild.static.mjs
compulim: Think about `a && require('b')` where `b` has side effects. We can't really convert this into `import`.

But it's relatively rare.

PR #5610, opened 2025-10-07
feat: bundle internal packages types

OEvgeny:
> Fixes #

## Changelog Entry

- Fix published package types containing internal package references

## Description

This PR adds a layer of bundling for our types as well as an extra validation layer to ensure we do not leak `@msinternal/*` packages type references to the outside.

## Design

As to the date `tsup` does not resolve external package types which is stated in their [docs](https://tsup.egoist.dev/#generate-declaration-file):
> Note that --dts does not resolve external (aka in node_modules) types used in the .d.ts file, if that's somehow a requirement, try the experimental --dts-resolve flag instead.

I've tried the above mentioned experimental `--dts-resolve` flag without any luck. So the only option seems to shop around for the solutions available.
 
The `dtsroll` package is designed to modify generated definition files in place according to the `package.json` definitions.

I shuffled a bit dependencies in the `package.json` files we have, but overall dependencies changes are not needed except for the `bundle` package.

I also removed Webpack 4 legacy entries for internal packages as these packages should be bundled and such entries trip `dtsroll` for some reason.

## Specific Changes

- updated public `package.json` files to include `dtsroll` transform step
- updated public `package.json` files to include validation step

## -

-

## Review Checklist

### DIFF
--- a/packages/api-middleware/package.json
+++ b/packages/api-middleware/package.json
@@ -85,11 +85,11 @@
     "typescript": "^5.7.3"
   },
   "peerDependencies": {
+    "react-chain-of-responsibility": "0.4.0-main.c2f17da",
     "react": ">= 16.8.6"
   },
   "dependencies": {
     "handler-chain": "^0.1.0",
-    "react-chain-of-responsibility": "0.4.0-main.c2f17da",
     "react-wrap-with": "0.1.0",
     "valibot": "1.1.0"
   }
--- a/packages/bundle/package.json
+++ b/packages/bundle/package.json
@@ -97,9 +97,12 @@
     }
   },
   "scripts": {
-    "build": "npm run build:static && npm run build:tsup",
+    "build": "npm run build:static && npm run build:tsup && npm run build:dtsroll && npm run build:validate",
     "build:static": "node ./esbuild.static.mjs",
     "build:tsup": "tsup",
+    "build:dtsroll": "dtsroll ./dist/*.d.*",
+    "build:validate": "npm run build:validate:dts",
+    "build:validate:dts": "if grep -q -P '@msinternal\\/' dist/*.d.* 2>/dev/null; then echo \"Error: dist/*.d.* is not compiled by dtsroll\" >&2; exit 1; fi",
     "bump": "npm run bump:prod && npm run bump:dev && (npm audit fix || exit 0)",
     "bump:dev": "PACKAGES_TO_BUMP=$(cat package.json | jq -r '(.pinDependencies // {}) as $P | (.localDependencies // {} | keys) as $L | (.devDependencies // {}) | to_entries | map(select(.key as $K | $L | contains([$K]) | not)) | map(.key + \"@\" + ($P[.key] // [\"latest\"])[0]) | join(\" \")') && [ ! -z \"$PACKAGES_TO_BUMP\" ] && npm install $PACKAGES_TO_BUMP || true",
     "bump:prod": "PACKAGES_TO_BUMP=$(cat package.json | jq -r '(.pinDependencies // {}) as $P | (.localDependencies // {} | keys) as $L | (.dependencies // {}) | to_entries | map(select(.key as $K | $L | contains([$K]) | not)) | map(.key + \"@\" + ($P[.key] // [\"latest\"])[0]) | join(\" \")') && [ ! -z \"$PACKAGES_TO_BUMP\" ] && npm install --save-exact $PACKAGES_TO_BUMP || true",
@@ -236,6 +239,7 @@
     "typescript": "~5.8.3"
   },
   "peerDependencies": {
+    "react-chain-of-responsibility": "0.4.0-main.c2f17da",
     "react": ">= 16.8.6",
     "react-dom": ">= 16.8.6"
   }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -56,7 +56,11 @@
     }
   },
   "scripts": {
-    "build": "tsup",
+    "build": "npm run build:tsup && npm run build:dtsroll && npm run build:validate",
+    "build:tsup": "tsup",
+    "build:dtsroll": "dtsroll ./dist/*.d.*",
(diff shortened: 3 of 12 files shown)


packages/bundle/package.json
```diff
@@ -90,16 +90,20 @@
       "allowSyntheticDefaultImports": true,
       "downlevelIteration": true,
       "jsx": "react",
+      "module": "NodeNext",
```
OEvgeny, line 93: We can try relax to `Node16` if you're comfortable with such.

tldr: `dtsroll` bails on `entry.js` files as it is using `@rollup/plugin-node-resolve` which seem to favor the entries instead of the `exports` map. But we should promote modern resolution for our internal modules anyway, and the packages we need to publish are never touched during processing as they should remain external.

compulim: I will keep as strict as possible, `NodeNext` is stricter than `Bundler`. 🤣

packages/bundle/package.json
```diff
@@ -236,6 +239,8 @@
     "typescript": "~5.8.3"
   },
   "peerDependencies": {
+    "@types/react": "^16.14.65",
```
compulim, line 242: I guess `peerDependencies` means it will be installed as production dependencies, right?

packages/component/src/Utils/createIconComponent.tsx
```diff
@@ -12,46 +24,86 @@ type ModifierMap<T> = {
...
-  const TModifiers extends SafeModifierList<TCSSModfuleClasses>
->(styles: TCSSModfuleClasses, modifiers: TModifiers[]) {
-  type CSSModuleModifiers = ModifierMap<TCSSModfuleClasses>;
+type ModifierSchemaEntry = OptionalSchema<PicklistSchema<readonly [string, ...string[]], undefined>, undefined>;
 
+export type IconModifierPropsSchema<TModifiers extends string> = BaseSchema<
+  Partial<Record<TModifiers, string>>,
+  Readonly<Partial<Record<TModifiers, string>>>,
+  BaseIssue<unknown>
+> & {
```
compulim, line 37: (You can keep it like this for now.)

TBH, I still don't know what's the best coding pattern to use valibot for React props. Is it `InferInput` or `InferOutput`?

- `InferInput`: we allow partial props (optional as `undefined`) and we rectify it (fill out the `undefined`), however, we lose the `readonly()`
- `InferOutput`: we get the `readonly()`, which is not trivial to set, but we lose the optional, everything must be defined

I feel `InferInput` is better but it's not trivial to add `readonly()` back, consider adding the readonly to both places: `{ readonly values: readonly number[] }`.

packages/core/package.json
```diff
@@ -49,14 +49,19 @@
...
       "resolveJsonModule": true,
       "skipLibCheck": true,
       "strict": false
     }
   },
   "scripts": {
-    "build": "tsup",
+    "build": "npm run build:tsup && npm run build:dtsroll && npm run build:validate",
+    "build:tsup": "tsup",
+    "build:dtsroll": "dtsroll ./dist/*.d.*",
```
compulim: Sort.

```suggestion
    "build:dtsroll": "dtsroll ./dist/*.d.*",
    "build:tsup": "tsup",
```

packages/api-middleware/package.json
compulim, line 36: Maybe we can remove CommonJS entrypoints from private packages, later.

packages/fluent-theme/src/components/icon/FluentIcon.tsx
```diff
@@ -37,13 +37,7 @@ const { component: FluentIcon, modifierPropsSchema } = createIconComponent(
...
 FluentIcon.displayName = 'FluentIcon';
 
-const fluentIconPropsSchema = pipe(
-  object({
-    ...baseFluentIconPropsSchema.entries,
-    ...modifierPropsSchema.entries
-  }),
-  readonly()
-);
+const fluentIconPropsSchema = pipe(union([baseFluentIconPropsSchema, modifierPropsSchema]), readonly());
```
compulim, line 40: Is it `union` (`|`) or `intersect` (`&`)? I always confuse with them.

packages/fluent-theme/package.json
```diff
@@ -40,10 +40,13 @@
     "static/**/*"
   ],
   "scripts": {
-    "build": "npm run build:static && npm run build:tsup && npm run build:validate",
+    "build": "npm run build:static && npm run build:tsup && npm run build:dtsroll && npm run build:validate",
     "build:static": "node ./esbuild.static.mjs",
     "build:tsup": "tsup",
-    "build:validate": "grep -q -P '\\.webchat-fluent \\.w([a-zA-Z-]){6}_' dist/*.css 2>/dev/null || { echo \"Error: dist/*.css is not compiled by Lightning CSS\" >&2; exit 1; }",
+    "build:dtsroll": "dtsroll ./dist/*.d.*",
```
compulim:
```suggestion
    "build:dtsroll": "dtsroll ./dist/*.d.*",
    "build:static": "node ./esbuild.static.mjs",
    "build:tsup": "tsup",
```

packages/bundle/package.json
OEvgeny: Yes, it gets installed, but it also is easier to swap. In this case you'll get a warning on versions mismatch

packages/component/src/Utils/createIconComponent.tsx
OEvgeny: When I was going through rewriting this I was just praying it to work across:
- tsd
- precommit hook
- vscode
- tsup

The code triggered errors on everything at least once.

packages/fluent-theme/src/components/icon/FluentIcon.tsx
OEvgeny: I can rollback to previous. I think it should be `union`. Let's keep it here, so if we need to change the underlying types, it'd serve as an additional sanity check :rofl:

PR #5611, opened 2025-10-08
Clean up <ThemeProvider> and various CSS related stuff

compulim:
### Added

- Cleaned up `<ThemeProvider>` and various CSS related code, in PR [#5611](https://github.com/microsoft/BotFramework-WebChat/pull/5611), by [@compulim](https://github.com/compulim)

## Description

This is for paving the way to host a new experience working side-by-side with existing experience. We need CSS Modules to work with the new experience.

Also took some time to audit the usage of nonce and clean it up so it's easier to audit.

## Design

> Note: this section need to be updated.

### New `<StyleOptionsComposer>`

Slowly refactoring things out of `<Composer>`. `<Composer>` is still the root of everything (except `<ThemeProvider>` which could be parent of `<Composer>`.)

All rectification logic is being refactored. We should slowly revisit it by using valibot to fill optional fields and for rectifications.

The current rectification logic does not take cascading into account. We need some modifications.

### Hook vs component

`useInjectStyles` is replaced by `<InjectStyleElements>` for cleaner DOM related manipulation.

- Component likely lead to DOM change
- Hook unlikely lead to DOM change, despite the DOM is outside of React tree
- `<InjectStyleElements>` is a leaf node component without `children` props
- `<InjectStyleElements>` will handle duplications, including handling same instance of `<link rel="stylesheet">`/`<style>` with and without nonce
- Dedupe is best effort, if the element is being injected twice, the instance will be moved to a latter position to make sure it is cascaded properly

As a result, `useCSSCustomProperties()` is also becoming `<CustomPropertiesContainer>`.

### CSS injection order

Instead of saving all CSS injections into context, to save some logics, we are injecting CSS when the `<InjectStyleElements>` is mounted. Ordering is important.

```html
<InjectStyleElementsComposer styleElements={[stylesA1, stylesA2]}>
  <InjectStyleElementsComposer styleElements={[stylesB1, stylesA1]} />
</InjectStyleElementsComposer>
```

Will be injected in the following order to honor CSS cascading:

```html
<head>
  <style>... stylesA2 ...</style>
  <style>... stylesB1 ...</style>
  <style>... stylesA1 ...</style>
</head>
```

Note:

- Deeper component has higher priority, will be injected later or moved to a latter position
   - `stylesB1` is after `stylesA2` because it is deeper in the tree
   - `stylesA1` get moved after `stylesB1` because of dedupe while keeping "deeper element is injected later"

Also thought about using `createPortal()` to inject DOM, but it is not available in React 16.8.6.

### Fixed bugs

- When mounting the same *parented* `<style>` element twice, it should not mount it twice
- Some normalization/rectification logic for style options, especially those logic that handle deprecated options, need to be rewrite so they handle cascading properly

## Specific Changes

- Refactored `<StyleOptionsComposer>` out of `<API.Composer>`
- Renames
   - `<WebChatTheme>` -> `<ComponentStylesheet>`/`<DecoratorStylesheet>`
   - `createStyles` -> `createComponentStyleElements`/`createDecoratorStyleElements`
- `useInjectStyleElements(styleElements)` -> `<InjectStyleElements nonce={} styleElements={}>`
   - `useCSSCustomProperties()` -> `<CSSCustomPropertiesContainer>`
- Fixed `--watch 200`
   - `esbuild --watch --watch-delay 200`
   - `tsup.watch({ delay: 200 })`

## Review Checklist


packages/bundle/src/boot/actual/internal.ts
```diff
@@ -3,12 +3,12 @@ export { type ActivityMiddleware, type TypingIndicatorMiddleware } from 'botfram
 export {
```
compulim, line 3: I think everything in the `bundle/internal` should be for testing only. For example, `fluent-theme` should not rely on `bundle/internal`, 1P = 3P.

So, maybe we should rename this to `bundle/test` later (or sooner.)

packages/component/src/providers/Theme/private/Context.ts
```diff
@@ -9,33 +9,29 @@ import type {
...
-  attachmentMiddleware: readonly AttachmentMiddleware[];
-  avatarMiddleware: readonly AvatarMiddleware[];
-  cardActionMiddleware: readonly CardActionMiddleware[];
-  groupActivitiesMiddleware: readonly GroupActivitiesMiddleware[];
-  polymiddleware: readonly Polymiddleware[];
-  scrollToEndButtonMiddleware: readonly ScrollToEndButtonMiddleware[];
-  sendBoxMiddleware: readonly SendBoxMiddleware[];
-  sendBoxToolbarMiddleware: readonly SendBoxToolbarMiddleware[];
-  styleOptions: StyleOptions;
-  styles: readonly (HTMLStyleElement | HTMLLinkElement)[];
```
compulim, line 36: Removed `styles` from context.

Now will inject when the `<InjectStyleElementsComposer>` is mounted and will handle the CSS order correctly.

packages/component/src/providers/Theme/ThemeProvider.tsx
```diff
@@ -137,14 +236,21 @@ const ThemeProvider = ({
...
-      mergedStyleOptions,
-      mergedStyles,
       mergedToastMiddleware,
       mergedTypingIndicatorMiddleware
     ]
   );
 
-  return <Context.Provider value={context}>{children}</Context.Provider>;
+  return (
+    <StyleOptionsComposer styleOptions={styleOptions}>
```
compulim, line 216: Didn't think thoroughly yet: `<APIComposer>` will still render `<StyleOptionsComposer>`.

It sounds a bit weird that we can mount `<ChildOfAPIComposer>`/`<StyleOptionsComposer>` somewhere before `<APIComposer>`, but it *seems* it makes things cleaner...

Maybe when we clean up `<APIComposer>`, the idea will be cleaner.

__tests__/html2/themeProvider/override.html
```diff
@@ -0,0 +1,162 @@
...
+        const styleElements = document.head.querySelectorAll('[data-webchat-injected]');
+
+        // THEN: Should have 5 <style> elements.
+        expect(styleElements).toHaveLength(5);
+
+        // THEN: <style> should be strictly ordered as below.
+        expect(Array.from(styleElements).map(element => element.dataset['webchatInjected'])).toEqual([
+          '',
+          'fluent-theme',
+          'component/decorator',
```
compulim, line 142: What do you think about the identifier naming?

packages/fluent-theme/src/private/FluentThemeProvider.tsx
```diff
@@ -88,11 +103,11 @@ function FluentThemeProvider({ children, variant = 'fluent' }: FluentThemeProvid
             activityMiddleware={activityMiddleware}
             sendBoxMiddleware={sendBoxMiddleware}
             styleOptions={fluentStyleOptions}
-            styles={styles}
+            styles={styleElements}
             typingIndicatorMiddleware={typingIndicatorMiddleware}
           >
             <AssetComposer>
-              <WebChatDecorator>
+              <WebChatDecorator nonce={nonce}>
```
compulim, line 112: Should `<WebChatDecorator>` default in `<ComponentComposer>`?

__tests__/html/hooks.useInjectStyles.changeNonce.html
```diff
@@ -39,32 +39,31 @@
...
             );
 
           const styleElement = createStyle('main { background-color: Red!important; }');
 
-          await renderWithFunction(() => useInjectStyles([styleElement], '1'));
-          await host.snapshot();
+          await render({ elements: [styleElement], nonce: '1' });
+          await host.snapshot('local');
 
-          const allInjectedStyles1 = document.head.querySelectorAll('style:not([data-emotion]):not([data-webchat-injected])');
```
compulim, line 57: All Web Chat injected elements will have the `data-webchat-injected` attribute so we can do prepend later. See `<InjectStyleElementsComposer>`.

__tests__/html2/provider/InjectStylesElementsComposer/dupeElement.withParent.html
```diff
@@ -0,0 +1,93 @@
...
+        const styleElement = createStyle('main { background-color: Red!important; }');
+
+        document.head.append(styleElement);
+
+        await render({ elements: [styleElement, styleElement] });
+        await host.snapshot('local');
+
+        const allInjectedStyles1 = document.head.querySelectorAll('style[data-testid="style"]');
+
+        expect(allInjectedStyles1).toHaveLength(2);
```
compulim: This test make sure parented `<style>` are not injected twice.

packages/fluent-theme/src/private/FluentThemeProvider.tsx
OEvgeny: Good question. Why it can't pull nonce from say ThemeProvider?

OEvgeny: I think it was experimental and we decided to not include it yet.

compulim: This is `fluent-theme`, if it need to pull nonce, it will be `import { useNonce } from 'bf-wc-component/internal'`.

I try not to export the `useNonce()` function. 🤣

OEvgeny: Sure, I mean `WebChatDecorator` itself. It's kinda weird it needs nonce :rofl:

__tests__/html2/themeProvider/override.html
OEvgeny: Looks good!

packages/fluent-theme/src/private/FluentThemeProvider.tsx
compulim: Ahhhh, yeah, you are right.

compulim: It's like we can rename `<WebChatDecorator>` -> `<EnableDecorator>`.

compulim: Fixed.

packages/component/src/cssContent.ts
```diff
@@ -0,0 +1,5 @@
+/** CSS (as string) for components. */
+export const componentCSSContent = '@--COMPONENT-STYLES-CONTENT--@';
+
+/** CSS (as string) for decorators. */
+export const decoratorCSSContent = '@--DECORATOR-STYLES-CONTENT--@';
```
OEvgeny: Why we have different injection paths for component vs fluent-theme?

packages/component/src/Styles/InjectCSS.tsx
```diff
@@ -0,0 +1,37 @@
+import { reactNode, validateProps } from '@msinternal/botframework-webchat-react-valibot';
+import { makeCreateStyles } from '@msinternal/botframework-webchat-styles';
+import React, { memo, useMemo } from 'react';
+import { object, optional, pipe, readonly, string, undefinedable, type InferInput } from 'valibot';
+import { ThemeProvider } from '../boot/component';
+
+const injectCSSPropsSchema = pipe(
+  object({
+    children: optional(reactNode()),
+    cssContent: string(),
```
OEvgeny: Consider how it could impact importing styles by URL, if say we would like to allow providing `<link>` elements instead of styles and let the browser do its job.

Previous implementation was pretty easy to modify to make this working.

packages/fluent-theme/src/components/theme/Theme.tsx
```diff
@@ -1,12 +1,29 @@
-import React, { type ReactNode } from 'react';
+import { reactNode, validateProps } from '@msinternal/botframework-webchat-react-valibot';
```
OEvgeny, line 1: Assuming we're ok bundling it in multiple times?

packages/component/src/stylesheet/DecoratorCSS.tsx
```diff
@@ -0,0 +1,37 @@
...
+  readonly()
+);
+
+type DecoratorCSSProps = InferInput<typeof decoratorCSSPropsSchema>;
+
+function DecoratorCSS(props: DecoratorCSSProps) {
+  const { children, nonce } = validateProps(decoratorCSSPropsSchema, props);
+
+  const styleElements = useMemo(() => {
+    const styleElements = makeCreateStyles(decoratorCSSContent)('component/decorator');
```
OEvgeny: Have you tried instantiating two Web Chat instances? You should see a single style tag per instance, not two.

OEvgeny: I just checked with two webchat instances:
- `.webchat__css-custom-properties` is per instance, so it should appear twice
- `.webchat-fluent` is shared module, so it appears once
- `.webchat-decorator` - same as fluent-theme: appears once

When injecting into two different places, should appear twice (we don't check document boundaries, we trust the developer using our APIs).

When two instances have different nonce (or say one has no specified nonce), the we should duplicate styles in the single root:
image

packages/fluent-theme/src/components/theme/Theme.tsx
compulim: Yea, for now. We don't have a better solution yet.

The best one is to re-export it under `api` package. 🤣

packages/component/src/stylesheet/DecoratorCSS.tsx
compulim: Will write a test (and fix it.)

That's the reason we want to have a better caching system despite a more complex codebase.

packages/component/src/providers/InjectStyleElements/InjectStyleElementsComposer.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+        ~index && sharedInstances.splice(index, 1);
+
+        // When ref count goes to zero, unmount the element.
+        sharedInstances.includes(instance) || instance.mountingElement.remove();
+      }
+    };
+  }, [root, styleElements]);
+
+  return <Fragment>{children}</Fragment>;
```
OEvgeny: It doesn't provide any value by wrapping children:
- it doesn't pass any context down
- it doesn't set any props
- it doesn't transform children

So maybe it shouldn't wrap and shouldn't increase our virtual tree depth?

packages/component/src/cssContent.ts
OEvgeny: Does it mean that say component chunk will have both content defined? I'd like us to not leave any placeholders in the final chunks. This was the reason why the files were separate.

packages/component/src/providers/InjectStyleElements/InjectStyleElementsComposer.tsx
compulim: For TypeScript reason... 🤣

I can do `as any`.

compulim: Fixed.

packages/component/src/cssContent.ts
compulim: I see...

packages/component/src/providers/InjectStyleElements/InjectStyleElementsComposer.tsx
OEvgeny: Sorry, I was speaking about the InjectStyleElementsComposer, not about the fragment

packages/component/src/stylesheet/DecoratorCSS.tsx
compulim: Fixed.

packages/component/src/providers/InjectStyleElements/InjectStyleElementsComposer.tsx
compulim: Let me think (and try.)

packages/component/src/Styles/InjectCSS.tsx
compulim: `InjectCSS` is a side-effect of refactoring dupe code in `component.css` and `decorator.css`. The ultimate guy to inject is still `<InjectStyleElementsComposer>` and we can pass `<link>` there like before.

packages/component/src/cssContent.ts
compulim: They are separated now. Didn't notice about this design decision. 🤣

packages/component/src/providers/InjectStyleElements/InjectStyleElementsComposer.tsx
compulim: Reverted back to hook. 🤣

packages/component/src/decorator/stylesheet/DecoratorCSS.tsx
```diff
@@ -0,0 +1,33 @@
+// TODO: [P2] This component can be replaced by `bindProps(InjectCSS)({ cssContent, identifier })`.
```
OEvgeny: :heart: looks cleaner than the original now!

packages/component/src/stylesheet/ComponentCSS.tsx
```diff
@@ -0,0 +1,33 @@
+// TODO: [P2] This component can be replaced by `bindProps(InjectCSS)({ cssContent, identifier })`.
+import { reactNode, validateProps } from '@msinternal/botframework-webchat-react-valibot';
+import { makeCreateStyles } from '@msinternal/botframework-webchat-styles';
+import { memo, type FunctionComponent } from 'react';
+import { object, optional, pipe, readonly, string, undefinedable, type InferInput } from 'valibot';
+
+import useInjectStyleElements from '../Styles/useInjectStyleElements';
+import { componentCSSContent } from './componentCSSContent';
```
OEvgeny: Any benefit on exposing raw string from the module opposed to the creation function or the actual style element?

Previous design: the module exposes creation function which is instantiated per module using styles, so each module has own copy of styles.

Imagine we have a set of Custom Elements aka Web Components aside from the current styles entry. It makes sense to allow it to have own style element derived from the original CSS string, rather sharing the same style element.

Sharing CSS string might be okay, but I try to not introduce any moderately big simple value leaking out of a single module scope, so we can be sure: if it is copied, it is something related to browser DOM implementation, not JavaScript Module implementation interistics.

packages/component/src/decorator/stylesheet/DecoratorCSS.tsx
compulim: We found a new toy to build. 🙂

packages/component/src/stylesheet/ComponentCSS.tsx
compulim: To me, either is fine, it's just result of refactoring between `component.css` and `decorator.css`... the LCM between them is the string (or placeholder), so it turn out to be a string right now. Will revert back to the create function.

One thing I want to change is to reduce the complexity of "inject sometimes clone or not", which led to the "if has parent, it dupe" bug. So instead, it is now "always clone on inject" but keep track of the original element copy.

The `<link>`/`<script>` element owned by the module will never be modified (attached nor nonce-d). The module can send their style elements to other libraries if they want to, without worrying Web Chat modified it for our own good.

On the other side, if the `<style>` element is going to be injected in the `<head>`, having one copy per module means multiple modules will have multiple copies. And this could lead to being injected multiple times (can't dedupe using ref equals.)

If a subclassing custom element want to modify the inherited style, they should clone and modify it. But IMO, instead of modifying the style elements, I think the subclass should do it in CSS way: use a new class name and cascade from the original style.

__tests__/html2/hooks/useInjectStylesElements/dupeElement.html.snap-4.png
OEvgeny, line 1: I'm ok with just white, but maybe as improvement add some text into `::after { content: "..." }` so it is easier to spot what happens

packages/styles/src/react/private/InjectStyleElements.tsx
```diff
@@ -0,0 +1,149 @@
...
+  readonly mountingElement: InjectedStylesElement;
+  readonly nonce: string;
+  readonly root: Node;
+};
+
+const sharedInstances: InjectedStylesInstance[] = [];
+
+const warnNonce = warnOnce('The elements passing to <InjectStyleElements> should not have "nonce" attribute set');
+
+function InjectStyleElements(props: InjectStyleElementsProps) {
```
OEvgeny, line 64: Why packaging it? Why not placing inside styles package? I mean now styles feels kinda dummy as it is rather too generic name 🤣

compulim: How about rename `styles` to `component-styles` and put it inside `component-styles`.

Why... because of code duplications due to private bundling... so it's better to have `component-*` and it implies it is exported from `component` than everyone bundling it.

OEvgeny: We use it for fluent-theme though. I was thinking component-* is more like a dependency on the component stuff, but styles don't really depend on anything on that level 🙂

OEvgeny: Maybe we can have `styles/component` entry :rofl:

compulim: Moved to `/styles/` due to other bundling issues.

__tests__/html2/hooks/useInjectStylesElements/dupeElement.html.snap-4.png
compulim: Change to yellow background for default.

PR #5622, opened 2025-10-24
Graph backend phase 1

compulim:
### Added

- New JSON-LD graph backend, by [@compulim](https://github.com/compulim) in PR [#5622](https://github.com/microsoft/BotFramework-WebChat/pull/5622)

## Description

> This is phase 1 of the work. More phases are coming soon.

We are introducing a new JSON-LD graph backend. This new backend is paving the way towards replacing Redux as the single source of truth. And in future, next generation chat adapter will emit nodes and edges, instead of activities.

The graph backend use JSON-LD to ensure data are linked in a standardized manner. The schema will be largely backed by Schema.org along with some custom fields to support backward compatibility.

In this pull request:

- (Existing) Chat adapter insert activities into Redux store, as single source of truth
   - 🆕 `channelData['webchat:internal:id']: string` keeps the (semi-)permanent ID, similar to `<ActivityKeyer>` and is intended to replace `<ActivityKeyer>` in the future
   - 🆕 `channelData['webchat:internal:position']: number` keeps the positional index of the activity in the chat history
- 🆕 Activities from Redux store transform into JSON-LD node and inserted into graph
   - Activity (as JS object) is kept as-is under custom property named `urn:microsoft:webchat:direct-line-activity:raw-json`
- 🆕 On graph change, query all `Message` nodes and sort them into an array `orderedActivities: readonly WebChatActivity[]`
- 🆕 `useActivities()` will read from the new `orderedActivities`, instead of from Redux store state
   - Assertion is added to make sure result from `orderedActivities` is 100% same as Redux store state
   - The assertion will be removed once Redux store is no longer the single source of truth

What is (obviously) missing from this phase:

- Web Chat still use Direct Line Activity for everything, instead of JSON-LD nodes
- `activity.attachments[]` and `activity.entities[]` are not serialized to the graph yet
- No new hooks to access JSON-LD nodes
- Message schema is not formalized yet
- `node.sender` property is of type `Audience` and this is not coerrect, should be more targeted thing, such as `Person` or `Organization`
- No way to dump graph and incoming activities
- `activity:raw-json` should be JSON literals (all values must be serializable as JSON)
- Replace `<ActivityKeyer>` with the new permanent ID
- How to modify polymiddleware to support node while keeping backward compatibility

### JSON-LD Opinions: halfway between styles

JSON-LD is great for storing node and edges. However, to effectively access and retrieve data from the graph for our use, the data is shaped with opinions, around a few principles:

- Simplifying downstream logics
   - Must have `@id`: every node in the graph must be identifiable
   - Uniform getter/setter: every property value is an array, except `@context` and `@id`
   - Unique typing: node reference must be `{ "@id": string }` to reduce confusion with plain string
   - Support multiple types: every `@type` must be an array of string
   - Reduce confusion: property value with empty array and `null` is removed
      - `[]` and `null` are same as if the property is removed
   - Flattened: property values must be non-null literals, node reference, or JSON literals
      - Any array containing `null` is not supported and will throw unless it is JSON literal, as it is likely a bug in code
   - JSON literals will have boxing kept: `{ '@type': '@json', '@value': JSONValue }`
      - `@value` could be null, if unwrapped, will be confusing as we removed nulls
   - Do not handle full JSON-LD spec: `@context` is an opaque string and its schema is not honored
   - Node reference only has `@id` and it should not contain `@type`
   - Reduce confusion: node reference must not appear at the root of the flattened graph, they are semantically empty
- Auto-linking for Schema.org: `hasPart` and `isPartOf` are auto-inversed
- Keep its root: every node is compliant to JSON-LD, understood by standard parsers
- Debuggability: must have at least one `@type`

The end result is *still 100% legit JSON-LD*. However, they are not 100% styled as flattened nor compact. In other words, the final data shape is not absolutely styled, but halfway between each style to better fit our scenario.

### Refactoring insertion sort algorithm

1. By part grouping ID/position
2. By livestreaming session ID/position
3. By sequence ID coalesced with Unix epoch timestamp
   - If sequence ID is unset, use Unix epoch timestamp instead
   - When inserting, ignore existing activities where they have no sequence ID nor timestamp
4. Otherwise, append to the chat history

This refactoring allows us to have a new `channelData['webchat:internal:position']: number` property to keep the positional index of the activity. We will put this number into `Message.position` property.

## -

-

## Review Checklist

### DIFF
--- a/packages/core-graph/src/private/schemas/flattenNodeObject.ts
+++ b/packages/core-graph/src/private/schemas/flattenNodeObject.ts
@@ -129,8 +129,6 @@ function flattenNodeObject_(
     }
   }
 
-  // const id = parse(optional(IdentifierSchema), targetMap.get('@id')) ?? `_:${randomUUID()}`;
-
   targetMap.set('@id', id);
 
   const output: FlatNodeObject = parse(FlatNodeObjectSchema, Object.fromEntries(Array.from(targetMap)));


packages/api-graph/src/private/GraphProvider.tsx
```diff
@@ -0,0 +1,105 @@
+import type { DirectLineActivityNode } from '@msinternal/botframework-webchat-core-graph';
+import { reactNode, validateProps } from '@msinternal/botframework-webchat-react-valibot';
+import { createStore, WebChatActivity } from 'botframework-webchat-core';
+import { createGraphFromStore, isOfType, type GraphSubscriber, type Identifier } from 'botframework-webchat-core/graph';
```
OEvgeny, line 4: I like that it is a part of core

packages/api-graph/src/private/GraphProvider.tsx
```diff
@@ -0,0 +1,105 @@
...
+type GraphProviderProps = InferInput<typeof graphProviderPropsSchema>;
+
+function GraphProvider(props: GraphProviderProps) {
+  const { children, store } = validateProps(graphProviderPropsSchema, props);
+
+  const graph = useMemo(() => createGraphFromStore(store), [store]);
+
+  const [orderedActivityNodes, setOrderedActivityNodes] = useState<readonly DirectLineActivityNode[]>(EMPTY_ARRAY);
+
+  useEffect(() => {
```
OEvgeny, line 31: I imagine this would be [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) in the future which is shimmed for an older React via `useEffect`: https://github.com/facebook/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js#L109-L131

packages/base/src/valibot/freeze.ts
```diff
@@ -0,0 +1,12 @@
+import { transform } from 'valibot';
+
+/**
+ * Valibot pipe action to freeze the array/object. The value will not be cloned before freeze.
+ *
+ * The `readonly()` pipe action in Valibot does not freeze.
+ *
+ * @returns
+ */
+export default function freeze<T>() {
```
OEvgeny, line 10: Migrating to own graph means we can freeze nodes without an issue. We won't touch activities.

packages/core-graph/src/private/schemas/colorNode.ts
```diff
@@ -0,0 +1,161 @@
...
+            array(union([JSONLiteralSchema, LiteralSchema, NodeReferenceSchema])),
+            JSONLiteralSchema,
+            LiteralSchema,
+            NodeReferenceSchema,
+            null_()
+          ],
+          'Only JSON literal, literal, node reference or null can be parsed into slant node'
+        ),
+        value
+      );
```
OEvgeny: Let's extract this to the module scope, so we don't recreate the schema for each entry.

packages/core-graph/src/private/schemas/flattenNodeObject.ts
```diff
@@ -0,0 +1,185 @@
...
+import { IdentifierSchema, type Identifier } from './Identifier';
+import { JSONLiteralSchema, type JSONLiteral } from './JSONLiteral';
+import { LiteralSchema, type Literal } from './Literal';
+import { isNodeReference, NodeReferenceSchema, type NodeReference } from './NodeReference';
+import isPlainObject from './private/isPlainObject';
+
+function randomUUID(): string {
+  // crypto.randomUUID() requires HTTPS context.
+  // However, our legacy Jest tests are not running over HTTPS.
+  return v4();
```
OEvgeny, line 16: @copilot opinion on whether v4 uuid is optimal for the use case?

packages/core-graph/src/private/schemas/flattenNodeObject.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      targetMap.set(key, parsedValue);
+    } else if (typeof value !== 'undefined') {
+      parsedValue = flattenNodeObject_(value, graphMap, refMap);
+
+      targetMap.set(key, parsedValue);
+    }
+  }
+
+  // const id = parse(optional(IdentifierSchema), targetMap.get('@id')) ?? `_:${randomUUID()}`;
```
OEvgeny: nit: commented code

packages/core-graph/src/private/SlantGraph/private/autoInversion.ts
```diff
@@ -0,0 +1,154 @@
...
+          return Object.freeze({
+            ...parentNode,
+            hasPart: Object.freeze([...(parentNode.hasPart ?? []), Object.freeze({ '@id': id })])
+          });
+        });
+      }
+    }
+
+    return next(nextUpsertingNodeMap);
+  };
```
OEvgeny, line 152: @copilot analyze the code complexity and make a guess using O() notation.

OEvgeny (commented):
Minor nits and questions

packages/api-graph/src/private/GraphProvider.tsx
compulim: ❤️

packages/api-graph/src/private/GraphProvider.tsx
compulim: Yea, I tried `useSyncExternalStore.shim` in very early POC and it keeps re-rendering. Possibly bad memoization and I didn't bother to revisit it. I should have.

packages/base/src/valibot/freeze.ts
compulim: 💯

I am still skeptical how to use `readonly()` and `freeze()` properly. Not sure if it should be a `parse` action or just do it manually.

packages/core-graph/src/private/SlantGraph/private/autoInversion.ts
compulim: Ha, this is the ugliest part I want to revisit too. 🤣

OEvgeny: > Worst case: Dense graph where R ≈ N → O(N³)

So no dense graphs I guess :rofl:

packages/core-graph/src/private/schemas/colorNode.ts
compulim: Extracted and fixed a test.

packages/core-graph/src/private/schemas/flattenNodeObject.ts
compulim: Removed.

packages/core-graph/src/private/SlantGraph/private/autoInversion.ts
compulim: At least for now.

We probably should have a `[@type="Conversation"]` node as a root in the graph in next phase.

PR #5663, opened 2025-12-20
[Debug API] Add debugging capability

compulim:
### Added

- New debug API, by [@compulim](https://github.com/compulim) in PR [#5663](https://github.com/microsoft/BotFramework-WebChat/pull/5663), see [`DEBUGGING.md`](docs/DEBUGGING.md) for more
   - Debug into element: open <kbd>F12</kbd>, select the subject in Element pane, type `$0.webChat.debugger`
   - Breakpoint: open <kbd>F12</kbd>, select the subject in Element pane, type `$0.webChat.breakpoint.incomingActivity`

## Description

New debug API is added for debuggability.

### Debug API

We are introducing a new debug API interface. The new API attach to HTML element directly for debugging.

Currently, 2 types of elements has the debug API attached:

- Root container `$('.webchat')`
- Activity `$('article')`

### `debugger` property

> Note: this method will not work when Content Security Policy is enabled without `script-src unsafe-eval`. We do not recommend the policy to be enabled. In such scenario, use the `debug()` method instead.

The user would trigger the `debugger` in Web Chat context.

1. Press <kbd>F12</kbd> to open Chrome DevTools
2. In the Elements pane, identify Web Chat root element to debug and click on it
   - Note: the element should have class name of `.webchat`
3. In console, type `$0.webChat.debugger` and press <kbd>ENTER</kbd>
4. Chrome should break into debugger with debug context in the closure
   - The debug context contains data from Web Chat backend

### `debug()` on published breakpoint

> Note: breakpoints are for debugging and may change from time to time.

1. Press <kbd>F12</kbd> to open Chrome DevTools
2. In the Elements pane, identify Web Chat root element to debug and click on it
   - Note: the element should have class name of `.webchat`
3. In console, type `debug($0.webChat.breakpoint.incomingActivity)` and press <kbd>ENTER</kbd>
4. When a message arrive, Chrome should break into debugger with debug context in the closure

### Components and their breakpoints

| Component | CSS selector | Breakpoint | Description |
| - | - | - | - |
| Root container | `$('.webchat')` | `incomingActivity` | When an activity arrive |
| Activity row | `$('article')` | `render` | When the activity render |

To break when an activity arrive, type in DevTools.

```js
debug($('.webchat').webChat.breakpoint.incomingActivity)
```

### Components and their debug context

| Component | CSS selector | Debug context | Description |
| - | - | - | - |
| Root container | `$('.webchat')` | `activities` | All messages in the chat history |
| Activity row | `$('article')` | `activity` | The current message |

#### Break into debugger of the first instance of Web Chat on the page

```js
$('.webchat').webChat.debugger;
```

#### Break into debugger of the last activity on the page

```js
$$('article').at(-1).webChat.debugger;
```

## Specific Changes

- Added new debug API to root container and `<ActivityRow>`
- Added `breakpoint` and `debugger` to debug API
- Fixed some flaky tests

- [x] I have added tests and executed them locally
- [x] I have updated `CHANGELOG.md`
- [x] I have updated documentation

## Review Checklist

- [x] ~Accessibility reviewed (tab order, content readability, alt text, color contrast)~
- [x] ~Browser and platform compatibilities reviewed~
- [x] ~CSS styles reviewed (minimal rules, no `z-index`)~
- [x] Documents reviewed (docs, samples, live demo)
- [x] ~Internationalization reviewed (strings, unit formatting)~
- [x] ~`package.json` and `package-lock.json` reviewed~
- [x] Security reviewed (no data URIs, check for nonce leak)
- [x] Tests reviewed (coverage, legitimacy)

### DIFF
--- a/__tests__/html2/debugAPI/playground/lockdown.skip.html
+++ b/__tests__/html2/debugAPI/playground/lockdown.skip.html
@@ -48,7 +48,7 @@
         // "Cannot assign to read only property 'incomingActivity' of object '#<Object>'"
         const scenario1 = () => spyOn(rootElement.webChat.breakpoint, 'incomingActivity');
 
-        expect(scenario1).toThrow("Cannot assign to read only property 'incomingActivity' of object '#<Object>'");
+        expect(scenario1).toThrow("Cannot assign to read only property 'incomingActivity' of object '[object Object]'");
         expect(rootElement.webChat.breakpoint.incomingActivity).toBe(originalIncomingActivity);
 
         // Scenario 2.


docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
+# Debugging
+
+Web Chat supports debugging through the JavaScript debugger in the browser DevTools.
+
+## HTML elements
+
+The following table list HTML elements that support debugging.
```
OEvgeny:
```suggestion
The following table lists HTML elements that support debugging.
```

packages/api/src/providers/DebugAPI/private/DebugAPIContext.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createContext, useContext } from 'react';
+
+type DebugAPIContextType = {
+  readonly rootDebugAPIState: readonly [RootDebugAPI];
+};
+
+const DebugAPIContext = createContext<DebugAPIContextType>(
+  new Proxy({} as DebugAPIContextType, {
+    get() {
+      throw new Error('botframework-webchat: Cannot call this hook outside of <DebugAPIComposer>');
```
OEvgeny: Should we use global context here to allow breaking on any root if accessing from global Web Chat or a module export?

packages/core-debug-api/src/private/PrivateDebugAPI.ts
```diff
@@ -0,0 +1,85 @@
...
+import type { BaseContextGetters, BreakpointObject, ContextOfGetters, PrivateDebugAPIType } from '../types';
+import { SHOULD_LOCKDOWN } from './constants';
+import DebugAPI from './DebugAPI';
+
+// 🔒 This function must be left empty.
+// eslint-disable-next-line @typescript-eslint/no-empty-function, @typescript-eslint/no-unused-vars
+const BREAKPOINT_FUNCTION = <T>(__DEBUG_CONTEXT__: T) => {};
+
+class PrivateDebugAPI<
+  TBreakpointName extends string,
```
OEvgeny: May use valibot for validating the names so the type is more strict.

packages/core-debug-api/src/createPrivateDebugAPI.ts
```diff
@@ -0,0 +1,8 @@
+import PrivateDebugAPI from './private/PrivateDebugAPI';
+
+export default function createPrivateDebugAPI<
+  const TBreakpointName extends string,
+  const TContextGetters extends { readonly [key: string]: typeof key extends '@' ? never : () => unknown }
+>(breakpointNames: readonly TBreakpointName[], contextGetters: TContextGetters) {
+  return new PrivateDebugAPI<TBreakpointName, TContextGetters>(breakpointNames, contextGetters);
```
OEvgeny: I don't like passing breakpoint names around.

Let's define them as static array and import instead. Or define them in the same module creating PrivateDebugAPI instances

packages/core-debug-api/src/types.ts
```diff
@@ -0,0 +1,62 @@
...
+   *
+   * Note: if Content Security Policy is enabled, it may not break into debugger.
+   */
+  get debugger(): void;
+}
+
+/**
+ * 🔒 This object must not be passed outside for external consumption.
+ */
+interface PrivateDebugAPIType<TBreakpointName extends string, TContextGetters extends BaseContextGetters> {
```
OEvgeny: Consider instead of threatening, having these types closer to implementation and ensuring they don't leak to the outside world.

packages/core/src/internal/index.ts
```diff
@@ -4,3 +4,7 @@ export {
   setRawStateActionSchema,
   type SetRawStateAction
 } from './actions/setRawState';
+
+export { type RootDebugAPI } from '../types/RootDebugAPI';
+
+export { createPrivateDebugAPI, type DebugAPIType } from '@msinternal/botframework-webchat-core-debug-api';
```
OEvgeny: How do we protect from confusion? We'll have this code duplicated if included into different packages.

packages/core/src/reducers/activities/createGroupedActivitiesReducer.ts
```diff
@@ -189,6 +191,9 @@ function createGroupedActivitiesReducer(
           payload: { activity }
         } = action;
 
+        // We cannot call breakpoint inside Redux because DebugContext cannot call getState(), need setTimeout.
+        ponyfill.requestIdleCallback?.(() => rootPrivateDebugAPI.UNSAFE_callBreakpoint.incomingActivity({ activity }));
```
OEvgeny: Why not use a microtask?

packages/core/src/createStore.ts
```diff
@@ -108,15 +109,28 @@ export function withOptions(options: CreateStoreOptions, initialState?, ...middl
...
-    createReducer(ponyfill),
+
+  store = createReduxStore(
+    createReducer(ponyfill, rootPrivateDebugAPI),
     initialState || {},
     options.devTools ? composeWithDevTools(enhancer) : enhancer
   );
 
+  // TODO: [P1] When we redesign the store and chat adapter, we should have the Debug API stored somewhere else.
+  store['debugAPI'] = rootPrivateDebugAPI.toPublic();
```
OEvgeny: What happens if this is replaced with another malicious `debugAPI`?

OEvgeny: Consider using `WeakMap<store, debug>` in the module scope.

packages/core-debug-api/src/createPrivateDebugAPI.ts
OEvgeny: Consider `Object.create(null)` + defining handlers attached to the object in the same module closure as a more secure solution to blocklisting.

packages/core-debug-api/src/private/PrivateDebugAPI.ts
compulim: Using `Object.create(null)` so we don't care about the names. 🤣

packages/api/src/providers/DebugAPI/private/DebugAPIContext.ts
compulim: Global Web Chat and module export... there could be multiple Web Chat instances on the page and all root from the same module.

This is "public" object and potentially other JS code could consume it:

- `breakpoint`
  - They can't spy on any calls, because the object is frozen, they can't modify the function to call something else
  - They can call the breakpoint, but there is no point, the breakpoint will break into debugger without the debug context
- `debugger`
  - They can't spy on `debugger` getter, because the object is frozen
  - They can call `debugger`, still no point, the debug context aren't there

compulim: We don't have `useStore()`. When we get rid of the Redux store, maybe we can move this inside where the `HTMLElement` is instantiated. We need to think about those breakpoints when we do that: `incomingActivity` could be come `graphChange`.

packages/core-debug-api/src/createPrivateDebugAPI.ts
compulim: > Renamed `PrivateDebugAPI` to `RestrictedDebugAPI` to better indicate sensitivity.

Marked `RestrictedDebugAPI` as abstract.

Yap, will put breakpoint names under class `RestrictedRootDebugAPI` and `RestrictedActivityDebugAPI`, which is the realization of the abstract class.

packages/core/src/internal/index.ts
compulim: This code is imported by `core` and re-exported. It will not be duplicated (unlike `base` package.)

packages/core/src/reducers/activities/createGroupedActivitiesReducer.ts
compulim: Using `queueMicrotask` now.

packages/core/src/createStore.ts
compulim: ~Using `Object.freeze()` to prevent modifications.~

Using module-level registry `WeakMap` to prevent backward compatibility issues.

docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
...
+| Activity       | `render`           |                          | When the activity render |
+
+To break into the debugger when an activity arrives:
+
+1. Press <kbd>F12</kbd> to open Chrome DevTools
+2. In the Elements pane, click on the Web Chat root container to break into
+   - Or type `inspect($('.webchat'))` to select the first root container on the page
+3. Switch to Console pane, type `debug($0.webChat.breakpoint.incomingActivity)`
+
+When an activity arrive, it will automatically break into the debugger.
```
OEvgeny, line 48:
```suggestion
When an activity arrives, it will automatically break into the debugger.
```

docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
...
+| -------------- | ------------------ | ------------------------ | ------------------------ |
+| Root container | `incomingActivity` | `{ activity: Activity }` | When an activity arrive  |
+| Activity       | `render`           |                          | When the activity render |
+
+To break into the debugger when an activity arrives:
+
+1. Press <kbd>F12</kbd> to open Chrome DevTools
+2. In the Elements pane, click on the Web Chat root container to break into
+   - Or type `inspect($('.webchat'))` to select the first root container on the page
+3. Switch to Console pane, type `debug($0.webChat.breakpoint.incomingActivity)`
```
OEvgeny, line 46:
```suggestion
3. Switch to the Console pane, type `debug($0.webChat.breakpoint.incomingActivity)`
```

docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
...
+
+> Note: this method will not work when Content Security Policy is enabled without `script-src unsafe-eval`. We do not recommend the policy to be enabled. In such scenario, use the `debug()` method instead.
+
+To break into the debugger for a specific element:
+
+1. Press <kbd>F12</kbd> to open Chrome DevTools
+2. In the Elements pane, click on the element to break into
+3. Switch to Console pane, type `$0.webChat.debugger`
+
+Then, it should break into the debugger.
```
OEvgeny, line 28:
```suggestion
Then it should break into the debugger.
```
Copilot say no comma needed here

docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
...
+
+![Screenshot of DevTools debugger breaking into the root container component](media/DEBUGGING/devtools-debugger.png)
+
+### Breakpoints
+
+Use DevTools Console `debug()` to attach a debugger to specific breakpoint function.
+
+| Component      | Breakpoint         | Arguments                | Description              |
+| -------------- | ------------------ | ------------------------ | ------------------------ |
+| Root container | `incomingActivity` | `{ activity: Activity }` | When an activity arrive  |
```
OEvgeny, line 38:
```suggestion
| Root container | `incomingActivity` | `{ activity: Activity }` | When an activity arrives  |
```

docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Break into debugger of the last activity on the page
+
+```js
+$$('article').at(-1).webChat.debugger;
+```
+
+Note: `$$()` will select all elements and put them into an array, while `$()` only select the first occurrence.
+
+### Break into debugger when an activity arrive
```
OEvgeny, line 79:
```suggestion
### Break into debugger when an activity arrives
```

docs/DEBUGGING.md
```diff
@@ -0,0 +1,83 @@
...
+
+When breakpoints are used, the debugger may offer extra diagnostic information relevant to the triggering event, available as function arguments.
+
+### Debug context
+
+While in the debugger, the `__DEBUG_CONTEXT__` variable contains various data useful for diagnostic purpose.
+
+| Component      | Debug context                | Description                      |
+| -------------- | ---------------------------- | -------------------------------- |
+| Root container | `{ activities: Activity[] }` | All messages in the chat history |
```
OEvgeny, line 58:
```suggestion
| Root container | `{ activities: Activity[] }` | All activities in the chat history |
```

OEvgeny (approved):

PR #5688, opened 2026-02-05
Speech to speech: Mute functionality

pranavjoshi001:
## Changelog Entry

- Added mute/unmute functionality for speech-to-speech with silent chunks to keep server connection alive, in PR [#5688](https://github.com/microsoft/BotFramework-WebChat/pull/5688), by [@pranavjoshi](https://github.com/pranavjoshi001)

## Description

This PR adds mute/unmute functionality for the Speech-to-Speech (S2S) feature as core API only, without UI changes. When muted, the microphone is turned off (browser indicator disappears) but silent audio chunks continue to be sent to keep the server connection alive. This prevents connection timeouts while allowing consumers to implement their own mute UI.

## Design

The mute functionality works at multiple levels:

1. **AudioWorklet Level**: 
   - Added `MUTE` and `UNMUTE` commands
   - When muted, the processor generates silent (all zeros) Int16 audio chunks instead of real audio data
   - This keeps chunks flowing at the same interval to maintain the server connection

2. **useRecorder Hook (`useRecorder.ts`)**:
   - Added `mute()` function that:
     - Sends `MUTE` command to the worklet
     - Disconnects the source node from the audio graph
     - Stops all MediaStream tracks (turns off browser mic indicator)
   - Returns an `unmute()` function that:
     - Sends `UNMUTE` command to the worklet
     - Re-acquires the microphone via `getUserMedia`
     - Reconnects the source node to the audio graph

3. **VoiceRecorderBridge (`VoiceRecorderBridge.tsx`)**:
   - Wires the `mute` function to the voice state machine
   - When voice state transitions to `muted`, calls `mute()` and stores the `unmute` function
   - When voice state transitions back to `listening`, calls the stored `unmute` function

4. **Redux Actions & Hooks**:
   - Added `muteVoiceRecording` and `unmuteVoiceRecording` Redux actions
   - Exposed `useVoiceRecordingMuted` hook for consumers that gives value and setter function.

## Specific Changes

- Added `MUTE` and `UNMUTE` command handling in AudioWorklet processor to generate silent chunks when muted
- Added `mute` function to `useRecorder.ts` hook that disconnects audio and stops MediaStream while continuing to send silent chunks
- Updated `VoiceRecorderBridge.tsx` to handle mute/unmute based on voice state changes
- Added `muteVoiceRecording.ts` and `unmuteVoiceRecording.ts` Redux actions
- Updated `voiceActivity.ts` reducer to handle `VOICE_MUTE_RECORDING` and `VOICE_UNMUTE_RECORDING` actions
- Added useVoiceRecordingMuted` hook and exported them from API, component, and bundle packages
- Added unit tests for mute functionality in `useRecorder.spec.tsx`

- [x] I have added tests and executed them locally
- [x] I have updated `CHANGELOG.md`
- [ ] I have updated documentation

## Review Checklist

- [x] Accessibility reviewed (tab order, content readability, alt text, color contrast)
- [x] Browser and platform compatibilities reviewed
- [x] CSS styles reviewed (minimal rules, no `z-index`)
- [ ] Documents reviewed (docs, samples, live demo)
- [x] Internationalization reviewed (strings, unit formatting)
- [x] `package.json` and `package-lock.json` reviewed
- [x] Security reviewed (no data URIs, check for nonce leak)
- [x] Tests reviewed (coverage, legitimacy)

### DIFF
--- a/packages/api/src/providers/Capabilities/private/fetchCapabilitiesFromAdapter.ts
+++ b/packages/api/src/providers/Capabilities/private/fetchCapabilitiesFromAdapter.ts
@@ -28,7 +28,7 @@ export default function fetchCapabilitiesFromAdapter(
     }
 
     // eslint-disable-next-line security/detect-object-injection
-    const getter = directLine[getterName];
+    const getter = directLine?.[getterName];
 
     if (typeof getter === 'function') {
       try {


packages/api/src/hooks/useMuteVoice.ts
```diff
@@ -0,0 +1,9 @@
+import useWebChatAPIContext from './internal/useWebChatAPIContext';
+
+/**
+ * Hook to mute voice mode (stops microphone input but keeps connection alive with silent chunks).
+ * The session remains active and can be unmuted to resume listening.
+ */
+export default function useMuteVoice(): () => void {
+  return useWebChatAPIContext().muteVoice;
```
OEvgeny: Why we need to extend the main context instead of relying on a separate provider?

packages/api/src/hooks/useUnmuteVoice.ts
```diff
@@ -0,0 +1,9 @@
+import useWebChatAPIContext from './internal/useWebChatAPIContext';
+
+/**
+ * Hook to unmute voice mode (resumes microphone input after muting).
+ * This reactivates speech-to-speech listening.
+ */
+export default function useUnmuteVoice(): () => void {
```
OEvgeny: Have you considered a single hook? `useRecordVoice`/`useControlVoiceRecording` with passes the current state of the recording voice.

packages/api/src/hooks/useMuteVoice.ts
pranavjoshi001: following the same pattern which we did for "useStartVoice" and "useStopVoice"

packages/api/src/hooks/useUnmuteVoice.ts
pranavjoshi001: Initially I created single hook to return voice with state but then William suggested to keep individual hook for each export and same followed for start voice, stop voice and voice state

packages/api/src/hooks/useMuteVoice.ts
compulim: I think EugeneO's concern is valid. We are slowly dissecting the `APIComposer`, `CoreComposer` into smaller `XXXComposer` to improve performance and enabling plug-and-play.

Why performance... if one thing in the composer context changed, it will propagate to every component that call `useContext(context)` to subscribe.

Says, if `styleOptions` changed, the `APIComposer` context would change, and every component that relies on `useMuteVoice()` will be re-rendered even if they did not call `useStyleOptions`.

Can you move it to `<SpeechToSpeechComposer>`?

packages/api/src/hooks/useUnmuteVoice.ts
compulim: Deeper reason... one-liner: `useStartVoice` will do some expensive operation that takes time to complete. Says, `new AudioContext()` is a UI blocker and it takes time to complete.

Let's say, in parallel universe, we have `useVoice(): [boolean, (enabled: boolean) => void]`, what would happen:

- Call `useVoice()[1](true)` to start the microphone
- Assume it takes 1 second to turn on the microphone
- Within 1 second, we call `useVoice()[0]`, it will return `false`
- 1 second later, `useVoice()[0]` will return `true`

The behavior will become quite undeterministic... says, we have a microphone button that essentially a push button like `<input type="checkbox">` and is backed by this `useVoice()` boolean.

When the user click on the push button, it will not be pushed/checked because the getter is still returning `false`. 1 second later, it is pushed/checked. The user will be confused, because it breaks WYSIWYG.

This is the main reason why `useStartVoice`/`useStopVoice` is preferred over `useVoice`... not to mention voice has more state than `true`/`false`. Primarily, the setter takes time to complete the operation, the getter and setter will be momentarily out-of-sync, this is the main reason we prefer callback functions than getter/setter.

---

In this case, if the "mute" is instant/synchronous, which I believe so, we should do it the normal way, i.e. `useMicrophoneMuted(): readonly [boolean, (value: Dispatch<SetStateAction<boolean>>) => void]`.

packages/api/src/providers/SpeechToSpeech/private/VoiceRecorderBridge.tsx
```diff
@@ -29,7 +30,13 @@ export function VoiceRecorderBridge(): null {
     [postVoiceActivity]
   );
 
-  const { record } = useRecorder(handleAudioChunk);
+  const { record, mute } = useRecorder(handleAudioChunk);
```
compulim: Sort.

```suggestion
  const { mute, record } = useRecorder(handleAudioChunk);
```

packages/api/src/providers/SpeechToSpeech/private/VoiceRecorderBridge.tsx
```diff
@@ -29,7 +30,13 @@ export function VoiceRecorderBridge(): null {
...
+
+  useEffect(() => {
+    if (muted) {
```
compulim, line 36: Curious, how about `unmute()`?

Can we put it here so things is symmetric and easier to debug?

packages/api/src/providers/SpeechToSpeech/private/useRecorder.ts
```diff
@@ -131,16 +164,57 @@ export function useRecorder(onAudioChunk: (base64: string, timestamp: string) =>
...
+    chunkIntervalMs,
+    initAudio,
+    onAudioChunk,
+    sampleRate,
+    workletRef
+  ]);
+
+  const muteRecording = useCallback(() => {
+    // Stop MediaStream (mic indicator OFF) and disconnect source
+    stopMediaStream();
```
compulim, line 186: Disconnect source is a good idea.

Did you test it on iPhone, Android, and Firefox? Few aspects:

- How long does it takes to acquire/connect again?
   - If it's > 500ms, that means we will lost what the user says but the user never know about it. It may be a better idea to keep the microphone continue to be connected.
- Will the browser prompt the privacy dialog again? It will be undesirable

After muted for 1 minute, should we stop recording?

packages/api/src/providers/SpeechToSpeech/private/useRecorder.ts
```diff
@@ -131,16 +164,57 @@ export function useRecorder(onAudioChunk: (base64: string, timestamp: string) =>
...
+    worklet.port.postMessage({ command: 'START' });
+  }, [
+    Date,
+    acquireAndConnectMediaStream,
+    audioCtxRef,
+    chunkIntervalMs,
+    initAudio,
+    onAudioChunk,
+    sampleRate,
+    workletRef
```
compulim: Sort.

packages/api/src/providers/SpeechToSpeech/private/useRecorder.spec.tsx
```diff
@@ -218,4 +221,74 @@ describe('useRecorder', () => {
...
+    act(() => {
+      hookData?.mute();
+    });
+
+    // Should send MUTE command to worklet
+    expect(mockWorkletPort.postMessage).toHaveBeenCalledWith({ command: 'MUTE' });
+    // Should stop media stream tracks (mic indicator OFF)
+    expect(mockTrack.stop).toHaveBeenCalledTimes(1);
+    // Should disconnect source node
+    expect(mockSourceNode.disconnect).toHaveBeenCalledTimes(1);
```
compulim, line 258: It should also test if the stream is zeroed out.

packages/api/src/providers/SpeechToSpeech/private/useRecorder.spec.tsx
```diff
@@ -218,4 +221,74 @@ describe('useRecorder', () => {
...
+      unmute?.();
+    });
+
+    // Should send UNMUTE command to worklet
+    expect(mockWorkletPort.postMessage).toHaveBeenCalledWith({ command: 'UNMUTE' });
+    // Should restart media stream
+    await waitFor(() => {
+      expect(mockMediaDevices.getUserMedia).toHaveBeenCalledTimes(1);
+    });
+  });
```
compulim, line 343: It should test it no longer send out zeroes.

packages/api/src/providers/SpeechToSpeech/private/useRecorder.ts
```diff
@@ -20,26 +22,37 @@ declare function registerProcessor(name: string, processorCtor: typeof AudioWork
...
+  const RENDER_QUANTUM = 128;
+
   class AudioRecorderProcessor extends AudioWorkletProcessor {
     constructor(options: AudioWorkletNodeOptions) {
       super();
       this.buffer = [];
       this.bufferSize = options.processorOptions.bufferSize;
+      this.muted = false;
       this.recording = false;
+      this.silentFrame = new Float32Array(RENDER_QUANTUM); // Pre-allocated zeros
```
compulim: Consider move `this.silentFrame` to module-level constant `const SILENT_FRAME = new Float32Array(RENDER_QUANTUM)`.

packages/api/src/providers/SpeechToSpeech/private/useRecorder.ts
```diff
@@ -20,26 +22,37 @@ declare function registerProcessor(name: string, processorCtor: typeof AudioWork
...
         }
       };
     }
 
     process(inputs: Float32Array[][]) {
-      if (inputs[0] && inputs[0].length && this.recording) {
-        this.buffer.push(...inputs[0][0]);
+      if (this.recording) {
+        // Use real audio when not muted, otherwise silenced chunk to keep connection alive (all zeros).
+        const audioData = !this.muted && inputs[0] && inputs[0].length ? inputs[0][0] : this.silentFrame;
```
compulim: Could simplify this condition.

packages/api/src/hooks/useMuteVoice.ts
pranavjoshi001: makes sense, since we have everything in "voice" redux slice and no local state in composer context, I have moved it to `useVoiceRecordingMuted` hook and that handles of returning value and setter function, I have named it to align with existing voice function to avoid ambiguity with existing microphone related hooks and functions.

packages/api/src/hooks/useUnmuteVoice.ts
pranavjoshi001: makes sense, since we have everything in "voice" redux slice and no local state in composer context, I have moved it to `useVoiceRecordingMuted` hook and that handles of returning value and setter function, I have named it to align with existing voice function to avoid ambiguity with existing microphone related hooks and functions.

packages/api/src/providers/SpeechToSpeech/private/useRecorder.ts
pranavjoshi001: Yes I have tested it for all of them and working fine
1) It takes hardly 8ms to 40ms to acquire/connect again so there won't be any loss. (keeping microphone connected shows browser mic enabled which can cause privacy concern so that's why disconnecting)
2) No, browser will not prompt again

After muted for 1 minute, should we stop recording? - good suggestion, will check with PM on this what is expected and if we can do this but for now I think we might want to keep it in mute state until user unmute it.

packages/api/src/providers/SpeechToSpeech/private/useRecorder.spec.tsx
pranavjoshi001: added

packages/api/src/providers/SpeechToSpeech/private/useRecorder.spec.tsx
pranavjoshi001: added

packages/api/src/providers/SpeechToSpeech/private/useRecorder.ts
pranavjoshi001: done

pranavjoshi001:
> Add end-to-end tests.

done

packages/api/src/providers/SpeechToSpeech/private/VoiceRecorderBridge.tsx
pranavjoshi001: I am not sure if I quite got you here
so, state always changes first and we act on this bridge for example for "recording" we initial record(), same way for "muted" we want to initial mute()

Can you please explain more on how you want me to use unmute() here ? are you suggesting to expose both mute and unmute function and not using unsubscribe patter here ?

__tests__/html2/speechToSpeech/mute.unmute.html
```diff
@@ -0,0 +1,214 @@
...
+
+        // Helper to check if audio has non-zero data (real audio)
+        function hasNonZeroAudio(base64Content) {
+          const binaryString = atob(base64Content);
+          const bytes = new Uint8Array(binaryString.length);
+          for (let i = 0; i < binaryString.length; i++) {
+            bytes[i] = binaryString.charCodeAt(i);
+          }
+          return bytes.some(byte => byte !== 0);
+        }
```
compulim, line 62: Aren't they simple negation of each other? 🤣

packages/api/src/providers/SpeechToSpeech/private/VoiceRecorderBridge.tsx
compulim: Ah, I missed the point... `return`...

`mute(): () => void` so it unmute on unmount, etc.

It's good design.

__tests__/html2/speechToSpeech/mute.unmute.html
compulim: It's okay to work on it when we touch it.

PR #5811, opened 2026-04-22
Fix undefined reference error in SendFailed.tsx

### LINKED ISSUE #5812: Undefined reference in SendFailed.tsx - TypeError: Cannot read properties of undefined (reading 'has')

No

No

Latest production

Bundle (webchat.js)

Dynamics 365 Omnichannel

Browser: Chrome (latest)

Chat history

### Which theme pack does this issue affect?

I did not test it on other theme packs

### What is the public URL for the website?

_No response_

### Please describe the bug

We are seeing the following error in our application telemetry on attempting to render the web chat control, which appears to be causing rendering errors from what is reported by our own customers:

```
TypeError: Cannot read properties of undefined (reading 'has')
```

This is coming from the bundled webChat.js, and GitHub Copilot has provided the following failing region at which the issue is occurring:

```js
uE = () => {
   let [e] = cE(),           // sendStatusByActivityKey (Map)
       t = aE(),             // activity key resolver
       r = sE(),             // localization hook
       n = oE(),             // live region announcer
       // Build Set of failed activity keys:
       i = useMemo(() =>
         Array.from(e).reduce(
           (e, [r, n]) => n !== tE || Dy(t(r)) ? e : e.add(r),
           new Set()
         ), [t, e]),
       o = r("TRANSCRIPT_LIVE_REGION_SEND_FAILED_ALT"),
       a = rE(i),            // usePrevious(i) → undefined on 1st render
       s = useMemo(() => {
         if (i === a) return false;
         for (let e of i.keys())
           if (!a.has(e))     // 💥 a is undefined on first mount
             return true;
         return false;
       }, [i, a]);
   return useEffect(() => { s && n(o) }, [s, o, n]), null;
 };
```

This appears to be an issue in SendFailed.tsx in which we can have an undefined value for `prevActivityKeysOfSendFailed` based on rendering order.

```js
See above.
```

Still determining exact repro steps.

### What do you expect?

No errors. :)

### What actually happened?

Customer reports indicate that the web chat control can fail to render; in cases where this has been reported, we see this error, so we believe this is linked to the observed experience.

### Do you have any screenshots or recordings to repro the issue?

Not available.

### Adaptive Card JSON

```json

```

### Additional context

_No response_

peterinnesmsft:
> Fixes #5812

## Description

In Dynamics 365 Contact Center, an error has been discovered in our own telemetry which points to an error occurring within SendFailed.tsx. The stack trace is as follows:

```
TypeError: Cannot read properties of undefined (reading 'has')
```

GitHub Copilot analysis points to the following minified code as being the problem:

```js
uE = () => {
   let [e] = cE(),           // sendStatusByActivityKey (Map)
       t = aE(),             // activity key resolver
       r = sE(),             // localization hook
       n = oE(),             // live region announcer
       // Build Set of failed activity keys:
       i = useMemo(() =>
         Array.from(e).reduce(
           (e, [r, n]) => n !== tE || Dy(t(r)) ? e : e.add(r),
           new Set()
         ), [t, e]),
       o = r("TRANSCRIPT_LIVE_REGION_SEND_FAILED_ALT"),
       a = rE(i),            // usePrevious(i) → undefined on 1st render
       s = useMemo(() => {
         if (i === a) return false;
         for (let e of i.keys())
           if (!a.has(e))     // 💥 a is undefined on first mount
             return true;
         return false;
       }, [i, a]);
   return useEffect(() => { s && n(o) }, [s, o, n]), null;
 };
```

## Design

The issue appears to be caused on initial render of the web chat control; we still don't quite fully understand the exact condition, but it seems like this is due to the ordering of the underlying `useRef` in `usePrevious` starting as undefined, and only being updated after initial render in `useEffect`, meaning that on initial render we can hit an undefined reference on `prevActivityKeysOfSendFailed`.

## Specific Changes

- Update `/component/src/hooks/internal/usePrevious.ts` to match the duplicated implementation that allows for setting an initial value.
- Update `SendFailed.tsx` to specify an empty Set<string> as the default value in `usePrevious` for `prevActivityKeysOfSendFailed`.
- Adding an explicit falsey check in `hasNewSendFailed` on `prevActivityKeysOfSendFailed` so that if it doesn't match `activityKeysOfSendFailed` and evaluates to false, we return true to indicate that a new send has indeed failed.

TODO: Need to address the below.

- [ ] I have added tests and executed them locally
- [ ] I have updated `CHANGELOG.md`
- [ ] I have updated documentation

## Review Checklist

- [ ] Accessibility reviewed (tab order, content readability, alt text, color contrast)
- [ ] Browser and platform compatibilities reviewed
- [ ] CSS styles reviewed (minimal rules, no `z-index`)
- [ ] Documents reviewed (docs, samples, live demo)
- [ ] Internationalization reviewed (strings, unit formatting)
- [ ] `package.json` and `package-lock.json` reviewed
- [ ] Security reviewed (no data URIs, check for nonce leak)
- [ ] Tests reviewed (coverage, legitimacy)


compulim:
Please write a test to show how we previously failed the case. This will make sure the new code is fortifying things in proper places.

Also, instead of modifying `usePrevious`, which could have a big impact, we could simplify things by doing `set?.has()`.

